PR #618, opened 2015-02-02
Fix SF Bug 1337 - Cannot post AP transactions

nick-prater:
Code was appending the account number to $form->{oldvendor}, even
though this was already done. Doing it twice broke later comparisons
between $form->{vendor} and $form->{oldvendor}.


mcr:
I have tested this patch, and it works, please merge and release 1.4.10.2.

PR #621, opened 2015-02-15
Fix html errors in /UI/payments/payments_detail.html

nick-prater:
/UI/payments/payments_detail.html has a number of errors. From a user-perspective, this causes the filter dates display to be messed up. In investigating this, I discovered a number of other errors in the html structure, which are also corrected by this patch.

* reformat for consistent indentation
    previously a mix of spaces and tabs made it hard to match opening/closing tags

* Remove label tag from around filter_to caption.
    This was causing incorrect display:
    was: "Filtering From:       To: 2015-02-13 2015-02-15"
    now: "Filtering From: 2015-02-13   To: 2015-02-15"

* add missing closing div tag to "date_filter_row"

* remove spurious closing td tag from "account_info" span

* change details_select span to a div (because it contains divs.
    A span cannot contain divs, only other inline tags.)

* invoice_list td element : fix misplaced closing td tag
    (content was outside closing tag)

* total_due_list td element : fix misplaced closing td tag
    (content was outside closing tag)

* invoice_detail_list td element : add missing closing tag

* container div : add missing closing tag

There are a number of places where a label tag is being used to add a caption
to a span. This is invalid html as label elements are only applicable to form
elements. I have left this as-is because changing it would affect the page
styling and in practice it is working OK.

### DIFF
--- a/UI/payments/payments_detail.html
+++ b/UI/payments/payments_detail.html
@@ -38,21 +38,12 @@
 	value = account_class
 
  } ?>
-
-
-
-
 <?lsmb INCLUDE input element_data = {
 	type = "hidden"
 	name = "payment_type_id"
 	value = payment_type_id
 
  } ?>
-
-
-
-
-
  <?lsmb INCLUDE input element_data = {
 	type = "hidden"
 	name = "department_id"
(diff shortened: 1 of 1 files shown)


einhverfr:
looks like we have a pod error and a test case that needs adjusting.  Both o those should be pretty minor.

Thanks for this.

einhverfr:
ok I am merging this and will fix the pod and test case separately on my branch and then merge back.  This isn't ideal but I don't see a better way of doing this at present.

PR #642, opened 2015-04-22
Update Changelog

pongraczi:


pongraczi:
Updating Changelog to include bugfix note for #1358

PR #644, opened 2015-04-22
1.4 fix header count

ehuelsmann:
Fixed: Headers which have associated accounts as their children, but no other headers, have a [Delete] action on the CoA screen due to incorrect reference counts.


ehuelsmann:
Merging but not changing Changelog, because this is a fix on top of a change already in the Changelog for this release.

Issue #648, opened 2015-04-23
employee.pl referenced in multiple places, but not functional anymore


ehuelsmann:
./sql/modules/menu_rebuild.sql:48	module	employee.pl	113
./UI/Admin/edit_user.html:       <a href="employee.pl?action=edit&entity_id=<?lsmb e_id ?>">[<?lsmb text('Edit') 
./UI/employee/filter.html:<form action="employee.pl" method="POST" name="search_filter">
./UI/employee.html:<form action="employee.pl" method="post">
./LedgerSMB/Sysconfig.pm:asset.pl    employee.pl  login.pl      setup.pl    vouchers.pl

ehuelsmann:
The merged commit fixes this bug and has been grafted to master. Closing.

PR #650, opened 2015-04-23
* Remove references to employee.pl script which proxies deleted employee...

ehuelsmann:
....pm.


ehuelsmann:
See https://github.com/ledgersmb/LedgerSMB/issues/648 for more references that need to be removed.

PR #703, opened 2015-07-20
Use Travis Perl helpers

ehuelsmann:
Tremendous load reduction on Travis; Our test-time drops from 12-15 minutes down to 3-4 minutes.


ehuelsmann:
Unfortunately, this commit changes the build output; can we silence the 'mkmanifest' step?

einhverfr:
I like this.

PR #727, opened 2015-08-07
User/employee fixes

einhverfr:
This corrects a number of disabling (but trivial to fix) problems with employee entry.


ehuelsmann:
Hi Chris,

The process works smooth with this change, however, I have a few findings in my testing:

 - Salutation gets lost when clicking "Save"
 - "Type" dropdown after saving indicates "Vendor", even though I explicitly selected "Employee" before saving

Note that my workflow was:

 - Go to Add employee
 - Fill the Person form
 - "Generate control code"
 - Save the Person form

this being the case, the salutation and type could have been lost on "Generate control code".

Issue #733, opened 2015-08-12
Local change in LedgerSMB/Scripts/Payment.pm


ehuelsmann:
John has the following "patch" is his local tree; we should investigate if we need to incorporate in the base line:

````
the other [patch] is in LedgerSMB/Scripts/Payment.pm
freelock: sub print
3 lines that change variable expansion, changing _$inv_id to _${inv_id}
must be something I fixed related to invoice numbers showing up on cash ->vouchers -> payments
````

PR #746, opened 2015-08-15
Expand ${inv_id} instead of ${inv}_id

ehuelsmann:
@freelock is this the change you have in your production system which I recorded in #733 too?


freelock:
Yes, this matches the code in my production...

PR #761, opened 2015-08-21
Support deletion of certain objects along with their referents

ehuelsmann:


ehuelsmann:
@einhverfr the cascade on the new_shipto table fixes deletion of orders. The others I added as part of my review of the schema. Please tell me which ones you agree and don't agree with.

Issue #765, opened 2015-08-23
"Manage users" permission (at db creation) doesn't work


ehuelsmann:
As of the consolidation of employee entry with contact management (or before?), the "manage users" permission for the initial user doesn't work anymore: the menu of a user created with those permissions consists only of:

* Contacts -> Search
* System -> Sessions
* Preferences
* New window
* Logout

Note that there's no option to create users...

ehuelsmann:
Didn't get to doing this. Moving to 1.4.17.

PR #770, opened 2015-08-23
1.4 setup user manager

ehuelsmann:
Works around #765


ehuelsmann:
@einhverfr please review.

Also, I could really use some ideas about what permissions the initial user should get... (I'm thinking to remove "Manage users" and to assign full permissions always...)

PR #810, opened 2015-09-09
Fix #792, fix #795, fix #796

### LINKED ISSUE #792: Can't post GL transaction
Due to Dojo's "required" checks, the additional (unused) rows in the GL entry screen report being "required". By consequence the form isn't being submitted.

Solution: as soon as either a debit or credit is entered, the account entry box should be set to "required" (but it should *not* be required until that time).

### LINKED ISSUE #795: journal entry screen doesn't create new lines after filling the 10 present
When clicking 'Update' with 10 lines in the journal entry filled, the update is supposed to come up with extra lines. This doesn't happen.

### LINKED ISSUE #796: Journal entry drops 10th account
When clicking "Update" on the Journal entry screen with 10 filled lines, the returned screen contains the figures entered on the 10th line, but not the account.

ehuelsmann:
Once validated, this PR fixes all the issues (Fix #792, fix #795, fix #796) relating to the malfunction of the GL entry screen.

However, problems with PGDate currently block my (manual) testing.


ehuelsmann:
With the fixes to PGDate merged, I was able to further test this branch (resulting in the extra commits).

Can merge now, as it works for me.

PR #824, opened 2015-09-12
* Instead of clearing the entire %ENV hash, set local values (fix #813)

### LINKED ISSUE #813: "local %ENV" clobbers PATH
As per http://archive.ledgersmb.org/ledger-smb-users/msg08586.html, the

```perl
   local %ENV;
```

statements in Database.pm (in 1.4-only), clobber the ```$ENV{PATH}``` setting.

The cause has been determined to be that the statement above completely clears %ENV.

ehuelsmann:


ehuelsmann:
User confirmed fix here: http://archive.ledgersmb.org/ledger-smb-users/msg08594.html

PR #913, opened 2015-10-18
fixing new window link, closes #892

### LINKED ISSUE #892: "New window" menu doesn't work
Due to URL target rewriting, the "New window" menu option no longer works in master (for 1.5).

einhverfr:


UI/menu/expanding.html
```diff
@@ -37,7 +37,10 @@
 					target = '_top'
 				<?lsmb END ?>
 				id="a_<?lsmb- item.id -?>"
-                                class="<?lsmb IF item.module; 'menu-terminus';
+                                class="<?lsmb IF item.label == 'New Window';
```
ehuelsmann: @einhverfr does this line work with translated menus?

einhverfr: see line 42/45 below.  <?lsmb text(item.label) ?> is where it is translated (so on display).

PR #1062, opened 2015-11-21
* Prepare tools/ directory for release -- clean up

ehuelsmann:
Specifically, we don't want to support CGI/FCGI scenarios anymore: we've decided on PSGI going forward


einhverfr:
I think we should leave the psgi files in place as they are example configurations.  Good to remove the rest though.

einhverfr:
(there were two psgi files removed by this)

einhverfr:
On second thought these are still on the 1.4 branch and I am not sure they are sufficiently maintained.  I am merging their removal on the view that we can bring them back if needed from 1.4 if we get a real maintainer.

PR #1167, opened 2015-12-28
1.3 po updates

ehuelsmann:
Even though 1.3 isn't going to see releases, people might want to download updated translations based on the translation efforts on Transifex. In order to take advantage of these translation improvements, we need the updated POT file in the 1.3 branch of the main repository.


ehuelsmann:
Tests failed DB version check (version is 1.3.48-dev; however -dev suffix is unexpected...). Already had this error before this PR.

ehuelsmann:
So, merging anyway.

Issue #1198, opened 2016-01-04
setup.pl needs an overhaul to improve user experience


sbts:
There are a number of user experience issues with the current implementation of setup.pl

The first one is we need some help information on the login screen. at the moment it is expected that users just know that the Database field will either 
    \* create a new db if the text is not a valid db name
    \* allow changes to and/or backup of an existing db if the text is a valid  db name
~~There is another ticket #1197 dealing with the hardcoded default 'postgres' for the login name.~~ (this is solved)

Really the Database field is probably not needed here these days.
Instead
    \* simple login screen
    \* once logged in present a list or dropdown list of available DB's
    \* make the list filterable (dojo can filter select elements)
    \* add a button "Create New Company Database"
    \* add text that makes it clear that the company name is the database name
    ~~\* add a system level (ledgersmb.conf) blacklist regex that hides db's from the list.
        this would be used to hide things like postgres, template0, template1~~ (a hard coded list exists now)
    ~~\* add a configurable (in databse) blacklist regex that hides db's from the list.
        this would allow site specific hiding~~ (site specific hiding works through LedgerSMB.conf)

Once you have opened a config page for a specific DB there are a couple of other issues.
    \* If the db is not a recognised ledgersmb db the cancel button should not be labeled "Yes"
        but either "cancel" or "back".
    \* The backup buttons should not be active, or not visible at all.
    \* when the cancel or back button is clicked it should return to the list, rather than the login screen

sbts:
we should also return result confirmation dialogs for otherwise silent actions.
A good example here is the "password change" button for changing a user password.
When clicked it silently completes with the only outward change to the UI being any entered password is cleared

ylavoie:
Switch order of 'Date of birth' and 'Country' on user creation page so that date format can be localized.

einhverfr:
I think a couple key improvements are possible here:

1.  App::LedgerSMB::Admin::Web should be able to handle the web-service backend calls here, but we need to add user handling etc to that.
2.  Supply a dojo-based admin widget here.  The above merely does a form and jquery interface for html (functional but minimalistic).
3.  After logging in, the screen should pop up either a db creation wizard (if no db) or a db console with functions with no confirmation portion, just a title, list of actions, and db statistics.

ehuelsmann:
First we need to define what setup.pl means to be/contribute.

ehuelsmann:
Consolidated into #3746

Issue #1197, opened 2016-01-04
setup.pl default user needs to be changed and retained during a session


sbts:
Currently when accessing setup.pl the postgres login name is pre-filled with postgres.
As of 1.5 this is needs to be changed to follow installation document recommendation that we use a default user of
    lsmb_dbadmin

Also during a browser session the login name should be retained (in case an installation uses something different.)
This may be worth storing in a local cookie, or a hint in ledgersmb.conf

Issue #1178, opened 2015-12-31
Minor spelling errors in v1.3 & v1.4 LedgerSMB/DBObject/Payment.pm


jame:
Minor spelling errors in  v1.3 LedgerSMB/DBObject/Payment.pm (v1.3.47):

Line 284 in ~/Projects/ledgersmb/work/ledgersmb/LedgerSMB/DBObject/Payment.pm  wich which
Line 423 in ~/Projects/ledgersmb/work/ledgersmb/LedgerSMB/DBObject/Payment.pm  BEEING BEING

jame:
Also;  there are minor spelling errors in v1.4 LedgerSMB/DBObject/Payment.pm (v1.4.21):

Line 281 in ~/Projects/ledgersmb/work/ledgersmb/LedgerSMB/DBObject/Payment.pm wich which
Line 421 in ~/Projects/ledgersmb/work/ledgersmb/LedgerSMB/DBObject/Payment.pm BEEING BEING

jame:
Will need to check v1.4.22 when it comes our, to check where the spelling errors in LedgerSMB/DBObject/Payment.pm are in that version.

jame:
Minor spelling errors in v1.3 LedgerSMB/DBObject/Payment.pm (v1.3.47):

Line 284 in LedgerSMB/DBObject/Payment.pm wich which
Line 423 in LedgerSMB/DBObject/Payment.pm BEEING BEING

jame:
Also; there are minor spelling errors in v1.4 LedgerSMB/DBObject/Payment.pm (v1.4.21):

Line 281 in LedgerSMB/DBObject/Payment.pm wich which
Line 421 in LedgerSMB/DBObject/Payment.pm BEEING BEING

PR #1218, opened 2016-01-05
Master issue 1179

sbts:
Changed setup.pl username input to checkbox

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,7 @@ ledgersmb.conf
 /doc/manual/*.toc
 /doc/manual/*.log
 /doc/manual/*.out
+
+# Custom extensions to Tests and Validations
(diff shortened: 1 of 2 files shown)


ehuelsmann:
@sbts I think the issue you want to be (partially) addressing is not #1178, but #1198 instead?

sbts:
Oops actually neither.
I should have referred to #1197 as being fixed.
Sorry about that.

ehuelsmann:
@sbts your code can be slightly improved: see https://dojotoolkit.org/documentation/tutorials/1.10/selects/#dijit-form-combobox ; the combobox has a "placeHolder" value you want to set for the "choose" value you're initializing with. The combobox then recognizes it doesn't have a value yet, but does show the placeholder.

PR #1234, opened 2016-01-08
Master issue 1198 Partial Fixes

sbts:
More fixes to go for #1198  but this is a good start
Also fixed missing defaults in Sysconfig.pm 
Also fixed problems if ledgersmb.conf was missing or unreadable.
      this caused everything to blow up badly

### DIFF
--- a/LedgerSMB/Sysconfig.pm
+++ b/LedgerSMB/Sysconfig.pm
@@ -104,7 +104,7 @@ our $DBI_TRACE=0;
 our %printer = ();
 
 # Start With an empty default config
-my $cfg=Config::IniFiles->();
+my $cfg=Config::IniFiles->new();
 if ( -r 'ledgersmb.conf' ) {
     # Try Loading the config file
     $cfg->ReadFile('ledgersmb.conf' );
--- a/LedgerSMB/Sysconfig.pm
+++ b/LedgerSMB/Sysconfig.pm
@@ -103,11 +103,14 @@ our $DBI_TRACE=0;
 # available printers
 our %printer = ();
 
-my $cfg;
-# Try Loading the config file
-eval { $cfg = Config::IniFiles->new( -allowempty => 1, -reloadwarn => 1, -file => "ledgersmb.conf" ); }
-# Fall back to an empty default config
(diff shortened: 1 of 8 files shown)


LedgerSMB/Sysconfig.pm
```diff
@@ -98,9 +101,13 @@ our $no_db_str = 'database';
...
 # available printers
-our %printer;
+our %printer = ();
 
-my $cfg = Config::IniFiles->new( -file => "ledgersmb.conf" ) || die @Config::IniFiles::errors;
+my $cfg;
+# Try Loading the config file
+eval { $cfg = Config::IniFiles->new( -allowempty => 1, -reloadwarn => 1, -file => "ledgersmb.conf" ); }
+# Fall back to an empty default config
+if ( ! $cfg ); $cfg=new Config::IniFiles( );
```
sbts: I think this is the cause of travis test failures.
The config file is being read but incorrectly overridden by line 110

ehuelsmann: @sbts that's correct. the $cfg=... piece of code isn't linked to the 'if' statement, because that statement was terminated by the semi-colon before the '$cfg='. So, it's an unconditional statement.

ehuelsmann: Actually, I think I would solve it this way:

my $cfg = Config::IniFiles->new();
$cfg->ReadFile('ledgersmb.conf') if -r 'ledgersmb.conf';

sbts: Only downside to that may be due to no longer getting a warning on stderr that there was no config file loaded

ehuelsmann: Agreed. That's easily solved:

my $cfg = Config::IniFiles->new();
if  (-r 'ledgersmb.conf') {
   $cfg->ReadFile('ledgersmb.conf');
}
else {
   warn "Could not load 'ledgersmb.conf'" if ! -r 'ledgersmb.conf';
}

ehuelsmann:
Closing this PR, pending splitting up in parts and resolution of the merge conflicts.

PR #1235, opened 2016-01-08
* Fix #976: fix 'Update' adding lines on orders and invoices

### LINKED ISSUE #976: duplicate line in order entry in 1.4.10.1, but not in 1.4.6
https://sourceforge.net/p/ledger-smb/bugs/1335/

ehuelsmann:
Note: With this commit, the minimum number of empty lines doesn't seem
  to be consistently enforced, but every update isn't adding lines anymore
  which is (IMO) more important.

### DIFF
--- a/bin/io.pl
+++ b/bin/io.pl
@@ -39,7 +39,6 @@
 #######################################################################
 
 package lsmb_legacy;
-use List::Util qw(max);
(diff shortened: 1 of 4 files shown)


ehuelsmann:
@einhverfr Please review and test (and merge if acceptable)

PR #1273, opened 2016-01-30
* Silence setup and login logs

ehuelsmann:
Depends on ledgersmb/PGObject-Util-DBAdmin#5


ehuelsmann:
New PGObject::Util::DBAdmin version released and available on CPAN, so, this can be merged now.

PR #1347, opened 2016-02-18
initial outline of db update framework with tests, Fixes #840

### LINKED ISSUE #840: Use Sqitch-like management and deployment
In order to improve on our schema deployment control, we want to use prior work done in this area by other projects, hence our preference for Switch, which supports postgresql and our choice of versioning tools (git/hg).

This item has been discussed on the development mailing list and with @einhverfr and @freelock . 

Next step is to come up with a plan to move from our current Fixes.SQL based system to a fully sqitch managed system.

When we have that, we also need to write an instruction guide on how to use sqitch to add customizations to the LedgerSMB schema.

einhverfr:
Unit tests only, no integration tests yet, not worked into the framework yet.


einhverfr:
ok things I dont like about this approach to start with

The sigils strike me as hard to maintain long run but they are a quick way of getting started.  Longer-run I would like to have richer information in a comment header on each change file (including dependencies).  This would allow for automatic detection of when a dependency needs to be reloaded.

But for now this seems to work (except that there is insufficient error checking in places).

ehuelsmann:
If you don't like the approach, should we instead discuss further before more time on this direction?

einhverfr:
ok in terms of current stuff I want to remove the reload subsequent flag from the LOADORDER handling.  However the succeed-if-fail flag is useful there.

Further handling I think should happen in scripts, not in the LOADORDER.  But that is for future work.

I like what we have right now as far as stuff working.

LedgerSMB/Database/Change.pm
```diff
@@ -0,0 +1,280 @@
...
+
+=item reload_subsequent
+
+If this one has changed, then reload further modules
+
+=back
+
+=cut
+
+sub new {
```
ehuelsmann: @einhverfr I see you didn't go with a Moose class. Any reason in particular?

LedgerSMB/Database/Change.pm
```diff
@@ -0,0 +1,280 @@
...
+
+=cut
+
+sub content {
+    my ($self, $raw) = @_;
+    unless ($self->{_content}) {
+        my $file;
+        local $!;
+        open(FILE, '<', $self->path) or
+            die 'FileError: ' . Cwd::abs_path($self->path) . ": $!";
```
ehuelsmann: Hmm.  Can we open with utf8 directly here instead of switching to it below?

LedgerSMB/Database/Change.pm
```diff
@@ -0,0 +1,280 @@
...
+
+Initializes the tracking system
+
+=cut
+
+sub init {
+    my ($dbh) = @_;
+    return 0 unless needs_init($dbh);
+    $dbh->do("
+    CREATE TABLE db_patch_log (
```
ehuelsmann: What happens on failure? If we depend on RaiseError to be true, then we should document here and where we receive the handle.

LedgerSMB/Database/Change.pm
```diff
@@ -0,0 +1,280 @@
...
+
+Returns true if the tracking system needs to be initialized
+
+=cut
+
+sub needs_init {
+    my ($dbh) = @_;
+    my $rows = $dbh->prepare(
+       "select 1 from db_patches"
+    )->execute();
```
ehuelsmann: What if RaiseError is true? Then when there is no patches table, this will die, I think?

LedgerSMB/Database/Loadorder.pm
```diff
@@ -0,0 +1,176 @@
...
+}
+
+=head2 scripts
+
+Returns a list of LedgerSMB::Database::Change objects
+
+=cut
+
+my $reload_subsequent;
+sub scripts {
```
ehuelsmann: Reload subsequent is set to 0 (false) twice in the routine below,  it only really used in the routine after that. I'd suggest moving the declaration down one function.

LedgerSMB/Database/Loadorder.pm
```diff
@@ -0,0 +1,176 @@
...
+    my ($self) = @_;
+    return @{$self->{_scripts}} if $self->{_scripts};
+    my $loadorder;
+    local $!;
+    local $@;
+    open(LOAD, '<', $self->{_path}) or
```
ehuelsmann: Do we put a restriction on file names? Like afaik only or utf8? In case of the latter, we want to open loadorder appropriately.

LedgerSMB/Database/Loadorder.pm
```diff
@@ -0,0 +1,176 @@
...
+
+Creates an index of the files at $path/LOADORDER.idx and locks the file.
+
+If the file is already locks throws an error "LockError"
+
+The LOADORDER.idx remains empty.
+
+=cut
+
+sub makeindex {
```
ehuelsmann: From the docs, it's not clear what we need the index for. Can we do with a temp file and lock $loadorder for exclusive reading? That prevents requiring write access in the stored application tree.

LedgerSMB/Database/Loadorder.pm
```diff
@@ -0,0 +1,176 @@
...
+    my ($self) = @_;
+    die 'LockError' if -f $self->path('LOADORDER.idx');
+    open TEMP, '>', $self->path('LOADORDER.idx');
+    $self->{_locked} = 1;
+    for my $script ($self->scripts){
+        $script->load_contents;
+    }
+    close TEMP;
```
ehuelsmann: Hmmm. If we leave things like this and we crash hard elsewhere, our patch directory remains locked forever. Can't we remove the lock file in something like a finally clause in the same routine where we open/create it?

LedgerSMB/Database/Loadorder.pm
```diff
@@ -0,0 +1,176 @@
...
+If the file is already locks throws an error "LockError"
+
+The LOADORDER.idx remains empty.
+
+=cut
+
+sub makeindex {
+    my ($self) = @_;
+    die 'LockError' if -f $self->path('LOADORDER.idx');
+    open TEMP, '>', $self->path('LOADORDER.idx');
```
ehuelsmann: This is a race condition. Can't we "create, fail if exists"?

sql/changes/LOADORDER
```diff
@@ -0,0 +1,31 @@
+# This is the base file used to build the LOAD.idx file
+# It specifies a lost of db changes in order.
```
ehuelsmann: Typo: 'lost'

ehuelsmann:
One general remark: can't we require pathname to contain no spaces? Then we could have a second "unspaced" block with application parameters, possibly delimited. Later on, we could add more blocks., much like fstab.

LedgerSMB/Database/Loadorder.pm
sbts: Perhaps rather than removing it we should rename it or otherwise alter it to indicate there was a crash and things may not be in a consistent state

sbts:
I'll second that, I think it should be a requirement for the entire project.
although we would need to document that fact well, and do a couple of checks and die loudly if there is a space in the base path.
Normally not an issue on linux systems, but if we have anyone from a windows world trying to install on windows (or linux for that matter) it could become a gotcha that is hard to debug without the noise

LedgerSMB/Database/Change.pm
einhverfr: Yeah.  there really isn't enough at present to justify one.  I might switch and go with Moo(se) if more functionality is added but for now it seemed like too small a data structure to really benefit from it.

LedgerSMB/Database/Change.pm
einhverfr: depends on the Perl version iirc. :-P

LedgerSMB/Database/Change.pm
einhverfr: Good point.  Will get this firmed up a bit.

LedgerSMB/Database/Change.pm
einhverfr: good point, will tighten that up too.

LedgerSMB/Database/Loadorder.pm
einhverfr: Actually planning on removing reload_subsequent.  I think if dependencies need to depend on this, this is way too broad to be useful.  So the data structure is even simpler for now.....

einhverfr:
I think we should require no spaces but would like the loadorder to be stuff that is extrinsic to the file.  I.e. we have to ell it to possibly ignore some failrues because they may have already been applied.  But for things like dependencies, etc. I would like those to be maintained from inside the files eventually.  This means all info is in one place and fewer opportunities to get things wrong.

LedgerSMB/Database/Loadorder.pm
einhverfr: actually we cant do an flock either.  I think we should just do he locking during the load process and have it be an advisory lock.

LedgerSMB/Database/Loadorder.pm
einhverfr: I am going to move this into a database advisory lock.  I cannot think of another safe way to do this, but it means the locking occurs at the actual change loading routine in the loadorder handler.

einhverfr:
ok additional error handling added.  once travis passes, will merge

PR #1355, opened 2016-02-22
Feature/1.5 dojo build

### LINKED ISSUE #904: Need better dojo delivery system
We need a better approach to including dojo.  We don't want to keep this in the repo anymore.

See https://sourceforge.net/p/ledger-smb/bugs/1182/ for past issues with the current approach

freelock:
This is the major PR to close #904 . This PR removes the Dojo source release entirely from the tree, creates a build, and sets the built Javascript as active.

The main thing incomplete here is how to switch back and forth between using Dojo sources and the built tree.

I dropped in a "dojo_built" variable into the template, but actually setting this I wasn't sure how to hook up -- I'm thinking this should be set in LedgerSMB.conf, with the default being "use built" because then you don't need a build step (because the built files are shipped with the project).

If you want to use source files (e.g. if you are working on the Javascript) you will need to pull the new git submodules into your working tree with:

git submodule init
git submodule update

... this is entirely optional for using LedgerSMB, but required if you want to develop the JS. Right now these are pinned at Dojo version 1.10.4.

Then you need to set dojo_built to false -- this is what needs to get fixed, because I've essentially hard-coded it into LedgerSMB/Template/HTML.pm.

Finally, to do a new build, you'll need to have node.js installed, the Dojo source submodules updated.

cd UI/lsmb
../dojo/utils/buildscripts/build.sh --profile lsmb.profile.js

... and then commit the changed code under UI/build.


ehuelsmann:
@freelock I played around with it for a bit. Works great! I'll add a separate issue for the "switching back and forth" which I'll expect to close later today (but others should feel free to pick up :-) ).

Now, I think we need to extend our developer documentation with (pointers to) instructions on how to develop our UI. I'll create an issue for that too.

Issue #1263, opened 2016-01-15
1.4 -- Sales Order regression -- cannot update existing sales order


freelock:
Hi,

We hit (again) the workflow issues creating sales orders and partially shipping them on invoices that have been a problem for a couple of months.

I just set out to create an issue for this based on the tip of 1.4 -- and found that it mostly works! Thanks for getting that squared away.

I'm still seeing one issue though:

1. Create a sales order, add some line items, and save.
2. Search for the sales order using Order Entry -> Reports -> Sales Orders, and re-open
3. Change the contents of the order -- e.g. the qty of the first line item, and update.
4. Click save.

Result: the line item disappears, and the invoice is now $0.

It gets a bit worse:

5. Add a line item back as the first entry.

Error!
 'LedgerSMB::PGNumber Invalid Number at LedgerSMB/PGNumber.pm line 167.

... If you click the back button, and update, you can apparently add a line item successfully -- but if you save and then re-open the sales order, same result as before -- sales order is at $0, no line item.

freelock:
Hi,

Erik asked for an example BDD test for this workflow, including the scenario broken in this issue. Got a jump on this, I don't see the BDD features merged into master yet, so I'm uploading here for now rather than doing a PR... Had to add a .txt so that I could upload here...

[sales-order-invoice-workflow.feature.txt](https://github.com/ledgersmb/LedgerSMB/files/92804/sales-order-invoice-workflow.feature.txt)

... This test is written assuming that all background data will get automatically cleaned up at the end of the test run, and that there's a COA already set up -- would probably need to add accounts at least to the parts background.

It also assumes that this is something you would run repeatedly on the same database, and so anything created needs to be cleaned up so you can re-run the tests (the last scenario will need to do this cleanup).

Lots of ways of creating usable patterns here for the text detection -- I think I used "click on _button text_" whereas the Drupal Extension code uses "click on" for links and "press" for buttons. And I haven't used the DrupalExtension table row selection rules enough to know the syntax they've built, but they do support this type of detecting text inside indexed table rows, or text found within the same table row...

Finally, I built these mostly around the current behavior, aside from what's actually broken. We might want to revisit this as we improve the UX, and simplify some of these steps.

Cheers,
John

freelock:
[...]

PR #1363, opened 2016-02-23
Fixing #1263

einhverfr:
…ber error


bin/oe.pl
```diff
@@ -399,7 +399,9 @@ sub form_header {
 |;
 
         $n = ( $form->{creditremaining} < 0 ) ? "0" : "1";
+<<<<<<< local
```
ehuelsmann: @einhverfr there is a merge marker added here

PR #1368, opened 2016-02-25
* Fix (seemingly correct) syntax and need for exception handling

ehuelsmann:


LedgerSMB/Scripts/contact.pm
```diff
@@ -340,7 +341,7 @@ sub _main_screen {
                    person => $person,
                  employee => $person,
                      user => $user,
-                    roles => eval { $user->list_roles },
```
einhverfr: should be roles =>  [ eval { $user->list_roles } ].

But your changes work too.

LedgerSMB/Scripts/contact.pm
```diff
@@ -340,7 +341,7 @@ sub _main_screen {
...
+                    roles => \@roles,
```
einhverfr: I suspect this isn't a syntax error.  It is rather a syntax/semantics mismatch.  Thanks for fixing!

LedgerSMB/Scripts/contact.pm
ehuelsmann: Right. I did it the way I did, because that reduces the scope of the exception handling suppression (that's basically the effect of ignoring $@)

LedgerSMB/Scripts/contact.pm
ehuelsmann: The error that was generated was "odd number of elements in anonymous hash". No idea if the curlies were parsed as a hash or not, but I suspect they were and the curlies have only 1 element....

Thanks for reviewing and commenting on the original code.

PR #1430, opened 2016-03-30
* Migrate 'unit' and 'repeat' columns to 'recurring_interval'

ehuelsmann:
There is at least 1 known schema which has the old 'unit' and 'repeat' columns on 1.4 and there was no conversion in Fixes.sql. Adding that.

### DIFF
--- a/sql/modules/Fixes.sql
+++ b/sql/modules/Fixes.sql
@@ -368,3 +368,16 @@ values (254, 28, 'module', 'transtemplate.pl'), (255, 28, 'action', 'list');
 DROP INDEX "je_unique_source";
 CREATE UNIQUE INDEX "je_unique_source" ON journal_entry(journal, reference) where journal in (1, 2) and not is_template;
 COMMIT;
+
+
+BEGIN;
+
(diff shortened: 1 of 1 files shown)


ehuelsmann:
The step from 'unit' and 'repeat' to 'recurring_interval' was made in Feb 2013 while 1.4.1 was released in Sept 2014. So, I hope nobody else is affected. Adding this for safety.

einhverfr:
If these columns have data in them, should we migrate the data?

ehuelsmann:
@einhverfr agreed, hence the UPDATE statement here: https://github.com/ledgersmb/LedgerSMB/pull/1430/files#diff-fa4c846c0ee60ea699b838773c1c893eR377

einhverfr:
Sorry missed that.  I think this is a good change.  Given the age, do we need to add a 1.5 upgrade script?

ehuelsmann:
I don't think we need a 1.5 update script: everybody on 1.4 will be updated now and everybody upgrading from 1.3 will now have the right columns anyway...

PR #1454, opened 2016-04-20
Hard check on versions to avoid accidently running 1.4 modules under 1.5

einhverfr:


tools/starman.psgi
```diff
@@ -7,6 +7,8 @@ use LedgerSMB::PSGI;
 use Plack::Builder;
 use Plack::Middleware::Static;
 
+die 'Cannot verify version of libraries, may be including out of date modules?' unless $LedgerSMB::PSGI::VERSION == '1.5';
```
ehuelsmann: @einhverfr any reason not to use the ```$LedgerSMB::VERSION```variable for this? It allows even more exact checking, if we add tools/starman.psgi to the list of files which have a version number encoded; it'll be updated on every release...

einhverfr:
I wanted to support major version checking only, but you are right, using the LedgerSMB version number would probably be better.

einhverfr:
I remember why I hit FCGI.  I have seen some cases where modules in LedgerSMB/ were included but the LedgerSMB.pm was not in the Perl path, so this seemed safer.

ehuelsmann:
There's one concern though: our release scripts automatically update the version numbers in a number of files, but not the one in FCGI and starman.psgi; nor do they update any 'minor version' numbers: they update the full patch version number.

How do we ensure the number here stays current?

(Note of concern: our "cookie" in the LedgerSMB config file is still called "LedgerSMB-1.3" and the default database is still called lsmb13, in the same file...)

jame:
> (Note of concern: our "cookie" in the LedgerSMB config file is still called "LedgerSMB-1.3" and the default database is still called lsmb13, in the same file...)

Yes, and I've been meaning to ask about why it is that they haven't been updated for 1.4 (and now, for 1.5...)  In the case the 1.4.x Debian packages, I have them patched to reflected the proper version number and plan the same for the v1.5.x packages...

ehuelsmann:
@jame Right. Too many places where version numbers are referenced and no good release instructions (or we do, but they were not followed) where all places are listed, would be my answer. "too much to do, too little time" would be another reason. I'm trying to automate as much as we can in the release process because of that. If you find more places that need to be updated, please let me know, so I can add those to the update scripts.

Even with that in place, there's still a lot that remains to be done manually (thinking about installation documentation and the ledgersmb.org site).

In general I'm hesitant to add more like these.

@einhverfr I saw you merged, but didn't switch the module reference to ```$LedgerSMB::Form::VERSION``, which is at the same directory level as PSGI.pm and already has the version number. Was there a particular reason not to?

PR #1470, opened 2016-04-29
Update INSTALL.md and p-c-d.pl

ehuelsmann:


ehuelsmann:
This PR is incomplete: it needs to have tests implemented so we know the script minimally works.

PR #1474, opened 2016-04-30
Fix creation of lsmb_dbadmin, and include step to install cpan deps

sbts:
Add new function CPAN_InstallPackages(), 
add missing -e to many echo statements, 
fix creation of lsmb_dbadmin and add additional features / User info

### DIFF
--- a/utils/install/lsmb-install.sh
+++ b/utils/install/lsmb-install.sh
@@ -329,33 +329,238 @@ SelectVersion() {
     popd >/dev/null
 }
 
+createPostgresSuperUser() {
+    echo
+    echo "do not use this function [ createPostgresSuperUser() ]"
+    echo
+    exit 999;
+    # First parameter is the user name
+    LSMBDBUSER=$1
+    # Second parameter is the password
+    LSMBDBPW=$2
+
+    su - postgres -c psql <<-EOT
+       DO
+       \$\$
+       DECLARE num_users integer;
+       BEGIN
+           SELECT count(*)
+               into num_users
+           FROM pg_user
+           WHERE usename = '$LSMBDBUSER';
+
+           IF num_users = 0 THEN
+               CREATE ROLE $LSMBDBUSER WITH SUPERUSER LOGIN NOINHERIT ENCRYPTED PASSWORD '$LSMBDBPW';
+           ELSE
+               ALTER ROLE $LSMBDBUSER WITH SUPERUSER LOGIN NOINHERIT ENCRYPTED PASSWORD '$LSMBDBPW';
+           END IF;
+       END
+       \$\$
+       ;
+       EOT
+}
+
 SetupPostgres() {
 
-#    echo "${Div}${Div}"
-#    echo "Configuring Postgres accessibility."
-#    echo "${Div}"
-#    echo " this is only needed if you want to directly connect to the database with management tools on other computers"
-#    GetKey yN "\n${Div} Enable Connections from other machines?\n${Div}this is only needed if you want to directly connect to the database with management tools on other computer\n${Div}Postgres: Allow remote access?"
-#    fixme: need to set Listen_Addresses in postges.conf
-    
-    echo "${Div}${Div}"
+    echo -e "${Div}${Div}"
+    echo "Configuring Postgres access for user lsmb_dbadmin."
+    echo -e "${Div}"
+    if (( `find /etc/postgresql -name postgresql.conf | grep -c '$'` == 1 )); then
+        echo -e "\tmodifying pg_hba.conf"
+        sudo sed -i.bak1 -r '
+            /local[[:space:]]*all[[:space:]]*lsmb_dbadmin*/ d
+            /local[[:space:]]*all[[:space:]]*all*/ ilocal\tall\t\tlsmb_dbadmin\t\t\t\tmd5
+          ' `find /etc/postgresql -name pg_hba.conf`
+    else
+        echo '%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%';
+        echo '%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%';
+        echo '%%   ERROR ERROR ERROR  ERROR ERROR ERROR   %%';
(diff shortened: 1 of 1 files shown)


ehuelsmann:
@sbts these are my review remarks:

 1. The script asks for my sudo password without explaining why
 2. Why are mc, htop, ssmtp and avahi-utils required (listed in "essential utilities")?
 3. The one that's extremely required (git) isn't listed? (But it does get installed? same as 'make' and 'rsync'?)
 4. Then, the same utilities are listed under "Base utilities"? (still no git?)
 5. After cloning master, why would I still need to pull it? (There's no "n" option?)
 6. After selecting 5(1.4 branch) and 12 (1.4.28 tag), I see "You have checked out version '' "; looks wrong
 7. Followed by 'on branch master', but I selected 1.4.28? Which is *on* the 1.4 branch?
 8. 'make' is listed as a Perl Package (?)
 9. libpgobject-simple-role-perl and libpgobject-util-dbmethod-perl are listed as PostgreSQL packages (yet are Perl packages)
10. There's no listing of missing packages which are going to be installed from CPAN; I think that's the kind of feedback a user (I) would want
11. The closing screen lists ```pushd ~/"src/git/LedgerSMB"``` which -while it works- looks awkward
12. When starting starman (without installing the missing CPAN packages, mind you), it crashes on a missing PGObject::Type::BigFloat
13. The above indicates that I *definitely* didn't get 1.4.28
14. Why do you recommend ```-l :8080``` over ```-p 8080``` or even Starman's default ```-p 5000``` ?

sbts:
Hi Erik,
    
    On 04/05/16 04:51, Erik Huelsmann
      wrote:
    
    
      @sbts these are my review remarks:
      
        The script asks for my sudo password without explaining why
      
    
    Oops, I'll add an explanatory bit of text in the one place it was
    missed. 
    
      
        Why are mc, htop, ssmtp and avahi-utils required (listed in
          "essential utilities")?
      
    
    I've already broken these out into suggested utils (or some similar
    phrase)
    
      
        The one that's extremely required (git) isn't listed? (But
          it does get installed? same as 'make' and 'rsync'?)
      
    
    I'll see if I can improve the semantics here. I moved git to be auto
    installed if it was missing before it gets first called. Although it
    probably should be in the essential tools/utils group.
    
      
        Then, the same utilities are listed under "Base utilities"?
          (still no git?)
        
      
    
    Yep, git should be in here.
    
      
        After cloning master, why would I still need to pull it?
          (There's no "n" option?)
      
    
    Yep, I'll put in a bypass if a clone was done.
    
      
        After selecting 5(1.4 branch) and 12 (1.4.28 tag), I see
          "You have checked out version '' "; looks wrong
        Followed by 'on branch master', but I selected 1.4.28? Which
          is on the 1.4 branch?
      
    
    Hmm, odd, I'll check on that, it will just be incorrect use of a
    pair of vars in the display stage.
    I probably need to refactor that area to make it more maintainable
    anyway.
    
      
        'make' is listed as a Perl Package (?)
        
      
    
    It's included as a Perl Package, as it's only being installed to
    satisfy the perl dep requirements.
    But you are correct, I'll move it to essentials.
    
      
        libpgobject-simple-role-perl and
          libpgobject-util-dbmethod-perl are listed as PostgreSQL
          packages (yet are Perl packages)
        
      
    
    Probably rather than PostgreSQL package I really meant PostgreSQL
    dependencies.
    However, as our code depends on those perl modules but doesn't
    depend on the posgresql server being installed,
    I'll move the perl modules into Perl Packages and only have the
    postgres server packages here.
    I can then change the description to indicate it's the "PostgreSQL
    Server Packages"
    
      
      
      
        There's no listing of missing packages which are going to be
          installed from CPAN; I think that's the kind of feedback a
          user (I) would want
      
    
    I'll have a think about the best way to do this. It will probably be
    using the new test::dependencies output I guess.
    
      
        The closing screen lists pushd ~/"src/git/LedgerSMB"
          which -while it works- looks awkward
      
    
    Yep, can change that to use $HOME instead of ~, or simply move the
    quote. 
    I was already considering switching to $HOME in the code for
    readability anyway
    
      
        When starting starman (without installing the missing CPAN
          packages, mind you), it crashes on a missing
          PGObject::Type::BigFloat
      
    
    Yep, need to install CPAN packages to fix that,
    PGObject::Type::BigFloat wasn't in the deb repos
    
      
        The above indicates that I definitely didn't get
          1.4.28
      
    
    Strange. I'll check that again.
    
      
        Why do you recommend -l :8080 over -p
            8080 or even Starman's default -p 5000
          ?
      
    
    QUOTE from the starman man page
           --port 8080
               Specifies the port to bind.
               This option is for a compatibility with
      plackup and you're recommended to use "--listen" instead.
    
    -l is the same as --listen.  The port chosen is arbitrary, and just
    what I have setup here. I generally would suggest not using the
    default port for something like this, it introduces more chance of a
    clash if someone runs starman for some other reason.
    We probably should come up with a "standard port range" to run our
    starman processes on, much as postgres uses 5432 to (5432 +n)
    Perhaps we could use 6543 to (6543 +n)
    
    
    
      
      
      
      
      
      
      
      
      —
        You are receiving this because you were mentioned.
        Reply to this email directly or view
          it on GitHub

jame:
David,

On Wednesday, May 04, 2016 06:35:14 AM David wrote:
> 
>         When starting starman (without installing the missing CPAN
>           packages, mind you), it crashes on a missing
>           PGObject::Type::BigFloat
> 
>     Yep, need to install CPAN packages to fix that,
>     PGObject::Type::BigFloat wasn't in the deb repos
> 

  Note that libpgobject-type-bigfloat-perl has been in Debian since December. and is now in 'stretch' ("Testing") and that also means it was sync'ed to Ubuntu soon enough to have made it to xenial. (16.04 LTS).

  So it rather depends on which distro this is running on...

ehuelsmann:
@sbts Responding to Jame's point: are we going to make this script maintain distro-specific lists? (Or maybe we can find some way around that, like listing all packages by name and checking if the "current distro" provides it?)

Regardless, I think this script is already in the repo and we have improved it big time here, so I think we should merge regardless of the comment above. lets move that to a separate issue instead.

sbts:
I'll break the distro specific bits out into separate files, but where possible support auto detection of available distro packages to prefer them over cpan

PR #1476, opened 2016-04-30
Extended sync-fork.sh for better behaviour

sbts:
[ci skip]

### DIFF
--- a/utils/devel/sync-fork.sh
+++ b/utils/devel/sync-fork.sh
@@ -46,6 +46,14 @@ fi
 
 CurrentBranch=`git rev-parse --abbrev-ref HEAD`
 
+# using stash create may cause issues if we fail during the other git operations, 
+# as when we re-run this script we won't know what the stashID is 
+# so won't be able to revert to the original state of the repo
+# at least if we use "stash save --all" a manual pop would be enough to restore state
+#stash=`git stash create`
+unset stash
+git stash save --all "automatic stash while running 'sync-fork.sh'"
+
 git checkout master || Error "Checking out Master."
 
 git fetch upstream || Error "Fetching Upstream."
@@ -54,6 +62,7 @@ git merge upstream/master || Error "Merging upstream/master with local /master"
 
 git checkout "$CurrentBranch" || Error "Checking out $CurrentBranch."
 
+git stash pop $stash
 
 cat <<EOF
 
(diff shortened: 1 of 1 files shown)


utils/devel/sync-fork.sh
```diff
@@ -42,14 +44,15 @@ if ! [[ $AvailableRemotes =~ $Upstream ]]; then
...
 
-git fetch upstream || Error "Fetching Upstream."
-
+CurrentBranch=`git rev-parse --abbrev-ref HEAD`
 
 git checkout master || Error "Checking out Master."
 
+git fetch upstream || Error "Fetching Upstream."
 
 git merge upstream/master || Error "Merging upstream/master with local /master"
```
ehuelsmann: @sbts I think this merges into the active branch, not into the master branch, right?

sbts: Hi Erik,
    
    Well sort of.
    It merges into the active branch, but as we start out by checking
    out master, it effectively merges into master.
    
    We then checkout the original active branch and warn the user that
    they will need to do a merge.
    We can't really do the merge on the active branch automatically as
    it's difficult / impossible to know what branch they may want to
    base from.
    
    Regards
    David G
    
    On 04/05/16 04:54, Erik Huelsmann
      wrote:
    
    
      In utils/devel/sync-fork.sh:
      >  
>  git checkout master || Error "Checking out Master."
>  
> +git fetch upstream || Error "Fetching Upstream."
>  
>  git merge upstream/master || Error "Merging upstream/master with local /master"

      @sbts I think this merges into the
        active branch, not into the master branch, right?
      —
        You are receiving this because you were mentioned.
        Reply to this email directly or view
          it on GitHub

utils/devel/sync-fork.sh
```diff
@@ -42,14 +44,15 @@ if ! [[ $AvailableRemotes =~ $Upstream ]]; then
     git remote add upstream "$Upstream"
 fi
 
-git fetch upstream || Error "Fetching Upstream."
-
+CurrentBranch=`git rev-parse --abbrev-ref HEAD`
 
 git checkout master || Error "Checking out Master."
```
ehuelsmann: @sbts I now see that you're checking out master here first, which is good. However, you can't always check out master: an error will be thrown when the current working directory isn't clean and conflicts arise. Better stash before checkout and then ```stash pop``` after returning to the branch.

PR #1547, opened 2016-05-16
1.4 multi comparisons

ehuelsmann:


ehuelsmann:
We're trying to get the same functionality merged into 1.5, preferably parallel to merging this one.

PR #1652, opened 2016-06-14
Sl30 migration fixes

ylavoie:
Fix upgrade failure on Transaction Template menu privilege and set the default sequences numbers which weren't assigned in SL30 migration


ehuelsmann:
Replaced by #1656 and #1657. Closing.

PR #1653, opened 2016-06-14
Documentation toys

ylavoie:
Web base documentation.
- Generate html pages from pod files
- Generate a summary of critiques from source code. Original idea from http://blogs.lessthandot.com/index.php/webdev/perl/create-html-from-output-of/


tools/starman.psgi
```diff
@@ -27,6 +28,11 @@ builder {
     enable match_if path(qr!.+\.(css|js|png|ico|jp(e)?g|gif)$!),
         'ConditionalGET';
 
+    enable "Plack::Middleware::Pod",
```
ehuelsmann: @ylavoie above, the Plack::Middleware::Pod module is required conditionally, yet the 'enable' statement here seems to be executed unconditionally. Does that work?

PR #1654, opened 2016-06-14
First attempt at html critique

ylavoie:
Web base documentation.
- Generate html pages from pod files
- Generate a summary of critiques from source code. Original idea from http://blogs.lessthandot.com/index.php/webdev/perl/create-html-from-output-of/

### DIFF
--- a/t/01-html-checks.t
+++ b/t/01-html-checks.t
@@ -1,0 +1,106 @@
+#!perl
+
+use strict;
+use warnings;
+
+use File::Find;
+use Test::More;
+eval { use HTML::Lint; 1 } or plan skip_all => "HTML::Lint not available";
+
+my @on_disk = ();
+my $is_todo = '';
+
+sub collect {
+    return if $File::Find::name !~ m/\.html$/ || $File::Find::name =~ m/\/pod\//;
+
+    my $module = $File::Find::name;
+    push @on_disk, $module
+}
+find(\&collect, 'UI/');
+
+sub content_test {
+    my ($filename) = @_;
+    my $ui_header_used = 0;
+    $ui_header_used = 1 if $filename =~ m/UI\/lib\//;
+
+    my ($fh, @tab_lines, @trailing_space_lines);
+    open $fh, "<$filename";
+    while (<$fh>) {
+        push @tab_lines, ($.) if /\t/;
+        push @trailing_space_lines, ($.) if / $/;
+        $ui_header_used = 1 if /ui-header\.html/;
+    }
+    close $fh;
+
+TODO: {
+    local $TODO = 'LedgerSMB HTML files is still under development';
+    $is_todo = Test::More->builder->todo;
+    ok((! @tab_lines) && (! @trailing_space_lines),
+        "Source critique for '$filename'");
+    diag("Line# with tabs: " . (join ', ', @tab_lines))
+        if @tab_lines;
+    diag("Line# with trailing space(s): " . (join ', ', @trailing_space_lines))
+        if @trailing_space_lines;
+
+    my $lint = HTML::Lint->new;
+    $lint->only_types(); # Get all
+
+    # Add dojo attributes
+
+    for my $dojo ( 'data-dojo-attach-event',
+                   'data-dojo-attach-point',
+                   'data-dojo-id',
+                   'data-dojo-mixins',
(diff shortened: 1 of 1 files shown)


cpanfile
```diff
@@ -16,6 +16,7 @@ requires 'Digest::MD5';
 requires 'File::MimeInfo';
 requires 'HTTP::Exception'; # YLA
 requires 'JSON';
+requires 'HTML::Lint'; # YLA t/01-html-checks.t
```
ehuelsmann: Can you move this requirement to the development section or its own feature?

t/01-html-checks.t
```diff
@@ -0,0 +1,60 @@
+#!perl
+
+use strict;
+use warnings;
+
+use File::Find;
+use Test::More;
+use HTML::Lint;
```
ehuelsmann: and then 
```perl
eval { use HTML::Lint; 1 } or plan skip_all => "HTML::Lint not available";
```
in this section to make this test optional.

t/01-html-checks.t
```diff
@@ -0,0 +1,106 @@
...
+        $ui_header_used = 1 if /ui-header\.html/;
+    }
+    close $fh;
+
+TODO: {
+    local $TODO = 'LedgerSMB HTML files is still under development';
+    $is_todo = Test::More->builder->todo;
+    ok((! @tab_lines) && (! @trailing_space_lines),
+        "Source critique for '$filename'");
+    fail("Line# with tabs: " . (join ', ', @tab_lines))
```
ehuelsmann: @ylavoie I'm thinking this fail() should probably be a [diag()](http://search.cpan.org/~exodist/Test-Simple-1.302031/lib/Test/More.pm#Diagnostics) ?

(Same with the fail 2 lines below.)

t/01-html-checks.t
```diff
@@ -0,0 +1,106 @@
...
+    while (<$fh>) {
+        push @tab_lines, ($.) if /\t/;
+        push @trailing_space_lines, ($.) if / $/;
+        $ui_header_used = 1 if /ui-header\.html/;
+    }
+    close $fh;
+
+TODO: {
+    local $TODO = 'LedgerSMB HTML files is still under development';
+    $is_todo = Test::More->builder->todo;
```
ehuelsmann: From the docs of [Test::More's TODO](http://search.cpan.org/~exodist/Test-Simple-1.302031/lib/Test/More.pm#TODO), I understood that the ```TODO: {}``` block and ```local $TODO = ``` steps should be enough to make the tests marked with "allowed fail". Or is the purpose of the ```->todo``` call different?

t/01-html-checks.t
```diff
@@ -0,0 +1,106 @@
...
+                      'table', 'tbody', 'td', 'textarea', 'th', 'tr', 'ul' ) {
+            HTML::Lint::HTML4::add_attribute( $tag, $dojo );
+        }
+    }
+    HTML::Lint::HTML4::add_attribute( 'input', 'constraints' );
+    HTML::Lint::HTML4::add_attribute( 'textarea', 'placeholder' );
+    HTML::Lint::HTML4::add_attribute( 'textarea', 'value' );
+    HTML::Lint::HTML4::add_attribute( 'textarea', 'width' );
+
+    # Add moose attributes
```
ehuelsmann: @ylavoie actually, the 'role' attribute is somethnig Dojo uses too. it seems to come from "aria" or "wia-aria"; a standard to help "assistive technologies"...

t/01-html-checks.t
```diff
@@ -0,0 +1,106 @@
...
+                   'data-dojo-id',
+                   'data-dojo-mixins',
+                   'data-dojo-obj',
+                   'data-dojo-props',
+                   'data-dojo-textdir',
+                   'data-dojo-type',
+                   'data-myscope-id',
+                   'data-myscope-props',
+                   'data-myscope-type') {
+        for my $tag ( 'body', 'button', 'center', 'div', 'fieldset', 'form', 'h1', 'h2', 'h3', 'h4',
```
ehuelsmann: @ylavoie from what I know about HTML5 (which isn't all that much), I think the idea is to attach all non-semantic information to ```data-*``` named attributes, so any ```data-``` attribute is valid. Is there a syntax to declare that?

PR #1697, opened 2016-06-28
Fix #1696 Missing Display of Username and Password expiry in preferen…

### LINKED ISSUE #1696: Username and password expiry not displayed in header on preferences page
This line from preferences.html renders without the username
```
<tr><th class="listtop"><?lsmb text('Preferences for [_1]', login) ?></th></tr>
```
and password expiry time doesn't render from this line either
```
          <?lsmb text('Your password will expire in [_1]', password_expires) ?>
```

sbts:
…ces.html

Is translating user.password_expires sane on line 23?
it is a string that will vary from day to day, and will normally contain at least one number.
I can't imagine that we can translate it as a static phrase, rather it would need to be translated as individual words (I *think*)

### DIFF
--- a/UI/users/preferences.html
+++ b/UI/users/preferences.html
@@ -20,7 +20,9 @@
           <?lsmb text('Preferences Saved') ?>
   </tr><?lsmb END # if action ?>
   <tr><th class="info">
-          <?lsmb text('Your password will expire in [_1]', user.password_expires) ?>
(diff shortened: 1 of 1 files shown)


sbts:
Hmm, but your proposed solution (ignoring the fact that the IF should be user.password_expires) won't work as expected as you would end up with things like
> will expire in 10 days days

Also, it's possible that there would be more than just hours and days to consider.
Perhaps months and years as well

PR #1728, opened 2016-07-15
Sl30 reconciliations working

ylavoie:


lib/LedgerSMB/DBObject/Reconciliation.pm
```diff
@@ -162,7 +162,9 @@ sub approve {
     # the user should be embedded into the $self object.
     my $report_id = shift @_;
 
-    my $code = $self->call_dbmethod(funcname=>'reconciliation__report_approve', args=>[$report_id]); # user
+    # Even with https://github.com/ledgersmb/LedgerSMB/issues/838 fixes, array parameters DO NOT WORK
+    #my $code = $self->call_dbmethod(funcname=>'reconciliation__report_approve', args=>[$report_id]); # user
+    my $code = $self->call_dbmethod(funcname=>'reconciliation__report_approve', args=> { in_report_id => $report_id }); # user
```
ehuelsmann: @ylavoie in the original code, it should have said:

```perl
   my $code = $self->call_procedure(....);
```
instead of 
```perl
   my $code = $self->call_dbmethod(....);
```
In that case the `args=>[]` would have worked; I don't think `args=>[]` will ever work for `call_dbmethod()`. (I don't see how to make it work)

lib/LedgerSMB/Scripts/recon.pm
```diff
@@ -275,34 +276,23 @@ sub _display_report {
...
-                $recon->{total_uncleared_credits}->to_output(money => 1);
+    $recon->{mismatch_our_debits} = $recon->{mismatch_our_debits}->to_output(money => 1);
+    $recon->{mismatch_our_credits} = $recon->{mismatch_our_credits}->to_output(money => 1);
+    $recon->{mismatch_their_debits} = $recon->{mismatch_their_debits}->to_output(money => 1);
+    $recon->{mismatch_their_credits} = $recon->{mismatch_their_credits}->to_output(money => 1);
+    $recon->{statement_gl_calc} = $recon->{statement_gl_calc}->to_output(money => 1);
+    $recon->{total_cleared_debits} = $recon->{total_cleared_debits}->to_output(money => 1);
+    $recon->{total_cleared_credits} = $recon->{total_cleared_credits}->to_output(money => 1);
+    $recon->{total_uncleared_debits} = $recon->{total_uncleared_debits}->to_output(money => 1);
+    $recon->{total_uncleared_credits} = $recon->{total_uncleared_credits}->to_output(money => 1);
```
ehuelsmann: @ylavoie if you're refactoring this anyway, can we do something to make it more concise, like:
```perl
    for my $amt_name (qw/ mismatch_our_ mismatch_their total_cleared_ total_uncleared_ /) {
      for my $bal_type (qw/ credits debits/) {
         $recon->{"$amt_name$bal_type"} = $recon->{"$amt_name$bal_type"}->to_output(money=>1);
      }
   }
```

sql/modules/Date_Formats.sql
```diff
@@ -0,0 +1,506 @@
+CREATE TABLE date_formats (
+  language text check (language ~ '[[:alpha:]_]') NOT NULL,
+  country text check (country ~ '[[:alpha:]_]') NOT NULL,
+  format text check (format ~ '[[:graph:]_]') NOT NULL
+);
+CREATE UNIQUE INDEX date_formats_key ON date_formats(language,country);
```
ehuelsmann: I think this should be the table's primary key?

sql/modules/Date_Formats.sql
```diff
@@ -0,0 +1,506 @@
...
+BEGIN
+  IF ((SELECT COUNT (C.short_name) FROM Country C
+      WHERE C.short_name = NEW.Country) = 0) THEN
+    RAISE EXCEPTION 'Unknown country %', NEW.country;
+  END IF;
+  RETURN NEW;
+END;
+$$ LANGUAGE PLPGSQL;
+
+CREATE TRIGGER date_formats_trigger
```
ehuelsmann: This trigger can probably be replaced with a REFERENCES construct; it makes PG check existence in the other table automatically.

sql/modules/Date_Formats.sql
```diff
@@ -0,0 +1,506 @@
+CREATE TABLE date_formats (
```
ehuelsmann: This is a DDL statement which belongs in the sql/changes/ folder, because otherwise it will be executed every time modules are being reloaded. (Hmmm, come to think of that, we probably need tests for it.)

sql/modules/Date_Formats.sql
```diff
@@ -0,0 +1,506 @@
...
+END;
+$$ LANGUAGE PLPGSQL;
+
+CREATE TRIGGER date_formats_trigger
+   BEFORE INSERT ON date_formats
+FOR EACH ROW EXECUTE PROCEDURE date_formats__check();
+
+-- See CPAN DateTime::Locale::Catalog
+INSERT INTO date_formats (language,country,format)
+VALUES
```
ehuelsmann: Very nice! But we also want to load these only once.

So, probably the best place to put them is in the DDL file which creates the initial table.

sql/modules/Date_Formats.sql
```diff
@@ -0,0 +1,506 @@
+CREATE TABLE date_formats (
+  language text check (language ~ '[[:alpha:]_]') NOT NULL,
+  country text check (country ~ '[[:alpha:]_]') NOT NULL,
+  format text check (format ~ '[[:graph:]_]') NOT NULL
+);
```
ehuelsmann: Here, I think we need a

```SQL
COMMENT ON TABLE date_formats IS $$
<documentation of the table>
$$;

File: t/43-dbtest.t [hunk 8]
```diff
@@ -5,7 +5,7 @@ if (!defined $ENV{LSMB_TEST_DB}){
         plan skip_all => 'Skipping all.  Told not to test db.';
 }
 else {
-        plan tests => 470;
+        plan tests => 465;
```
ehuelsmann: The only thing that changed is the number of tests? No tests added/removed?

sql/modules/Robot.sql
```diff
@@ -0,0 +1,126 @@
...
+ WHERE e.control_code = $1;
+$$ LANGUAGE SQL;
+
+DROP FUNCTION IF EXISTS robot__save (int, text, text, text, int);
+CREATE OR REPLACE FUNCTION robot__save
+(in_entity_id integer,
+in_first_name text, in_middle_name text, in_last_name text,
+in_country_id integer
+)
+RETURNS INT AS $$
```
ehuelsmann: Hmm. Why would a robot have names, not just a description?

sql/modules/Reconciliation.sql
```diff
@@ -383,10 +391,10 @@ create or replace function reconciliation__add_entry(
 
                 END IF;
         END IF;
-        return 1;
+        return lid;
 
     END;
-$$ language 'plpgsql';
+$$ language 'plpgsql' VOLATILE;
```
ehuelsmann: VOLATILE is the default (https://www.postgresql.org/docs/current/static/xfunc-volatility.html)

sql/changes/1.5/open_forms_callers.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+ALTER TABLE public.open_forms ALTER COLUMN id SET DEFAULT floor(random()*(1000000))+1;
+ALTER TABLE public.open_forms ADD COLUMN caller1 character varying(100);
+ALTER TABLE public.open_forms ADD COLUMN caller2 character varying(100);
+ALTER TABLE public.open_forms ADD COLUMN caller3 character varying(100);
+ALTER TABLE public.open_forms ADD COLUMN caller4 character varying(100);
+ALTER TABLE public.open_forms ADD COLUMN last_used timestamp without time zone;
```
ehuelsmann: Do we want this in production? I mean, it's a great debugging tool, obviously; and caller1 to 5 make great input into a hash, but...

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -757,18 +756,27 @@ sub fix_tests{
 
     my $table = $request->{dbh}->quote_identifier($request->{table});
     my $edit = $request->{dbh}->quote_identifier($request->{edit});
-        my $where = $request->{id_where};
-    my $sth = $request->{dbh}->prepare(
-            "UPDATE $table SET $edit = ? where $where = ?"
+    my $where = $request->{id_where};
```
ehuelsmann: `quote_identifier()` missing?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -757,18 +756,27 @@ sub fix_tests{
...
-            $request->error($sth->errstr);
+        $sthu->execute($request->{"$request->{edit}_$id"}, $id) ||
+        $request->error($sthu->errstr);
+#        my $status = $sthu->execute($request->{"$request->{edit}_$id"}, $id);
+#        if ( $status == 0 ) {
+#            $sthi->execute($request->{"$request->{edit}_$id"}, $id) ||
+#            $request->error($sthi->errstr);
+#        } elsif ( $status < 0 ) {
+#            $request->error($sthu->errstr);
+#        }
```
ehuelsmann: Can this commented out code be eliminated? Or is there a reason it should remain?
(Note: I'm not a fan of having commented out code)

sql/Pg-database.sql
```diff
@@ -487,6 +499,7 @@ INSERT INTO country(short_name,name) VALUES ('VN','Viet Nam');
 INSERT INTO country(short_name,name) VALUES ('VU','Vanuatu');
 INSERT INTO country(short_name,name) VALUES ('WF','Wallis and Futuna Islands');
 INSERT INTO country(short_name,name) VALUES ('WS','Samoa');
+INSERT INTO country(short_name,name) VALUES ('XK','Kosovo');
```
ehuelsmann: Pg-database isn't loaded anymore on existing installs. Do we want these in existing installs too? Then I think we need to add a schema-change for them. (Those *are* being run on existing schemas.)

sql/Pg-database.sql
```diff
@@ -546,7 +559,9 @@ VALUES (1,'Vendor'),
        (5,'Lead'),
        (6,'Referral'),
        (7,'Hot Lead'),
-       (8,'Cold Lead');
+       (8,'Cold Lead'),
+       (9,'Sub-contractor'),
+       (10,'Robot');    -- Software robot for automation of user-based tasks, Migration reconciliation approval authority, for example
```
ehuelsmann: Same comment here about existing schemas.

sql/Pg-database.sql
```diff
@@ -743,6 +758,18 @@ CREATE TABLE person (
...
+    id serial PRIMARY KEY,
+    entity_id integer references entity(id) not null,
+    first_name text check (first_name ~ '[[:alnum:] _\-\.\*]?'),
+    middle_name text,
+    last_name text check (last_name ~ '[[:alnum:] _\-\.\*]') NOT NULL,
+    created date not null default current_date,
+    unique(entity_id)
+ );
+
+COMMENT ON TABLE robot IS $$ Every robot, must have an entity to derive a common or display name. The correct way to get class information on a robot would be robot.entity_id->entity_class_to_entity.entity_id. $$;
```
ehuelsmann: Here too; basically, we don't want to change Pg-database for a (long?) while. It's a DDL change, so, goes into sql/changes/.

sql/modules/Reconciliation.sql
```diff
@@ -275,8 +270,13 @@ create or replace function reconciliation__add_entry(
         SELECT value into t_prefix FROM defaults WHERE setting_key = 'check_prefix';
 
         t_uid := person__get_my_entity_id();
+        IF t_uid IS NULL THEN
+                t_uid = robot__get_my_entity_id();
+        END IF;
```
ehuelsmann: How about a unified "entity__get_my_entity_id()" ?

lib/LedgerSMB/DBObject/Reconciliation.pm
ylavoie: 1. Then all this module would benefit from the upgrade, I am still seeing mixed uses of call_procedure & call_dbmethod.
2. Positional parameters open the door to future problems if function interface changes.

sql/modules/Date_Formats.sql
ylavoie: File removed from current PR. I am reconsidering the method.

lib/LedgerSMB/DBObject/Reconciliation.pm
ehuelsmann: @ylavoie There might be a need to have a look at the entire module. I'll schedule that for tomorrow. Although positional parameters indeed do require tight coupling, I think that's the point here: the db function and the module and the sproc *are* tightly coupled in the sense that the module provides a Perlish wrapping interface to the stored proc.

t/43-dbtest.t
ylavoie: Using random form_id key instead of sequential. (9a5d554) dropped the amount of tests because the form id is now generated automatically by the database. I think that session ids should use the same method.

sql/modules/Robot.sql
ylavoie: Because they fit in the same table as others.

sql/modules/Reconciliation.sql
ylavoie: Fixed.

sql/changes/1.5/open_forms_callers.sql
ylavoie: What we really want is the name of the form.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Unrelated modification and not ready. Removed.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Unrelated modification and not ready. Removed.

sql/Pg-database.sql
ylavoie: Will be adjusted. Moved to localized-setup branch

sql/Pg-database.sql
ylavoie: Moved into schema changes

sql/modules/Reconciliation.sql
ylavoie: Accepted. Postponed.

sql/changes/1.5/open_forms_callers.sql
einhverfr: I am also unclear as to why this change.  The records get deleted from the table when the session goes away or when the form is closed, so would this not be better handled by logging if it is for debugging?

t/43-dbtest.t
einhverfr: I like the random form_id keys.

Session ids use a random number to generate the token.  The sequential portion is just for foreign key purposes.  I think changing that is not worth the possibility of accidental collision.

einhverfr:
BTW, the complexity of this code review makes things rather difficult to review.  For future pr's it would be really good to try to keep the scope clear, definite, and one topic to a PR.  Otherwise it is really very difficult to see what is going on, and harder still to discuss it.

lib/LedgerSMB/DBObject/Reconciliation.pm
ehuelsmann: @ylavoie I'm not sure what happened; I thought you said in #ledgersmb that you removed this edit, in favor of calling `call_procedure` (instead of `call_dbmethod`).
But the change is still in the diff?

einhverfr:
(For example, the form_id issues could be broken out into a separate PR for easier discussion and a note added as to what, if anything changed, during the whitespace cleanup of the templates).

lib/LedgerSMB/DBObject/Reconciliation.pm
ylavoie: I was about to, but call_procedure is using positional parameters and I have a problem with that.

ylavoie:
My intent was to have them on separate PR, they are on different commits. But PR grabbed all and squashed them in a huge chunk. Won't happen again

t/43-dbtest.t
ylavoie: I used a number big enough to lower collisions possibilities but we'll have to think more carefully if we want to use the same method for session ids

sql/modules/Robot.sql
ylavoie: This is only the first draft for automation robots. It will have to evolve toward a more developed implementation with authorization, revoking, expiration, scope of action...

einhverfr:
No worries.  BTW if you have a few commits to break off into a separate PR at some point one of us can show you how to do that.

I have had a few of these cases where everything ends up in one PR myself.

ylavoie:
I'll have a script set to create branches from each cherry-picked commit soon.
The problem is that many ideas emerge at the same time or from another and they tend to materialize concurrently. Reviewers do not share my thought patterns though and commits have to be published independently. A script will handle that.

lib/LedgerSMB/DBObject/Reconciliation.pm
ehuelsmann: Well, if we use that line of reasoning, we should write of most of our code base :-)

I think using positional parameters is quite al right here: the stored procedure is "part" of the same module, just coded in a different language.

ylavoie: Fixed.

sql/changes/1.5/open_forms_callers.sql
ylavoie: Reverted. When I start thinking about that, I was expecting this to have a longer life. Having the current opened form name could have been usefull for debugging or error logging.

einhverfr:
I think we are getting close on this one ::-)  As soon as tests pass, let's merge.

As a note on positional vs named parameters....   In this case I have no strong preference, but it is worth noting a couple things.

Named parameters are usually preferred because it means things don't break when we add more parameters.  However where we have a single parameter which is an identifier, positional parameters were originally used because we didn't initially support the naming overrides.

In those cases (things like getting a record based on an ID, or the like) the chance of changing the parameter list is pretty minor.  I suppose positional parameters avoid one additional db round trip, but that isn't a great reason to use them.  The larger issue is that the API only makes sense with a single argument.  So, I think thy may be clearer.

So in that case I have no strong preference.  I just wanted to note the historical reasons and issues on both sides.

einhverfr:
Also if we have a ticket for the issue this one fixes, we should set the milestone to 1.5-rc1.  Doing this on the PR as well.

PR #1729, opened 2016-07-15
fixes error during printing orders

einhverfr:


einhverfr:
Note I think orders still fail to print all info.  But now they don't give errors so presumably what is left is a template problem.

PR #1741, opened 2016-07-16
fixing password reset issues

einhverfr:


ehuelsmann:
@einhverfr there's a problem here: the "Change Password" button also saves the preferences, but there's also a preference saving button. I think the fact that the prefs will be saved is either unintentional or too confusing.

PR #1743, opened 2016-07-17
Fixes #1472

### LINKED ISSUE #1472: 1.5: AR/AP Transaction -> Print errors on HTML and CSV formats
The code behind the AR/AP transaction Print button doesn't support other formats than PDF and Postscript, even though the dropdown also shows HTML and CSV...

einhverfr:


einhverfr:
Hmmm this apparently is not needed for 1.4 bcause we added the create_links call in 1.5.  However it adds a little bit of code safety to that call.  I am moving the ticket back to 1.5-tc1

PR #1744, opened 2016-07-17
* Alternative solution which doesn't require a timer

ehuelsmann:
And fixes my comment about the "Save password" button saving the preferences too.


ehuelsmann:
Re #1742

Issue #1745, opened 2016-07-17
contact form not displaying users


einhverfr:
When I try to render a form with a user, I get a blank.

PR #1746, opened 2016-07-17
Fixing #1745

einhverfr:


einhverfr:
This probably fixes a fairly large number of bugs.....

Issue #1148, opened 2015-12-23
LedgerSMB::Scripts::setup.pm untranslated strings


ehuelsmann:
While checking the results from the new translated-strings scanner, this error appears:

./LedgerSMB/Scripts/setup.pm:178: text() called with non string first argument; resetting scan (consider using maketext() directly!)

However, the strings referenced at that line ($dispatch_entry->{$field}), isn't marked for translation (it's an entry from @login_actions_dispatch_table defined on LedgerSMB/Scripts/setup.pm:93.

einhverfr:
This is trivial to fix actually.  Will fix it now, for rc1

einhverfr:
In the process I am removing pre-1.2 support from the dispatch table.

einhverfr:
there's also a bug here.
translating next_action is quite definitely unsafe.

ehuelsmann:
You mean 1.0 and 1.1? Or 1.2 as well?

einhverfr:
sorry, I mean 1.1.x and before.  Sorry for that.

einhverfr:
Also removed 1.3 betas from the dispatch table too.
1.2 and 1.3 still show up correctly.

PR #1749, opened 2016-07-17
Fixing #1148

einhverfr:
Additionally removes detection for lsmb 1.0 and 1.3 betas.


einhverfr:
sorry that is 1.0, not 1.x

PR #1776, opened 2016-07-19
Ensure that migrated reconciliations are not viewed as mismatched

ylavoie:
Migration from SL30 must have our_balance set from their_balance otherwise they will appear as problem imports.

### DIFF
--- a/sql/upgrade/sl3.0-1.5.sql
+++ b/sql/upgrade/sl3.0-1.5.sql
@@ -651,7 +651,8 @@ FROM (
   FROM _cr_report_line
 ) cr1
 WHERE cr.id = cr1.id;
-UPDATE cr_report_line SET insert_time = post_date;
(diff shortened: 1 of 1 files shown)


sql/upgrade/sl3.0-1.5.sql
```diff
@@ -651,7 +651,8 @@ FROM (
   FROM _cr_report_line
 ) cr1
 WHERE cr.id = cr1.id;
-UPDATE cr_report_line SET insert_time = post_date;
+UPDATE cr_report_line SET insert_time = post_date,
+                          our_balance = their_balance;
```
einhverfr: having their_balance = 0 is fine, or having our_balance = their_balance.

I think this is a good change in part because our_balance = lsmb balance (as always) and their_balance in this case idicates sl balance.  Merging

PR #1777, opened 2016-07-19
Disable submit button on unbalanced reconciliation report.

ylavoie:


UI/reconciliation/report.html
```diff
@@ -260,6 +260,7 @@
     name = "action",
     text = text('Submit'),
     value = 'submit_recon_set',
+    disabled = ( their_total - beginning_balance != total_cleared_credits - total_cleared_debits ? 'true' : 'false'),
```
einhverfr: you can use the variance for this value for this.  Where variance is 0, the report is balanced.

PR #1819, opened 2016-07-24
Most business unit addition issues fixed

einhverfr:
parent selection options not showing so no hierarchy.


einhverfr:
This fully fixes 1799

PR #1833, opened 2016-07-24
Fixing db test harness,

### LINKED ISSUE #1830: Tests returning NULL on db are not treated as failures and they need to be
We have tests in Reconciliation and Report that are failing but incorrectly ignored by our test harness.

I am going to fix this and merge my change even though it will cause Travis to fail.

einhverfr:
Correcting test harness. Expect Reconciliation.sql and Reports.sql to fail until we fix it

Fixes #1830


einhverfr:
Expecting two (correct) failures in 43-dbtests.pl.  These should have been failing but were not.

PR #1838, opened 2016-07-25
Optimize make dojo

ylavoie:
Save last compiled Dojo, Dijit & lsmb in a tar file. Use it instead of recompiling on next invocation is git reports identical SHAs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,7 @@ matrix:
             - nginx
 
 before_cache:
-  - mkdir $HOME/dojo_archive 2>/dev/null
+  - mkdir -p $HOME/dojo_archive
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -72,10 +72,14 @@ matrix:
           packages:
             - nginx
 
+before_cache:
+  - mkdir $HOME/dojo_archive 2>/dev/null
```
sbts: rather than redirecting std err to /dev/null I'd prefer to see 
```bash
mkdir -p $HOME/dojo_archive
```
That way the mkdir is silent if the dir already exists or is successfully created, but if the creation fails we will see any errors

sbts:
As per Line comments A change to the mkdir is probably advisable

PR #1850, opened 2016-07-25
1.4 New COA file for New Zealand, contributed by kiwi (irc/matrix)

sbts:


einhverfr:
Waiting until I can merge both this and master.  Recommend same changes to both for easier mainteance (silence critic and then bring over the commits)

PR #1854, opened 2016-07-26
This is a lot of work but fixes #1852

### LINKED ISSUE #1852: Danish characters Å and å improperly escaped
These are escaped as \RA and \ra instead of \AA and \aa.

Unsure why.  the LaTeX::Encode seems correct.

einhverfr:


einhverfr:
Recommend switching to LaTeX::Encode for 1.6

PR #1894, opened 2016-07-29
Fixing account closing before transactions are entered

einhverfr:


sql/modules/EndOfYear.sql
```diff
@@ -55,7 +55,8 @@ BEGIN
                 from account_checkpoint
                 WHERE end_date = cp_date
                 ) cp on (a.chart_id = cp.account_id)
-        group by COALESCE(a.chart_id, cp.account_id);
+        RIGHT JOIN account ON account.id = a.chart_id or account.id = cp.account_id
+        group by COALESCE(a.chart_id, cp.account_id), account.id;
```
ehuelsmann: @einhverfr souldn't this line become:

```sql
   GROUP BY account.id
```
?

A variant of it is what I had to do on master-mc to make this function work.

PR #1899, opened 2016-07-30
Qty discounts,

### LINKED ISSUE #1790: Qty discounts via pricematrix
It would be nice to be able to specify bulk pricing via the pricematrix.

There are a number of problematic corner cases here.  These include:

1.  Do all have to be on the same line?
2.  If a lower price is found, do we re-calculate, possibly overriding something manually entered?

einhverfr:
fixes #1790

### DIFF
--- a/lib/LedgerSMB/PriceMatrix.pm
+++ b/lib/LedgerSMB/PriceMatrix.pm
@@ -71,7 +71,7 @@ sub price_matrix_query {
 
     if ( $form->{customer_id} ) {
         $sth = $dbh->prepare('
-            SELECT * FROM pricematrix__for_customer(?, ?, ?) 
+            SELECT * FROM pricematrix__for_customer(?, ?, ?, ?) 
         ') || $form->dberror('pricematrix__for_customer');
     }
     elsif ( $form->{vendor_id} ) {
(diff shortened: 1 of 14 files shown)


lib/LedgerSMB/PriceMatrix.pm
```diff
@@ -173,6 +124,11 @@ sub price_matrix {
                            - ($sellprice * ($mref->{pricebreak} / 100));
             }
             $ref->{sellprice} = $sellprice;
+            if ($mref->{qty} > $form->{qtycache}->{$ref->{id}}){
```
ehuelsmann: @einhverfr just ran into a problem with the quantity cache, I think: when loading an existing invoice from the database, the price matrix is applied to each item individually instead of running the pricematrix at the end.

(which explains part of the cases of 'rowcount' being '' or undefined...

ehuelsmann: It seems that this: https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/IS.pm#L1600
loop needs to be split in a loop loading from the database and a loop applying the prices from the price matrix.

I'm assuming there's something similar going on in the 'ir' and 'oe' sides of things.

PR #1901, opened 2016-07-30
Prefactoring for new widgets.

einhverfr:
Reason is, need some logic sharing with new code framework but not entirely

This eliminates the lsmb-request.pl file and thus ensures no new code in bin/


einhverfr:
althugh this is tested and works, I think it makes a mess of the PSGI.pm and so will close this and start over.

PR #1905, opened 2016-07-30
adding current directory explicitly

einhverfr:


lsmb-request.pl
```diff
@@ -19,6 +19,7 @@ =head1 COPYRIGHT
 =cut
 
 package LedgerSMB::Handler;
+use lib '.';
```
sbts: rather than adding '.' as a relative path, can we explicitly add the Current Working Directory?
 Add a line note

sbts: I'd propose this block of code
```
use Cwd 'abs_path';
use File::Basename;
use $dirname;
```
Or something similar

einhverfr: I don't have a problem with a block of code similar to that.  However there are a couple costs to think about:

1.  That block won't work as such.  You can't really use variables with use because use is compile time. while your path operations are run-time
2.  Working around 1  above means significant effort, a begin block, and is something far more brittle.

I guess the question is given that this is only going in the http request handlers, what would be gained by that complexity.  I do agree it would not be a great idea to put in command line tools but then you would not run a command line tool in that way from just anywhere, so presumably we'd make findbin work (and for some reason on Debian it is *not* working so this is a workaround).

sbts: Actually, that exact solution won't quite work if the file isn't in the TL dir of LedgerSMB, rather we would want to strip off the trailing portion of the path, so we end up with the correct parent.

einhverfr: http://search.cpan.org/~nwclark/perl-5.8.7/lib/FindBin.pm has some docs for FindBin, but again, it is *not* working for Debian at present, so my guess is they broke something more significant.....

sbts: Ok, thanks for the corrections and valid points.
As discussed with ehuelsmann it seems releasing with `use '.';` is the first step, then we can consider going further down the track.

PR #1906, opened 2016-07-30
adding current directory explicitly

einhverfr:


lsmb-request.pl
```diff
@@ -19,7 +19,7 @@ =head1 COPYRIGHT
 =cut
 
 package LedgerSMB::Handler;
-
+use lib '.';
```
sbts: rather than adding '.' as a relative path, can we explicitly add the Current Working Directory?

sbts: I'd propose this block of code
```
use Cwd 'abs_path';
use File::Basename;
use $dirname;
```
Or something similar

sbts: Actually, that exact solution won't quite work if the file isn't in the TL dir of LedgerSMB, rather we would want to strip off the trailing portion of the path, so we end up with the correct parent.

einhverfr: you would want to use FindBin, but it is also currently broken on Debian so......

einhverfr: After some further testing, it looks like FindBin actually works but lib->import() doesn't which is why we saw the breakage.  I think for 1.4.32 we should replace the above with:

```
use FindBin;
use lib $FindBin::Bin;
```

PR #1913, opened 2016-07-30
Removing probably unneccessary use lib '.' from modules in master,

### LINKED ISSUE #1911: Use Absolute Path instead of Relative '.'
From various discussions in a matrix room, and two PR's
[#ledgersmb-security](https://vector.im/develop/#/room/!fbDhnxnlASGdwPNIxG:matrix.org/$1469895146714623hTdDM:matrix.org)   (This room is invite only. Contact dcg_mx, John, ehuelsmann in [#ledgersmb](https://vector.im/develop/#/room/#ledgersmb:matrix.org) for access)
https://github.com/ledgersmb/LedgerSMB/pull/1905/files
https://github.com/ledgersmb/LedgerSMB/pull/1906/files

In July 2016 Debian (and presumably other distros) applied a security patch to Perl that broke code relying on modules being included from the implied current working directory.
To resolve this quickly we have added 
```perl
use '.';
```
to a number of files.

In the future rather than adding '.' as a relative path, can we explicitly add the Current Working Directory?

I'd propose something equivalent to this block of code (which as einhverfr pointed out won't work as you can't use variables in a `use` clause)

```perl
use Cwd 'abs_path';
use File::Basename;
use $dirname;
```
Or something similar

einhverfr
> I don't have a problem with a block of code similar to that. However there are a couple costs to think about:

> That block won't work as such. You can't really use variables with use because use is compile time. while your path operations are run-time
Working around 1 above means significant effort, a begin block, and is something far more brittle.
I guess the question is given that this is only going in the http request handlers, what would be gained by that complexity. I do agree it would not be a great idea to put in command line tools but then you would not run a command line tool in that way from just anywhere, so presumably we'd make findbin work (and for some reason on Debian it is not working so this is a workaround).

sbts
>Actually, that exact solution won't quite work if the file isn't in the TL dir of LedgerSMB, rather we would want to strip off the trailing portion of the path, so we end up with the correct parent.

einhverfr
> http://search.cpan.org/~nwclark/perl-5.8.7/lib/FindBin.pm has some docs for FindBin, but again, it is not working for Debian at present, so my guess is they broke something more significant.....

sbts
>Ok, thanks for the corrections and valid points.
As discussed with ehuelsmann it seems releasing with use '.'; is the first step, then we can consider going further down the track.

einhverfr
>After some further testing, it looks like FindBin actually works but lib->import() doesn't which is why we saw the breakage. I think for 1.4.32 we should replace the above with:
```perl
use FindBin;
use lib $FindBin::Bin;
```

einhverfr:
fixes #1911
Note that we already have to use lib 'lib';

### DIFF
--- a/tools/starman.psgi
+++ b/tools/starman.psgi
@@ -6,7 +6,8 @@ BEGIN {
 
 package LedgerSMB::FCGI;
 
-use lib 'lib';
+use FindBin;
+use lib $FindBin::Bin . "/../lib";
 use CGI::Emulate::PSGI;
 use LedgerSMB::PSGI;
 use LedgerSMB::Sysconfig;
--- a/Changelog
+++ b/Changelog
@@ -53,6 +53,9 @@ JohnL is John Locke
 Changelog for 1.4 Series
 Released 2014-09-15
 
+Changelog for 1.4.32
+* Moved to findbin from use lib '.' (Chris T, #1911)
+
 Changelog for 1.4.31
 * Fixed user with contact all privileges can't create contacts (Chris T #1219)
 * Fixed new user with manage users can't see employee (Chris T #1841)
--- a/bin/lsmb-request.pl
+++ b/bin/lsmb-request.pl
@@ -1,4 +1,3 @@
-use lib '.';
 sub get_script {
     my ($locale, $request) = @_;
 
--- a/bin/old-handler.pl
+++ b/bin/old-handler.pl
@@ -48,7 +48,6 @@
 
 our $logger=Log::Log4perl->get_logger('old-handler-chain');#make logger available to other old programs
 Log::Log4perl::init(\$LedgerSMB::Sysconfig::log4perl_config);
-use lib '.';
 
 # Clearing all namespaces for persistant code use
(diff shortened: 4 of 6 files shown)


tools/starman.psgi
```diff
@@ -6,7 +6,8 @@ BEGIN {
 
 package LedgerSMB::FCGI;
 
-use lib 'lib';
+use FindBin;
+use lib $FindBin::Bin . "/../lib";
```
ehuelsmann: @einhverfr did you test this? Is this running off the script's directory, or off Starman's directory? (the FindBin::Bin directory, I mean).

My intuition is that this doesn't work the way you intend it to.

einhverfr: Yes I did test it.   It runs relative to the starman.psgi file.
One serious issue is that the directory that you run this now must have a valid ledgersmb.conf file in it but that is a separate problem really.

einhverfr: As a note, I am looking at
1.  Removing the implicit load from Sysconfig and requiring a path passed in to an API when we want to load.
2.  Removal of other relative directories in PSGI.pm

That should improve both testability and safety.  I have the second working but the first is giving me some trouble

PR #1940, opened 2016-07-31
Round_amount must comes from Form

ylavoie:


lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1147,9 +1147,9 @@ for my $ref (0 .. $#array_options) {
          # same names are used for ap/ar accounts w/o the cash prefix.
          #
      my $sign = "$array_options[$ref]->{due_fx}" <=> 0;
-     if ( $request->round_amount($sign * "$array_options[$ref]->{due_fx}")
+     if ( Form::round_amount($sign * "$array_options[$ref]->{due_fx}")
```
einhverfr: hmmmm  this was moved to the number classes iirc but we might have accidently removed the rounding from that.  I don't think we should rely on Form in any new code.

ehuelsmann:
Actually, we didn't forget to implement rounding: LedgerSMB::PGNumber is a PGObject::Type::BigFloat which in turn is a Math::BigFloat. The last one has rounding built in. The idea might then be to:
```perl
if (LedgerSMB::PGNumber($sign * ...)->from_input->bround($digits_to_preserve))
```

Issue #1944, opened 2016-08-01
Clicking on an account number on Trial balance report crashes the application


ylavoie:
Undefined subroutine &LedgerSMB::Report::maketext called at lib/LedgerSMB/Report/GL.pm line 168.

ylavoie:
Fixed by PR https://github.com/ledgersmb/LedgerSMB/pull/1945

PR #1946, opened 2016-08-01
Revert "Use text instead of undefined maketext"

ehuelsmann:
Reverts ledgersmb/LedgerSMB#1945

### DIFF
--- a/lib/LedgerSMB/Report/GL.pm
+++ b/lib/LedgerSMB/Report/GL.pm
@@ -166,7 +166,7 @@ sub columns {
     my @bclasses = LedgerSMB::Business_Unit_Class->list('1', 'gl');
     for my $class (@bclasses){
         push @COLS, {col_id =>  "bc_" . $class->id,
-                       name => LedgerSMB::Report::text($class->label),
+                       name => LedgerSMB::Report::maketext($class->label),
                        type => 'text',
                      pwidth => '2'};
     }


ehuelsmann:
Found it: we should be using Text() instead; both come fro  I18N

ehuelsmann:
Hmmm. The columns above use `$self->Text()`, but those want to be extracted, because they use inline string constants.

Puzzled about a good solution now...

('maketext' would be the name to use for any dynamically translated argument everywhere...)

einhverfr:
Yeah, I looked at it before I merged.  The problem is two-fold:
1.  This is exactly the wrong place for using our .po's.   It should be manual translation, but
2.  We probably have at least partial .po coverage here.

I think a correct solution would be too intensive for a quick fix.  A partial "it works" fix might be to create a hash of default keys/translated strings and substitute if and only if it is on that list.

ehuelsmann:
Hmm. The default list of business unit classes is being extracted from the SQL files and inserted into the po file.

While that may work for translation of the standard classes, this doesn't work for custom classes. For the custom classes, I think you're correct: we should have a manual (in-app) translation method for those.

ehuelsmann:
What the last comment meant to say: calling 'maketext()' will result in a translated class name "in the usual case". Except that there apparently isn't a maketext() to call.

ehuelsmann:
Instead of #1944, I've created a new pull request (#1948) which comes in the place of #1944 and #1946.

PR #1948, opened 2016-08-01
* Add 'maketext' and 'Maketext' functions (replaces #1946)

ehuelsmann:


ehuelsmann:
@ylavoie please evaluate this solution instead of the other one and confirm.

PR #1956, opened 2016-08-02
do not call pricematrix on existing order/invoice, fixes #1954

### LINKED ISSUE #1954: Merge of #1899 causes problems loading existing invoices
@einhverfr As of the merge of #1899, I can't load existing invoices anymore. I'm getting:

screenshot from 2016-08-01 23 09 08

Note that this is with the simple case of loading a single-line invoice (where the customer doesn't have customer-specific prices, but I think that doesn't matter).

The cause as I have been able to identify it, is that `$form->{rowcount}` is still `''` or `undef` during the loading of the invoice, which means that the `qtycache` in the form doesn't get initialized for the specific part -- not even set to `0`.

Now, my first reaction was to use `// 0`; however, when I do that, I ignore the fact that (as per my comment on the code in #1899) that will lead to incorrect calculation of the qty discount (calculation based on zero items) for invoices that should have qty discounts applied.

Is this a relatively easy fix? Or should we back out for the sake of stabilizing 1.5.0?

einhverfr:


lib/LedgerSMB/PriceMatrix.pm
```diff
@@ -94,6 +94,7 @@ Updates $ref with the price matrix outcomes given $transdate and $form.
 
 sub price_matrix {
     my ( $pmh, $ref, $transdate, $decimalplaces, $form, $myconfig) = @_;
+    return if $form->{id};
```
ehuelsmann: @einhverfr I know we discussed this, but what happens if/when a saved invoice gets edited before it gets posted?

PR #1972, opened 2016-08-04
Fixing starman on debian, fix not appicable to master

einhverfr:


tools/starman.psgi
```diff
@@ -2,6 +2,9 @@
 
 package LedgerSMB::FCGI;
 
+no lib '.';
```
einhverfr: Two reasons for this line btw.
1.  Tightens security just a little bit (though in this case, really not an issue)
2.  It ensures that test harnesses find the breakage if it exists.

PR #1980, opened 2016-08-04
* Fix #1937 by adding a '_download' link target

### LINKED ISSUE #1937: 1.5: Can't download file attached to <anything>
The error shows "Invalid server response: document contains no BODY tag"; which is probably not just restricted to the Entity/ECA screens.

The issue is that the form handler tries to load the content into the `#maindiv` instead of actually downloading it.

## Proposed solution

We need to special-case the downloads, just as we do for the printed documents with the "Print" button. One way to special case links in the new system is to provide an explicit `target` attribute on the A tag.

ehuelsmann:


UI/journal/journal_entry.html
```diff
@@ -321,7 +321,7 @@
 <?lsmb FOREACH file IN form.files ?>
 <tr>
 <td><a href="file.pl?action=get&file_class=1&ref_key=<?lsmb form.id 
-            ?>&id=<?lsmb file.id ?>"
+            ?>&id=<?lsmb file.id ?>" target="_download"
```
ehuelsmann: This is the only change in this file, apart from tabs-to-spaces conversion.

UI/lib/dynatable.html
```diff
@@ -38,57 +38,60 @@
...
                       HREF = COL.href_base _ ROW.$HREF_SFX;
                    ELSE;
                       HREF = COL.href_base _ ROW.row_id;
-                   END
-          ?><a href="<?lsmb HREF ?>"><?lsmb ROW.${COL.col_id} ?></a>
+                   END;
+                   IF COL.href_target;
+                      HREF_TGT = ' target="' _ COL.href_target _ '"';
+                   END;
+          ?><a href="<?lsmb HREF ?>"<?lsmb HREF_TGT ?>><?lsmb ROW.${COL.col_id} ?></a>
```
ehuelsmann: This block is the only functional change here; the rest are tabs-to-spaces conversion.

PR #1988, opened 2016-08-05
Fixes to JS & CSS for standardization and Codacy compliance.

ylavoie:
Codacy tests CSS & JS against lots of rules for Code Style, Performance and Error Prone code.


UI/js-src/lsmb/Form.js
```diff
@@ -20,10 +20,10 @@ define([
                       this.inherited(arguments);
 
                       // <button> tags get rewritten to <input type="submit" tags...
-                      query('input[type="submit"]', this.domNode)
+                      query("input[type=\"submit\"]", this.domNode)
```
ehuelsmann: @ylavoie can't we use `query("input[type='submit']"` here then? much easier to read than all the escaped dquotes

PR #2008, opened 2016-08-06
Check for duplicated functions

ylavoie:


sql/modules/Exempt_funcs.sql.inc
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE pg_temp.test_exempt_funcs (funcname text primary key);
```
einhverfr: just a note for the future.

wondering if we want to allow for ongoing diagnostics?  If so we'd want to keep this in a permanent table and delete from it before reload.

einhverfr: that's just a note for future endeavors, not blocking on this one.

PR #2023, opened 2016-08-07
Moves the preferences into a tabbed interface

### LINKED ISSUE #1740: Password change widget needs to be rewritten as a full dojo widget
This requires getting to know the following:

1.  Dojo 18n
2.  Full form widget designs
3.  Generalizing event handers

It would also likely serve as a basis for login widgets and the like.

einhverfr:
Adds a password dojo widget
There is a minor bug where it is not setting the class for the strength indiator correctly (expect to have color coding via css).
Fixes #1740

### DIFF
--- a/UI/js-src/lsmb/users/ChangePassword.js
+++ b/UI/js-src/lsmb/users/ChangePassword.js
@@ -1,0 +1,30 @@
+
+define(['lsmb/TabularForm',
+        'dojo/_base/declare',
+         'dijit/_WidgetBase',
+         'dijit/_TemplatedMixin',
+         'dijit/_WidgetsInTemplateMixin',
+         'dijit/_Container'],
+       function(tabform, declare, _widgebase, _templatemixin, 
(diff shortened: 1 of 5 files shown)


UI/js-src/lsmb/users/ChangePassword.js
```diff
@@ -0,0 +1,155 @@
...
+                      var I = this;
+                      require(
+                           ['dojo/request'],
+                           function(r) {
+                              console.log('change_pw clicked');
+                              var login = document.getElementById('username').value;
+                              var old_password = I.oldpw.get('value');
+                              var new_password = I.newpw.get('value');
+                              var confirm_pass = I.verified.get('value');
+                              if (old_password == '' || new_password == '')
```
ehuelsmann: after @ylavoie's branch being merged, I'd expect this line to be:

```javascript
if (old_password === "" || new_password === "")
```

PR #2028, opened 2016-08-07
Fix reconciliation reports

ylavoie:
Pull only the proper transactions in reconciliation reports
Adjust tooltip for suspect cleared delay


einhverfr:
As a note, this looks fine to me.  I am merging it in.  The scope fo this change is solely o bring in SL reconciliations and it does not impact the LedgerSMB recon process.

Issue #994, opened 2015-11-01
SF#1384 No "save" button for non-posted transactions


einhverfr:
See https://sourceforge.net/p/ledger-smb/bugs/1384/

This affects sales invoices but not AR transactions if memory serves me.

ehuelsmann:
Heavy workflow problem, but does not result in incorrect accounting. Not a 'bug:accounting' --> reclassifying.

ehuelsmann:
Git branch 1.4/be56030e42

    Create new AP transaction: AP › Add Transaction
    Save. (Twice. See #1383.)
    Locate via Transaction Approval › Drafts
    Open new tx

=> There is an 'Update' button (which recalculates form), but no 'Save' button (see image ap-edit-tx-no-save.png)

ehuelsmann:
Like #975, this is a design problem with old code. (Or probably with LedgerSMB in general; I've not seen any workflow designs written down anywhere so far.)

Based on my findings so far, moving to 1.6, which can address this at the earliest (with 1.5 being feature complete, there's no way we can get shocking changes in).

einhverfr:
turned out to be a bug, in that we were checking for roles that didn't exist.  Pending PR #2063

einhverfr:
backport to 1.4 will look just a little different, but will backport.

einhverfr:
The problem was in ar/ap transactions and gl because a role name changed and we didn't catch it.

einhverfr:
last worked in 1.3

PR #2063, opened 2016-08-13
Fixes #1715

### LINKED ISSUE #1715: Fix user-interface related handling of roles for reconciliation
The current Reconciliation module only protects deletes, not approves or rejects.
It also relies on role checking which isn't allowed for a package with LedgerSMB 1.3 architecture.
Either put the access checks in the database or complete porting this portion to 1.5 with Moose.

einhverfr:
This is a somewhat disruptive fix, but it centralizes the problem domain,
enforces better code consistency, and makes debugging easier.
It also corrects another problem in master and a fix for part of this will
need to be backported.

However, for now, is_allowed_role is supported on PGOld, LedgerSMB and Form
objects out of the box and can be trivially added elsewhere.

### DIFF
--- a/bin/aa.pl
+++ b/bin/aa.pl
@@ -965,12 +965,12 @@ sub form_footer {
                    ndx   => 3,
                    key   => 'O',
                    value => $locale->text('Post') };
-           if ($form->is_allowed_role(['draft_modify']){
+           if ($form->is_allowed_role(['draft_modify'])){
                $button{edit_and_save} = {
                    ndx   => 4,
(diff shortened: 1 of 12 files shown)


einhverfr:
It also ensures cucumber test failures when we ask for roles that don't exist  (which we currently don't fail on)

einhverfr:
This also fixes #994

einhverfr:
This has the potential to be slightly disruptive but it also means that there are classes of errors we will now catch that we could not catch before.

PR #2074, opened 2016-08-14
making qty default to 0

einhverfr:
for timecards, and adding error if not submitted properly


einhverfr:
This avoids the error while making the behavior conform to 1.5

PR #2077, opened 2016-08-14
A few fixes for batch entry for ar/ap/gl

einhverfr:


lib/LedgerSMB/Scripts/vouchers.pm
```diff
@@ -189,16 +189,19 @@ sub add_vouchers {
...
-            { no strict; no warnings 'redefine'; do $script; }
+            require 'bin/bridge.pl'; ## no critic
+            $form->{script} = $script;
+            $form->{script} =~ s#^bin/##;
             lsmb_legacy::locale($locale);
             lsmb_legacy::form($form);
+            { no strict; no warnings 'redefine'; do $script; }
             $vouchers_dispatch->{$request->{batch_type}}{function}($request);
 
             exit;
```
ehuelsmann: there is an 'exit' here... I think that deserves a comment on why it is valid. (The diff doesn't show enough info to see it)

einhverfr: The exit occurs in a fork, so it exits the child.  Without the else, it then tries to run again.

PR #2104, opened 2016-08-19
whitespace cleanup for overpayment2

einhverfr:
and changing repeated push to a single one.  Does not need to be merged before rc2


einhverfr:
Not the right approach.  trying a different approach.

PR #2117, opened 2016-09-02
Redirect to /login.pl if url is empty

ylavoie:


tools/starman.psgi
```diff
@@ -42,6 +42,11 @@ my $old_app = LedgerSMB::PSGI::old_app();
 my $new_app = LedgerSMB::PSGI::new_app();
 
 builder {
+
+    enable 'Redirect', url_patterns => [
+        '^/?$' => ['/login.pl',302]
```
ehuelsmann: Shouldn't this be qr// instead of a string in quotes?

PR #2119, opened 2016-09-02
Allow remote headless browsers

ylavoie:


xt/65-browser.t
```diff
@@ -0,0 +1,46 @@
+#!perl
+
+use Test::More;
+use YAML::Syck;
+
+my @missing = grep { ! $ENV{$_} } @reqenv;
```
ehuelsmann: `@reqenv `isn't defined?

xt/65-browser.t
```diff
@@ -0,0 +1,46 @@
+#!perl
+
+use Test::More;
```
ehuelsmann: `use strict;
use warnings;`
?

ylavoie:
Fixed

PR #2181, opened 2016-10-05
Enable Log4perl globally for PSGI modules

ylavoie:


ehuelsmann:
@ylavoie I've merged a change which initializes Log4perl at the beginning of starman.psgi; meaning before any requests are being handled and before forking off any "old code" instances.

That's the minimal change. I'd like to discuss how to handle the difference between logging routes and logging back-end functionalities on the mailing list. For now, I think we don't want to merge this PR, pending the outcome of that discussion...

PR #2203, opened 2016-10-13
Limit reconciliation__get_cleared_balance cleared dates to asked repo…

ylavoie:
…rt date


lib/LedgerSMB/DBObject/Reconciliation.pm
```diff
@@ -70,10 +70,6 @@ sub update {
     my $beginning_balance = $self->{beginning_balance} // 0;
     my $total_cleared_credits = $self->{total_cleared_credits} // 0;
     my $total_cleared_debits = $self->{total_cleared_debits} // 0;
-    $self->{submit_allowed} =
-        abs(($their_total - $beginning_balance)
-            - ($total_cleared_credits - $total_cleared_debits))
-        >= 0.001;
```
einhverfr: For safety, probably better make that greater than 0.005 since that would round up to 0.01.

ylavoie: But this was removed in favor of variance.

PR #2204, opened 2016-10-13
T 15 reconciliation  get cleared balance

ylavoie:


einhverfr:
I think this is a positive change.  We just need to deal with the fact that underlying files moved and then all for committing this.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -429,6 +429,25 @@ push @tests, __PACKAGE__->new(
 );
 
 
+# They should be multiple selected from account_link_description(description)
```
ehuelsmann: This change does not make sense and does not belong in this change: accounts are not required to have links and whether or not they do is unrelated to reconciled balances

ehuelsmann:
There is a large number of replacements is where `true` is replaced by `TRUE` and similar. It feels like that changes the coding style of the file. Was that necessary to implement this change?

PR #2274, opened 2016-11-16
Fully Dojo enabled Reconciliation report

ylavoie:
A few improvements:
- Highlight selected lines for easier handling
- Removes the checkboxes when the report is submitted or approved
- Makes cleared, mismatch and outstanding tables foldable and fold them when they have no entries or are not appropriate. They can be unfolded at will.


UI/js-src/lsmb/SubscribeToggleButton.js
```diff
@@ -0,0 +1,21 @@
+define(["dojo/_base/declare",
+        "dojo/on",
+        "dojo/topic",
+        "dijit/form/ToggleButton"],
+       function(declare, on, topic, ToggleButton) {
+           return declare("lsmb/SubscribeCheckBox", [ToggleButton], {
```
ehuelsmann: copy/paste-o? presumably that needs to be `lsmb/SubscribeToggleButton` ?

UI/js-src/lsmb/Reconciliation.js
```diff
@@ -0,0 +1,31 @@
...
+       ["dojo/_base/declare",
+        "dojo/topic",
+        "dojo/query",
+        "lsmb/Form",
+        "dijit/_Container",
+        "dojo/NodeList-dom",     // To load extensions in query
+        "dojo/domReady!"],
+       function(declare, Topic, Query, Form, _Container) {
+           return declare("lsmb/Reconciliation", [Form, _Container], {
+               update: function(targetValue, prefix) {
```
ehuelsmann: Why did you decide to *replace* the form with the reconciliation object? Are we sure that we will never have a situation where the reconciliation will be *part* of a form? (i.e. where the widget isn't the only thing contained in the form)

UI/js-src/lsmb/Reconciliation.js
```diff
@@ -0,0 +1,31 @@
+define("lsmb/Reconciliation",
+       ["dojo/_base/declare",
+        "dojo/topic",
+        "dojo/query",
+        "lsmb/Form",
+        "dijit/_Container",
+        "dojo/NodeList-dom",     // To load extensions in query
+        "dojo/domReady!"],
+       function(declare, Topic, Query, Form, _Container) {
```
ehuelsmann: you use `Topic` and `Query` here, but all other files in our code base us `topic` and `query` (no capitals)

UI/js-src/lsmb/ReconciliationLine.js
```diff
@@ -0,0 +1,32 @@
+define("lsmb/ReconciliationLine",
+       ["dojo/_base/declare",
+        "dojo/on",
+        "dijit/registry",
+        "dojo/dom-class",
+        "dijit/form/CheckBox"],
+       function(declare, on, registry, domClass, CheckBox) {
+           return declare("lsmb/ReconciliationLine", [CheckBox], {
+               publish: function(targetValue) {
```
ehuelsmann: If you look at https://github.com/ehuelsmann/LedgerSMB/blob/master/UI/js-src/lsmb/PublishSelect.js#L10, you see that that function (also called `publish`) invokes an action on a channel. I'm missing that part below, right?

UI/reconciliation/report.html
```diff
@@ -293,7 +342,7 @@
     name = "action",
     text = text('Approve'),
     value = 'approve',
-    disabled = submit_allowed == 0 ? 'false' : 'true',
+    disabled = out_of_balance == '0.00' ? 'false' : 'true',
```
ehuelsmann: This is a string comparison, but what happens when the number of digits is set to 3 (or, in case of Japanese Yen, to zero)?

UI/reconciliation/report.html
```diff
@@ -220,14 +261,16 @@
...
+        <tr class="record" id=<?lsmb "recon-line-$row.id" ?>>
+            <td><?lsmb IF !approved and !submitted;
+            INCLUDE input element_data = {
             type = "checkbox"
             name = "cleared_$row.id"
             value = row.id
             checked = row.cleared
-            }
-            ?></td>
+            "data-dojo-type"  = "lsmb/ReconciliationLine"
```
ehuelsmann: On the invoice, the `InvoiceLine` matches the `<tbody><tr>...</tr></tbody>` tags in the table representing invoice lines. Isn't it confusing to have these named `...Line` when they are just a checkbox?

ehuelsmann:
@ylavoie also, could you take a few more words than "Fully dojo enabled reconciliation"? What does that mean? I haven't been able to deduce from the code changes what has exactly changed on the recon screen. Does the checkmark we're adding allow the selection of all checkmarks before the rows on the recon screen? If so (and nothing more), I'm going to see how this functionality is implemented on the payments screen, because something similar was implemented there (and I just love code reuse, whenever possible!).

UI/js-src/lsmb/SubscribeToggleButton.js
ylavoie: Yes. Fixed

UI/js-src/lsmb/Reconciliation.js
ylavoie: Fixed, but we use Form & _Container here and elsewhere.

UI/js-src/lsmb/Reconciliation.js
ylavoie: lsmb/Reconciliation is a misnomer but it really inherit from lsmb/Form to specialize it and enable it to fold/unfold the cleared, mismatch and outstanding display sections.

UI/js-src/lsmb/ReconciliationLine.js
ylavoie: Misnomer. Renamed to _display. That handles the display of cleared, mismatch or outstanding, based on a message received from a ToggleButton

UI/reconciliation/report.html
ylavoie: Fixed, now comparing actual and desired totals.

UI/reconciliation/report.html
ehuelsmann: Ah. Now I see the structure. What I'd like to see here is a structure similar to the one in the invoice lines: the elements within the line send out messages on state-change events to a line-specific channel. Interested widgets (such as the line itself) listen on the channel and act appropriately. E.g. the widget to delete the invoice line only sends out a click-event on the channel. The InvoiceLine widget takes action and removes itself both from the DOM as well as from its parent.
I'm thinking the structure of this problem is the same and it would be good to handle it the same way.

PR #2288, opened 2016-11-19
Limit reconciliations to ordered dates

ylavoie:
Current reconciliation code rely on the fact that approval closely follows submit and fails when many submit are done without approval because it doesn't take reconciliation date into account.
This PR fix that and also enable proper SQL-Ledger migrations to be automatically submitted, but have delayed approval.

### DIFF
--- a/lib/LedgerSMB/Scripts/recon.pm
+++ b/lib/LedgerSMB/Scripts/recon.pm
@@ -217,7 +217,6 @@ sub _display_report {
     $recon->add_entries($recon->import_file($contents))
         if $contents && !$recon->{submitted};
     $recon->{can_approve} = $request->is_allowed_role({allowed_roles => ['reconciliation_approve']});
-    $recon->get();
     $recon->{form_id} = $request->{form_id};
     my $template = LedgerSMB::Template->new(
         user=> $recon->{_user},
--- a/sql/Pg-database.sql
+++ b/sql/Pg-database.sql
@@ -1167,7 +1167,6 @@ CREATE TABLE cr_report (
     approved_by int references entity(id),
     approved_username text,
     recon_fx bool default false,
-    max_ac_id int,
     CHECK (deleted is not true or approved is not true)
 );
 
@@ -1479,8 +1478,6 @@ CREATE TABLE acc_trans (
   voucher_id int references voucher(id),
   entry_id SERIAL PRIMARY KEY
 );
-
-ALTER TABLE cr_report ADD FOREIGN KEY (max_ac_id) REFERENCES acc_trans(entry_id);
 
 COMMENT ON TABLE acc_trans IS
 $$This table stores line items for financial transactions.  Please note that
--- a/sql/modules/Reconciliation.sql
+++ b/sql/modules/Reconciliation.sql
@@ -27,16 +27,24 @@ set client_min_messages = 'warning';
 
 BEGIN;
 
+DROP FUNCTION IF EXISTS reconciliation__submit_set(int, int[]);
 CREATE OR REPLACE FUNCTION reconciliation__submit_set(
-        in_report_id int, in_line_ids int[]) RETURNS bool AS
+        in_report_id int, in_line_ids int[], in_end_date date
+      ) RETURNS bool AS
 $$
 BEGIN
         UPDATE cr_report set submitted = true where id = in_report_id;
-        PERFORM reconciliation__save_set(in_report_id, in_line_ids);
+        PERFORM reconciliation__save_set(in_report_id, in_line_ids, in_end_date);
 
         RETURN FOUND;
 END;
 $$ LANGUAGE PLPGSQL;
+
+COMMENT ON FUNCTION reconciliation__submit_set(
+        in_report_id int, in_line_ids int[], in_end_date date) IS
+$$Submits a reconciliation report for approval.
+in_line_ids is used to specify which report lines are cleared, finalizing the report.
+in_end_date is used to specify the clearing date. Defaults to now.$$;
 
 CREATE OR REPLACE FUNCTION reconciliation__check(in_end_date date, in_chart_id int)
 RETURNS SETOF defaults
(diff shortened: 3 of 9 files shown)


sql/modules/Reconciliation.sql
```diff
@@ -207,18 +231,24 @@ CREATE OR REPLACE FUNCTION reconciliation__report_approve (in_report_id INT) ret
...
-        UPDATE acc_trans SET cleared = TRUE
-        WHERE entry_id = any(ac_entries);
+        FOREACH ac_entry IN ARRAY ac_entries LOOP
+            UPDATE acc_trans
+               SET cleared = TRUE, cleared_on = clear_time, reconciled_on = now()
+              FROM cr_report_line
+             WHERE ledger_id = ac_entry
+               AND entry_id = ac_entry
+               AND report_id = in_report_id;
+        END LOOP;
```
ehuelsmann: Since the UPDATE doesn't depend on the value of `ac_entry`, why does this need to be a LOOP?

Wouldn't the `UPDATE .... WHERE entry_id = ANY(...)` be more efficient?

xt/42-reconciliation.pg
```diff
@@ -48,161 +47,272 @@ BEGIN;
...
---    SET value = 'Recon gl test '
---    WHERE setting_key = 'check_prefix';
-
-    PREPARE test AS SELECT reconciliation__new_report_id(test_get_account_id('-11111'), 100, now()::date, false);
-    SELECT results_eq('test',ARRAY[1],'Create Recon Report');
+    \set report_end1 '''2001-02-03'''
+    \set report_end2 '''2001-03-04'''
+    
+    --TODO: Do we still need this?
+    PREPARE test AS SELECT count(*) = 1
```
ehuelsmann: Maybe better to create issues for these TODOs and reference the issue number in the file?

xt/42-reconciliation.pg
```diff
@@ -48,161 +47,272 @@ BEGIN;
...
-    PREPARE test AS SELECT reconciliation__report_approve(currval('cr_report_id_seq')::int) > 0;
-    SELECT results_eq('test',ARRAY[true],'1 Report Approved');
+--    PREPARE test AS SELECT reconciliation__submit_set(currval('cr_report_id_seq')::int, (select as_array(id::int) from cr_report_line where report_id = currval('cr_report_id_seq')::int),:report_end1);
+--    SELECT throws_ok(
+--        'test',
+--        'P0001',
+--        'Unbalanced report by -20.00',
+--        'We should get a unique violation for an unbalanced report'
+--    );
+--    DEALLOCATE test;
```
ehuelsmann: Maybe document why this test is commented out, if you happen to know why?

sql/modules/Reconciliation.sql
```diff
@@ -1,15 +1,42 @@
...
+        PERFORM reconciliation__save_set(in_report_id, in_line_ids, in_end_date);
 
         RETURN FOUND;
 END;
 $$ LANGUAGE PLPGSQL;
+COMMENT ON FUNCTION reconciliation__submit_set(
+        in_report_id int, in_line_ids int[], in_end_date date) IS
+$$Submits a reconciliation report for approval.
+in_line_ids is used to specify which report lines are cleared, finalizing the report.
+in_end_date is used to specify the clearing date. Defaults to now.$$;
```
ylavoie: All cleared entries are declared cleared at the end of the report. I do not have a problem using a better name

xt/42-reconciliation.pg
ylavoie: Yes, as soon as we agree.

xt/42-reconciliation.pg
ylavoie: We should get, but we don't.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -429,6 +429,25 @@ push @tests, __PACKAGE__->new(
 );
 
 
+# They should be multiple selected from account_link_description(description)
+push @tests, __PACKAGE__->new(
+   test_query => "select distinct accno from chart
+                   where link is null or link = ''
+                   and charttype = 'A'",
+ display_name => $locale->text('Link unset in "chart" table'),
```
ylavoie: It is very possible. This test will have to be revised.

sql/modules/Reconciliation.sql
einhverfr: You would do better to use a FROM clause and add a join condition in the WHERE here.  With a lot of entries your approach will grind to a halt.  For performance reasons I think this needs to be changed.  Even correlated subqueries will do better here than a loop and single-item updates.  The reason is at some point you will have cache spillover and performance will screech to a halt.

sql/Pg-database.sql
```diff
@@ -1466,6 +1466,7 @@ CREATE TABLE acc_trans (
   entry_id SERIAL PRIMARY KEY
 );
 
+-- Drop this, no longer used
 ALTER TABLE cr_report ADD FOREIGN KEY (max_ac_id) REFERENCES acc_trans(entry_id);
```
ehuelsmann: max_ac_id was there to ensure that it wouldn't take 100 years to find the new entries and because the original author apparently thought it a safe criterion. It's highly unlikely that the criterion ever was actually safe, but it's very likely it worked in practice.

When removing the cache, we really need to introduce a different criterion (index, maybe?) which prevents the full transactions table from being scanned for each reconciliation. We can't *just* remove one thing and not replace it by another.

ylavoie: Yes it was but it is not relevant with that PR and doesn't even work.

sql/upgrade/sl2.8-1.5.sql
```diff
@@ -9,6 +9,397 @@ BEGIN;
 
 -- adding mapping info for import.
 
+CREATE OR REPLACE FUNCTION pg_temp.account__save
```
ehuelsmann: Is this required for the reconciliation?

lib/LedgerSMB/Scripts/recon.pm
```diff
@@ -204,17 +204,16 @@ sub _display_report {
     $recon->unapproved_checks;
 
     my $contents = '';
-    {
+    if ($request->{csv_file}) {
```
ehuelsmann: Does the `csv_file` field (still?) exist? I think it gfs been removed.

ehuelsmann:
All in all I still see many changes that need to be separate PRs:

- migration logging output fixes
- cancelable migrations
- deletable reconciliation check improvements
- migration improvements by use of copied stored procedures
- etc.
- etc.

sql/Pg-database.sql
```diff
@@ -1167,7 +1167,6 @@ CREATE TABLE cr_report (
     approved_by int references entity(id),
     approved_username text,
     recon_fx bool default false,
-    max_ac_id int,
```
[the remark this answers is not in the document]
ylavoie: max_ac_id makes all SL based migration of reconciliations fail because it assumes that no future submitted record exist. When we migrate, they all are there and that pulls all transactions in each cr_report.

locale/LedgerSMB.pot
```diff
@@ -9364,3 +9364,7 @@ msgstr ""
 #: lib/LedgerSMB/Scripts/payment.pm:1854 lib/LedgerSMB/Scripts/payment.pm:1871
 msgid "use of an overpayment"
 msgstr ""
+
+#: t/data/04-gettext.tex:8
+msgid "zh_CN"
+msgstr ""
```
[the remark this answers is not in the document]
ylavoie: Noise. Removed

sql/Pg-database.sql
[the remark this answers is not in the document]
ylavoie: max_ac_id usage ensure failure for SL migrations

ehuelsmann: If you don't create the column here, you can't drop it safely in the schema migration later on (can't drop a non-existing column?).

PR #2298, opened 2016-11-21
die with a human readable error message if we are run as root

sbts:
running as root, even if --user is passed to starman to drop privileges causes us grief as starman drops privileges too late, and we have already initialised things that will break after the drop

### DIFF
--- a/lib/LedgerSMB/Sysconfig.pm
+++ b/lib/LedgerSMB/Sysconfig.pm
@@ -422,7 +422,7 @@ if(!(-d LedgerSMB::Sysconfig::tempdir())){
 sub check_permissions {
     use English qw(-no_match_vars);
 
-    if($EUID == 1000){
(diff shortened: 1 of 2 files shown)


lib/LedgerSMB/Sysconfig.pm
```diff
@@ -422,6 +422,18 @@ if(!(-d LedgerSMB::Sysconfig::tempdir())){
 sub check_permissions {
     use English qw(-no_match_vars);
 
+    if($EUID == 0){
+        die_pretty( "Running LedgerSMB as root is a security problem",
```
ehuelsmann: While true, can we make this sound as if this is a general security probem and not specifically LedgerSMB related?

sbts: Shall do.

PR #2300, opened 2016-11-22
Display unapproved transactions before current report

ylavoie:
Add links to unapproved transactions and fix get routine


ehuelsmann:
While I understand that you took this route as a solution, I'm not really happy with the UI side of things. Looking at the screenshot, the message "There's something wrong" kind of disappears in the screenclutter. I'll send a mail about it on the mailing list for further discussion (as the tickets aren't ideal for that).

screenshot from 2016-12-26 12 33 28

ehuelsmann:
This PR has been replaced by PR #2482

PR #2325, opened 2016-11-29
Upgrade SL30 migration to 1.6

ylavoie:
1. Remove debugging leftover warning
2. Not all settings are defined in SL or previous LSMB databases . Make sure we do not fail migrations in that case.
3. SL didn't have uniqueness requirements for invoices. Suggest adjusted names to ease migration
4. Adjust SL30 migration code for 1.6

### DIFF
--- a/lib/LedgerSMB/Upgrade_Tests.pm
+++ b/lib/LedgerSMB/Upgrade_Tests.pm
@@ -704,7 +704,7 @@ push @tests, __PACKAGE__->new(
 # There's no AP uniqueness requirement?
 push @tests, __PACKAGE__->new(
     test_query => "SELECT id, concat(invnumber,'-',row_number() over(partition by invnumber order by id)) AS invnumber,
-                          transdate, duedate, datepaid, ordnumber, quonumber, approved
+                          dcn, description, transdate, duedate, datepaid, ordnumber, quonumber, approved
                      FROM ap
                     WHERE invnumber IN (SELECT invnumber
                                           FROM ap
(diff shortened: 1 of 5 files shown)


lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -703,20 +703,21 @@ push @tests, __PACKAGE__->new(
...
 push @tests, __PACKAGE__->new(
-    test_query => "select *
-                     from ap
-                    where invnumber in (select invnumber
-                                               from ap
-                                              group by invnumber
-                                              having count(*) > 1)
-                    order by invnumber",
-    display_name => $locale->text('Non-unique invoice numbers'),
+    test_query => "SELECT id, concat(invnumber,'-',row_number() over(partition by invnumber order by id)) AS invnumber,
```
ehuelsmann: @ylavoie with this change, we no longer show the user the problematic content. I'd expect that to be confusing more than that this commit is helpful (in the sense that this commit prevents data-entry)

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -703,20 +703,21 @@ push @tests, __PACKAGE__->new(
...
+                                        HAVING count(*) > 1)
+                 ORDER BY invnumber",
+    display_name => $locale->text('Non-unique invoice numbers detected'),
     name => 'no_duplicate_ap_invoicenumbers',
     display_cols => ['id', 'invnumber', 'transdate', 'duedate', 'datepaid',
                      'ordnumber', 'quonumber', 'approved'],
     column => 'invnumber',
  instructions => $locale->text(
-                   'Please make all AP invoice numbers unique'),
+                   'Please review suggestions to make all AP invoice numbers unique'),
```
ehuelsmann: Maybe we should list the conflicting data as well as the suggested data and explain something about that in the text above?

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: We still show all problematic entries, but added a suffix to make them unique.

ehuelsmann: Sure, but the entries don't look problematic anymore...

ylavoie:
This depends on #2288

PR #2330, opened 2016-12-02
Temporary fix for Travis issue between node.js and java

ylavoie:
Travis recently upgraded their images and brought an issue for java which prevents node.js to find it.
This is th recommended patch until a permanent fix is generated.


ehuelsmann:
This function has already been merged as part of another branch.

PR #2331, opened 2016-12-02
Test dojo unbuilt

ylavoie:


ehuelsmann:
@ylavoie I searched the sources for the DOJO_BUILT environment variable, but I can't find it. So, I'm not sure about the effect of this PR. Can you explain its effect?

PR #2341, opened 2016-12-04
Move sed code before LSMB start

ylavoie:


.travis.yml
```diff
@@ -116,6 +116,9 @@ before_script:
   - tar xzf tar-output/ledgersmb-travis.tar.gz
   - cd ledgersmb
   - cp conf/ledgersmb.conf.default ledgersmb.conf
+  - if [ "${DOJO_BUILT}" != "" ]; then
+      sed -r -i -e "s/^( |\t|#)*dojo_built( |\t)*=.+/dojo_built = ${DOJO_BUILT}/" ledgersmb.conf ;
```
sbts: )*=.+/
needs to be
)*=.*/

utils/test/Is_LSMB_running.sh
```diff
@@ -82,6 +82,10 @@ if $UPDATE; then
     cp $current $src
 fi
 
+# so we don't fail simply for dojo_built=0 (js-src) causing changes
+# modify the current file to appear to be only js instead of js-src
+sed -i.bak 's|"js-src/|"js/|' "$current"
```
sbts: This change isn't needed as it's already in place for master and 1.5

.travis.yml
```diff
@@ -122,6 +122,9 @@ before_script:
   - tar xzf tar-output/ledgersmb-travis.tar.gz
   - cd ledgersmb
   - cp conf/ledgersmb.conf.default ledgersmb.conf
+  - if [ "${DOJO_BUILT}" != "" ]; then
+      sed -r -i -e "s/^( |\t|#)*dojo_built( |\t)*=.*/dojo_built = ${DOJO_BUILT}/" ledgersmb.conf ;
+    fi
```
sbts: Why do we need to do this?
we have 2 ledgersmb.conf files to cover dojo built and unbuilt.
Also we can specify which file is used in an env-var

.travis.yml
```diff
@@ -122,6 +122,9 @@ before_script:
   - tar xzf tar-output/ledgersmb-travis.tar.gz
   - cd ledgersmb
   - cp conf/ledgersmb.conf.default ledgersmb.conf
```
sbts: Rather than copying the config file, we should simply set the LSMB_CONFIG_FILE environment variable

.travis.yml
ylavoie: They are out of sync. Maintaining them will always be a problem. I advocate using only one file and toggling the option.

sbts: I'll solve the maintenance issue with a test to confirm all 3 are in sync (default, unbuilt, and travis)
at the same time, I'll make sure they are all in synv

ylavoie:
Already handled.

PR #2360, opened 2016-12-07
Propose a buy exchange rate if none defined.

ylavoie:
Exchangerate varies constantly. Buy rates can be provided automatically buy the application by using a [web service](http://currencies.apps.grandtrunk.net) and providing current date and origin and destination currencies.
When the transaction is reconcilied, the sell rate effectively used would take into account the rate variation and conversion fees


ylavoie:
The Web service is free for personal and non-profit use only.
As discussed, we will need to make that an opt-in.

ehuelsmann:
Held for definition of the opt-in option.

PR #2381, opened 2016-12-12
Complete move toward person ID for TimeCards

ylavoie:
Affects 1.4 and 1.5, this move started in 2012.
Previously, we relied on the entity ID instead of the person ID. It so happens that most of the time, when starting a new company, employees are created first, then customers and vendors follows. This makes the person and entity match each other for a while, until new employees are added.

### DIFF
--- a/sql/modules/Person.sql
+++ b/sql/modules/Person.sql
@@ -14,6 +14,16 @@ $$ LANGUAGE SQL;
 COMMENT ON FUNCTION person__get_my_entity_id() IS
 $$ Returns the entity_id of the current, logged in user.$$;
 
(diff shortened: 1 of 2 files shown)


sql/modules/Person.sql
```diff
@@ -14,6 +14,16 @@ $$ LANGUAGE SQL;
 COMMENT ON FUNCTION person__get_my_entity_id() IS
 $$ Returns the entity_id of the current, logged in user.$$;
 
+CREATE OR REPLACE FUNCTION person__get_my_id() RETURNS INT AS
```
ehuelsmann: This one causes the blacklist test to fail: there's a new function signature now, which needs to be added to the black/whitelist.

PR #2384, opened 2016-12-13
Add timezones to PGDate

ylavoie:


lib/LedgerSMB/PGDate.pm
```diff
@@ -87,7 +87,7 @@ our $formats = {
 # Since we only match a limited set of patterns, below is what I
 # had expected DateTime::Format::Strptime would have done.
 my $regexes = {
-    'YYYY-MM-DD' => [ { regex => qr/^(\d{4,4})\-(\d\d)\-(\d\d)$/,
+    'YYYY-MM-DD' => [ { regex => qr/^(\d{4,4})\-(\d\d)\-(\d\d)\s?/,
```
ehuelsmann: Hmm. This matches a space character or nothing. However, what when the input is 2016-12-2016 ? That's invalid input and the pattern was meant to reject the input.

Along the same lines, we don't support parsing a full iso 8601 DateTime at the moment, do we? I think that it'd make more sense to add an additional pattern which matches a full date+time for iso 8601

lib/LedgerSMB/PGDate.pm
```diff
@@ -271,6 +274,7 @@ sub from_input{
     bless $dt, __PACKAGE__;
     $dt->is_date(1);
     $dt->is_time(($input && $input =~ /\:/) ? 1 : 0); # Define time
+    $dt->is_tz(($input && $input =~ /\:\d\d[\+\-]\d+/) ? 1 : 0); Define TZ
```
ehuelsmann: This line has a non-commented comment at the end: `Define TZ`

lib/LedgerSMB/PGDate.pm
```diff
@@ -294,14 +298,18 @@ sub to_output {
     $fmt = $formats->{uc($fmt)}->[0] if defined $formats->{uc($fmt)};
 
     $fmt .= ' %T' if $self->is_time();
+    $fmt .= '%z' if $self->is_tz();
     $fmt =~ s/^\s+//;
-
+#use Data::Printer;
+#warn p($self);
+#warn $fmt;
```
ehuelsmann: These lines can be (and should be) removed as they clutter the view on the code-flow. (Anybody can re-add them, if they want/need).

t/03-date-handling.t
```diff
@@ -335,6 +335,5 @@ is(LedgerSMB::PGDate->from_input(undef)->to_output, '',
    'round-tripping "undef" returns an empty string (for easy concatenation)');
 is(LedgerSMB::PGDate->from_input('2016-01-01')->to_output, '2016-01-01',
    'round-tripping valid ISO-8601 date returns that date');
-
-
-
+is(LedgerSMB::PGDate->from_input('2016-01-01 00:00:00-05')->to_output, '2016-01-01 00:00:00-0500',
+   'round-tripping valid ISO-8601 date+tz returns that date with normalized tz');
```
ehuelsmann: What about other round-tripping non-tz values?

t/03-date-handling.t
```diff
@@ -335,5 +335,5 @@ is(LedgerSMB::PGDate->from_input(undef)->to_output, '',
    'round-tripping "undef" returns an empty string (for easy concatenation)');
 is(LedgerSMB::PGDate->from_input('2016-01-01')->to_output, '2016-01-01',
    'round-tripping valid ISO-8601 date returns that date');
-is(LedgerSMB::PGDate->from_input('2016-01-01 00:00:00-05')->to_output, '2016-01-01 00:00:00-0500',
+is(LedgerSMB::PGDate->from_input('2016-01-01 01:23:45-06')->to_output, '2016-01-01 01:23:45-0600',
```
ehuelsmann: Your pattern and test use a space as the separator between date and time, but from https://en.wikipedia.org/wiki/ISO_8601, I see we should at least parse a format where the space you're currently allowing, is filled with the capital letter `T`

ylavoie:
After discussions, we decided against TimeZones in TimeCards. Closing this.

PR #2400, opened 2016-12-17
Add Upgrade test to validate that links are set

ylavoie:


lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -429,6 +429,25 @@ push @tests, __PACKAGE__->new(
 );
 
 
+# They should be multiple selected from account_link_description(description)
+push @tests, __PACKAGE__->new(
+   test_query => "select distinct accno from chart
+                   where link is null or link = ''
+                   and charttype = 'A'",
+ display_name => $locale->text('Link unset in "chart" table'),
```
ehuelsmann: I know enough books with at least one account without links. To be honest, I don't think this is an upgrade failure.

Can we go back to the problem this solved for you and find a better test/ upgrade fix? I'm thinking it would be unfair to block everybody with the books that I know of to migrate from SL to LSMB -- if they hadn't already been on lsmb...

ylavoie:
Retiring

PR #2407, opened 2016-12-17
Display unapproved transactions before current report

ylavoie:
Add links to unapproved transactions and fix get routine


ehuelsmann:
This PR has been replaced by PR #2482

PR #2408, opened 2016-12-17
Adjust warning styles

ylavoie:


ehuelsmann:
Could you rebase this one? It's now conflicted and the original diff is probably way too large.

ehuelsmann:
This branch could easily be merged, if the other changes to report.html would not be there...

ylavoie:
Agreed but the changes in style were for the report.

ylavoie:
Only styling left.

PR #2409, opened 2016-12-17
Show number of unapproved reconciliations in message

ylavoie:


ehuelsmann:
Could you rebase this one? I think the diff is (now) *much* smaller than it used to be, but I can't see the true/actual difference now.

ehuelsmann:
Looks like this one needs rebasing again... Sorry about that.

PR #2411, opened 2016-12-17
Prevent date wrapping

ylavoie:


UI/reconciliation/report.html
```diff
@@ -249,46 +248,50 @@
...
+                <td>
+                <?lsmb IF !approved and !submitted;
             INCLUDE input element_data = {
             type = "checkbox"
             name = "cleared_$row.id"
             value = row.id
             checked = row.cleared
             "data-dojo-type" = "lsmb/PublishCheckBox"
-            "data-dojo-props" = "topic:'ui/reconciliation/report/cleared_$row.id'"
+                        "data-dojo-props" = "topic:'ui/reconciliation/report/cleared_" _ row.id _ "'"
```
ehuelsmann: out of curiosity, why did you change this bit?

UI/css/ledgersmb.css
```diff
@@ -5,6 +5,22 @@ td.money {
...
+
+td.check {
+    text-align: center;
+}
+
+div#title {
+    font-size: large;
+}
+
+th.header {
```
ehuelsmann: Actually, isn't TH *always* used as a header in our code? That's its semantic meaning, isn't it? Shouldn't we then default to centered, with resetting to "initial" on more specific use cases?

PR #2418, opened 2016-12-20
Revamp template processors

ylavoie:
Fix header to propose a filename for output file
Move common parts to Template.pm and simplify
Restore support for XLS files


lib/LedgerSMB/Template.pm
```diff
@@ -270,26 +275,105 @@ sub _valid_language {
...
 
-    if ($type eq 'SCALAR' || !$type){
-        return;
-    }
-    if ($type eq 'ARRAY'){
-        for (@$vars){
-            if (ref($_)){
-                $self->_preprocess($_);
-            }
+    { # pre-5.14 compatibility block
```
ehuelsmann: 1.6 drops Perl pre-5.14 support. No need to have the block this way..

lib/LedgerSMB/Template.pm
```diff
@@ -270,26 +275,105 @@ sub _valid_language {
...
+      # Really? -YL
```
ehuelsmann: Yup. Really. Now, if you wonder what problem is being solved, there's a better palce to discuss that than in the code 😊.

The problem worked around btw, is with the localized reference to $@.

lib/LedgerSMB/Template/TXT.pm
```diff
@@ -117,7 +117,6 @@ sub preprocess {
 sub escape {
     my $vars = shift @_;
     return undef unless defined $vars;
-    $vars = escapeHTML($vars);
```
ehuelsmann: Why is this an incorrect action?

ylavoie: Why does TXT has to be escaped in HTML?

lib/LedgerSMB/Template.pm
```diff
@@ -278,14 +278,9 @@ sub _preprocess {
     my ($rawvars, $escape) = @_;
     return undef unless defined $rawvars;
 
-    { # pre-5.14 compatibility block
-      # Really? -YL
-        local ($@); # pre-5.14, do not die() in this block
-        if (eval {$rawvars->can('to_output')}){
-            $rawvars = $rawvars->to_output;
-        }
+    if (eval {$rawvars->can('to_output')}){
```
ehuelsmann: This is now missing a

```perl
    local ($@);
```
The thing about the compat block is that the local has a tightly restricted scope, which is no longer required.

lib/LedgerSMB/Report.pm
```diff
@@ -248,10 +248,7 @@ sub _render {
     $self->run_report($request) if !defined $testref;
     # This is a hook for other modules to use to override the default
     # template --CT
-    { # pre-5.14 compatibility block
-        local ($@); # pre-5.14, do not die() in this block
-        eval {$template = $self->template};
-    }
+    eval {$template = $self->template};
```
ehuelsmann: Same remark about the removal of the compat block

lib/LedgerSMB/Template/XLS.pm
```diff
@@ -0,0 +1,255 @@
...
+
+package LedgerSMB::Template::XLS;
+
+use strict;
+use warnings;
+
+use Template;
+use Template::Parser;
+use LedgerSMB::Template::TTI18N;
+use CGI::Simple::Standard qw(:html);
```
ehuelsmann: What's CGI::Simple::Standard for? I couldn't find it.

lib/LedgerSMB/Template/XLS.pm
```diff
@@ -0,0 +1,255 @@
...
+
+=cut
+
+package LedgerSMB::Template::XLS;
+
+use strict;
+use warnings;
+
+use Template;
+use Template::Parser;
```
ehuelsmann: What is Template::Parser used for?

lib/LedgerSMB/Template/XLS.pm
ylavoie: It pulls escapeHTML

lib/LedgerSMB/Template/XLS.pm
ylavoie: Leftover. Removed. Same for Data::Dumper

lib/LedgerSMB/Template/XLS.pm
```diff
@@ -0,0 +1,255 @@
...
+
+sub preprocess {
+    my $rawvars = shift;
+    return LedgerSMB::Template::_preprocess($rawvars, \&escape);
+}
+
+sub escape {
+    my $vars = shift @_;
+    return undef unless defined $vars;
+    $vars = escapeHTML($vars);
```
ehuelsmann: Maybe add a comment inthe code as to why we need to html escape in an xls file...

lib/LedgerSMB/Template/XLS.pm
```diff
@@ -149,13 +149,11 @@ sub _xls_process {
...
-            italic => sub { &_named_format('italic', @_) },
-            shadow => sub { &_named_format('shadow', @_) },
-            strikeout => sub { &_named_format('strikeout', @_) },
-=cut
+#            bold => sub { &_named_format('bold', @_) },
+#            hidden => sub { &_named_format('hidden', @_) },
+#            italic => sub { &_named_format('italic', @_) },
+#            shadow => sub { &_named_format('shadow', @_) },
+#            strikeout => sub { &_named_format('strikeout', @_) },
             },
```
ehuelsmann: Is it vital to retain the comment block? If so, maybe add a comment as to why? In general, I'm a big believer of thorough clean-up

lib/LedgerSMB/Template.pm
```diff
@@ -213,7 +214,7 @@ sub new {
     $self->{debug} = $args{debug};
     $self->{binmode} = undef;
     $self->{outputfile} =
-        "${LedgerSMB::Sysconfig::tempdir}/$args{output_file}" if
+        "${LedgerSMB::Sysconfig::tempdir}/${LedgerSMB::Sysconfig::cache_template_subdir}/UI/$args{output_file}" if
```
ylavoie: We may have to revise this then, because ${LedgerSMB::Sysconfig::tempdir}/$args{output_file} doesn't lead to a valid directory.

lib/LedgerSMB/Template/XLS.pm
ylavoie: Do we want to refer to named formats in XLS files? If not, this goes. If we do, then I have to find a way to implement.

lib/LedgerSMB/Template.pm
```diff
@@ -181,13 +190,16 @@ sub available_formats {
     local ($@);
     if (eval {require LedgerSMB::Template::LaTeX}){
         push @retval, 'PDF', 'PS';
-    }
+    } { warn $@ if $@; }
```
ehuelsmann: hmm. why did you add the block around `warn` here? the above is equivalent to:
```perl
  }
  warn $@ if $@;
```

But, more importantly, we want to log it instead of warn, because warnings go to STDERR and loggers go everywhere our users want to, I'd say.

lib/LedgerSMB/Template.pm
```diff
@@ -270,26 +282,104 @@ sub _valid_language {
...
+    local ($@);
+    if (eval {$rawvars->can('to_output')}){
+        $rawvars = $rawvars->to_output;
     }
-    if ($type eq 'ARRAY'){
-        for (@$vars){
-            if (ref($_)){
-                $self->_preprocess($_);
-            }
+    use LedgerSMB;
```
ehuelsmann: Hmm. We'd want this use statement at the top of the file.

But more importantly, what do we need the LedgerSMB module for specifically? I'm not seeing it being used below...

ylavoie: use LedgerSMB removed.

PR #2444, opened 2016-12-28
Master issue 2376

ehuelsmann:


ehuelsmann:
Too early; was planning to submit a different PR.

PR #2475, opened 2017-01-02
Revamp timecards

ylavoie:
- Complete adding of Time, Materials and Overhead cards
- Allow Edit and Delete of unprocessed cards
- Fix import of CSV cards
- Fix Order generations of cards.
- Add TimeZone support to PGDates to support TimeCards

### DIFF
--- a/UI/js-src/lsmb/lsmb.profile.js
+++ b/UI/js-src/lsmb/lsmb.profile.js
@@ -108,6 +108,7 @@ var profile = (function(){
                     "lsmb/SubscribeSelect",
                     "lsmb/SubscribeShowHide",
                     "lsmb/TabularForm"
+                    "lsmb/Timecard",
                 ]
             }
         },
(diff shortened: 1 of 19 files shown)


ylavoie:
The PGDate modifications are also on another PR, they have to be removed from here before merging.

UI/js-src/lsmb/lsmb.profile.js
```diff
@@ -108,6 +108,7 @@ var profile = (function(){
                     "lsmb/SubscribeSelect",
                     "lsmb/SubscribeShowHide",
                     "lsmb/TabularForm"
```
ehuelsmann: this line wants a terminating comma

UI/js-src/lsmb/lsmb.profile.js
```diff
@@ -108,6 +108,7 @@ var profile = (function(){
...
+                    "lsmb/Timecard",
```
ehuelsmann: And this line doesn't want one.

PR #2476, opened 2017-01-02
Add diagnostic message when action is not defined

ylavoie:


UI/js-src/lsmb/Form.js
```diff
@@ -40,9 +40,12 @@ define([
                                  ? "GET"
                                  : this.method;
                       var url = this.action;
-
                       var options = { "handleAs": "text" };
                       if ("get" === method.toLowerCase()){
+                          if (!url) {
+                              alert('Form contains no action. Please file a bug');
```
ehuelsmann: Maybe we can be more helpful by telling them where? (the GitHub URL?)

PR #2495, opened 2017-01-04
Proxy grandtrunk.net to get FXrates

ylavoie:


old/lib/LedgerSMB/Form.pm
```diff
@@ -1805,7 +1813,7 @@ sub get_name {
 
     # SC: Check for valid table/view name.  Other values will cause SQL errors.
     if ($table !~ /^(vendor|customer|employee)$/i) {
-        $self->error('Invalid name source');
+        $self->error('Invalid name source "$table"');
```
ehuelsmann: This will return the string

```
Invalid name source "$table"
```

because single-quoted strings don't get their variables replaced.

lib/LedgerSMB/PSGI.pm
```diff
@@ -227,6 +227,8 @@ sub setup_url_space {
              pod_view => 'Pod::POM::View::HTMl' # the default
                  if $development;
 
+        mount '/getrate' => Plack::App::Proxy->new(remote => 'http://currencies.apps.grandtrunk.net/getrate')->to_app;
```
ehuelsmann: This opens up the server as an anonymous proxy from anywhere on the web. I don't like that. Also, it does not keep a local cache (in the exchange rate table) for any values already requested.

This leads to inconsistencies (potentially) with the exchangerate table and the values returned by this service.

ehuelsmann:
In general, my problem with this PR is the same as with the other one: there's nowhere where we point the user to the efforts of the source, which explicitly asks for donations. I don't think this respects their efforts enough.
We should at least make this opt-in in order to be able to refer the user to the data source and become aware of the plea for donations.

lib/LedgerSMB/PSGI.pm
```diff
@@ -199,6 +208,17 @@ sub _run_old {
     }
 }
 
+sub _fxrate_app {
+    my $env = shift;
+    my $session = $env->{'psgix.session'};
+    if ( $session && $session->{valid}) {
```
ehuelsmann: From the diff I think I can see this doesn't work. Is this tested?

ylavoie: This does work, psgix.session is written behind the scene by a memory write, so that's why it seems to appear from nowhere. Tested and proven correct.

ehuelsmann:
Hmm. But does that work on the forking server that starman is?

PR #2503, opened 2017-01-06
Prevent crash on application timeout, fixes #2502

### LINKED ISSUE #2502: Prevent crash on application timeouts
Instead of seeing the logged out dialog, one is often shown 'Can't call method prepare on an undefined value at /usr/local/share/perl/5.22.2/PGObject.pm line 206, <DATA> line 1.
at /usr/local/share/perl/5.22.2/PGObject.pm line 206'

ylavoie:
Report gracefully when action is not defined.

### DIFF
--- a/lib/LedgerSMB/PSGI.pm
+++ b/lib/LedgerSMB/PSGI.pm
@@ -161,8 +161,14 @@ sub psgi_app {
         }
 
         $LedgerSMB::App_State::DBH = $request->{dbh};
-        ($status, $headers, $body) = @{&$action($request)};
-
+        if ( defined $action ) {
+            ($status, $headers, $body) = @{&$action($request)};
+        } else {
+            ($status, $headers, $body) = 
+                      (500, 
+                       ['text/html'],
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/PSGI.pm
```diff
@@ -161,8 +161,14 @@ sub psgi_app {
         }
 
         $LedgerSMB::App_State::DBH = $request->{dbh};
-        ($status, $headers, $body) = @{&$action($request)};
-
+        if ( defined $action ) {
```
ehuelsmann: This check is in row 133; I don't expect this to be triggred.

ehuelsmann: The cause of the failure is a bit deeper: even though there is no database, something tries to access the database. I identified that to be Template.pm. See my PR.

ehuelsmann:
A fix to do exactly this is in 1.5 now. I'm afraid this wasn't the fix. Closing.

PR #2518, opened 2017-01-07
Fix timecard csv import

ylavoie:
- Add correct separators for CSV import
- Use proper columns
- Use user currency

### DIFF
--- a/lib/LedgerSMB/Scripts/import_csv.pm
+++ b/lib/LedgerSMB/Scripts/import_csv.pm
@@ -12,21 +12,24 @@ data.
 =cut
 
 package LedgerSMB::Scripts::import_csv;
-use Moose;
-use LedgerSMB::Template;
-use LedgerSMB::Form;
 use strict;
 
-my $default_currency = 'USD';
+use Moose;
+
+use LedgerSMB::Template;
+use LedgerSMB::Form;
+use LedgerSMB::Setting;
+
+my $default_currency = 'USD'; #LedgerSMB::Setting->get('curr');
 our $cols = {
    gl       =>  ['accno', 'debit', 'credit', 'source', 'memo'],
    ap_multi =>  ['vendor', 'amount', 'account', 'ap', 'description',
                  'invnumber', 'transdate'],
    ar_multi =>  ['customer', 'amount', 'account', 'ar', 'description',
                  'invnumber', 'transdate'],
-   timecard =>  ['employee', 'projectnumber', 'transdate', 'partnumber',
-                 'description', 'qty', 'noncharge', 'sellprice', 'allocated',
-                'notes'],
+   timecard =>  ['employee', 'business_unit_id', 'transdate', 'partnumber',
+                 'description', 'qty', 'non_billable', 'sellprice', 'allocated',
+                'notes', 'jctype','curr'],
    inventory => ['partnumber', 'onhand', 'purchase_price'],
    inventory_multi => ['date', 'partnumber', 'onhand', 'purchase_price'],
 };
@@ -470,6 +473,7 @@ This displays the begin data entry screen.
 
 =cut
 
+use Data::Printer;
 sub begin_import {
     my ($request) = @_;
     my $template_file =
@@ -503,6 +507,8 @@ data in $request and processes it according to the dispatch tables.
 
 sub run_import {
     my ($request) = @_;
+warn p($request);
+
     my @entries = parse_file($request);
     my $headers = shift @entries;
     if (ref($preprocess->{$request->{type}}) eq 'CODE'){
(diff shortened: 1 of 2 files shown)


lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -12,21 +12,24 @@ data.
...
 our $cols = {
    gl       =>  ['accno', 'debit', 'credit', 'source', 'memo'],
    ap_multi =>  ['vendor', 'amount', 'account', 'ap', 'description',
                  'invnumber', 'transdate'],
    ar_multi =>  ['customer', 'amount', 'account', 'ar', 'description',
                  'invnumber', 'transdate'],
-   timecard =>  ['employee', 'projectnumber', 'transdate', 'partnumber',
-                 'description', 'qty', 'noncharge', 'sellprice', 'allocated',
-                'notes'],
+   timecard =>  ['employee', 'business_unit_id', 'transdate', 'partnumber',
```
ehuelsmann: Instead of using the business_unit_id, I'd rather use the class and control code here and look that up during processing to get to the business_unit_id: much better for the user.

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -352,14 +356,23 @@ sub _process_sic {
...
     my ($request, $entries) = @_;
-    my $myconfig = {};
-    my $jc = {};
+    my @floats = qw {qty non_billable sellprice allocated};
     for my $entry (@$entries) {
+        my $jc = {};
         my $counter = 0;
         for my $col (@{$cols->{timecard}}){
+            if ($request->{sep} eq ";" &&
+                any { $_ eq $col } @floats) {
```
ehuelsmann: Can you explain what this bit is looking for? It doesn't ring a bell for me.

ylavoie: Floats are using comma instead of period when the CSV separator is the semi-column

lib/LedgerSMB/Scripts/import_csv.pm
ylavoie: It will be corrected by the other PR, along with the rest of the TC functionality.

ehuelsmann: Ok. Since both will land on master shortly, that's acceptable, but it's better to group for the best user experience.

lib/LedgerSMB/Scripts/import_csv.pm
ehuelsmann: Actually, `$request->{sep}` isn't initialised anywhere.

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -428,32 +441,41 @@ our $process = {
...
+    my $n = @{$cols->{$self->{type}}}-1; # 1 separator less than fields
 
     $self->{import_entries} = [];
     for my $line (split /(\r\n|\r|\n)/, $contents){
-        next if ($line !~ /,/);
+        if ( $sep eq "" ) {
+            if    ($n == (() = $line =~  /,/g)) { $sep =  "," }
+            elsif ($n == (() = $line =~  /;/g)) { $sep =  ";" }
+            elsif ($n == (() = $line =~ /\t/g)) { $sep = "\t" }
+            else  { die "Unrecognized file format" }
```
ehuelsmann: When using heuristics like this, there are many occasions where we can go wrong. I'd prefer to "simply" use settings from upload dialog; after all, this is an interface type of activity and that's exactly where you want things to be deterministic.

ylavoie: We do not have settings for CSV field separators and assume comma here before, bringing havoc too

ylavoie: And I decide on headers for the separator and we know there are none in the words, so heuristic felt safe

ehuelsmann: Yes. Absolutely :-) especially when importing files where comma is the decimal separator. However, the format __is__ called Comma Separated Values...

At this point, I have no idea if there is an override for the number formats, but I think we should offer the same number formats as we do on the Preferences page and offer the user's own defaults -- basically introducing an override mechanism.

ehuelsmann: regarding detection on the headers: ok. you have a good point.

ylavoie: It is called Comma Separated Values, and RFC 4180 was an attempt to standardize, it never made it globally. CSV has been abused by many vendors and export tools for Delimiter Based Fields (typically a single reserved character such as comma, semicolon, or tab; sometimes the delimiter may include optional spaces). See https://en.wikipedia.org/wiki/Comma-separated_values for a description.

PR #2526, opened 2017-01-10
Restrict date field to gl and inventory

ylavoie:


UI/import_csv/import_csv.html
```diff
@@ -44,6 +44,7 @@
                  value = request.description
                  } ?>
         </div>
+        <?lsmb IF request.type == "gl" || request.type == "inventory"  ?>
```
ehuelsmann: Would it be an idea to put this in the controller (import_csv.pl) instead of in the view (UI/ template)?

ehuelsmann: The criterion, that is.

ylavoie: Yes it would.

PR #2589, opened 2017-01-27
Use uglify to compress Dojo

ylavoie:


UI/js-src/lsmb/lsmb.profile.js
```diff
@@ -18,8 +18,8 @@ var profile = (function(){
         releaseDir: "../../js",
         releaseName: "",
         action: "release",
-        layerOptimize: "closure.keeplines",
-        optimize: "closure",
+        layerOptimize: "uglify.keeplines",
+        optimize: "uglify",
```
ehuelsmann: with your current measurements that uglify is much smaller (400k?), I'm inclined to go with Uglify. One thing: I think we decided in the channel that now is a good time to drop the 'keeplines' strategy.

PR #2591, opened 2017-01-29
HTML Checks

ylavoie:
Upgrade to HTML5
Fix missing and misplaced tags.


UI/am-list-departments.html
```diff
@@ -32,7 +32,7 @@
   END -?>
 <tr class="<?lsmb c ?>">
   <?lsmb FOREACH column IN columns ?>
-  <td class="<?lsmb c ?>" <?lsmb IF (column == 'cost' || column == 'profit') ?>align="center"<?lsmb ELSIF column == 'discount' ?>align="right"<?lsmb END ?>>
+  <td class="<?lsmb c ?>" align=<?lsmb (column == 'cost' || column == 'profit') ? "center" : column == 'discount' ? "right" : "" ?>>
```
ehuelsmann: `"center"` returns `center` into the template, leading to `align=center` instead of `align="center"`; I'd propose to return `'"center"'` instead (same for `"right"`)

PR #2594, opened 2017-01-29
* Fix Sauce tests by switching APIs

ehuelsmann:


ehuelsmann:
Needs an upstream fix first.

Issue #2275, opened 2016-11-16
Error: Can't locate lsmb-request.pl in @INC


jame:
On Debian systems with perl-base version 5.22.2-4 or newer, LedgerSMB will cause the following type of error to show up in the Apache error.log when attempting to log in or to run setup.pl (example is from setup.pl):

> Can't locate lsmb-request.pl in @INC (@INC contains: /etc/perl/usr/local/lib/i386-linux-gnu/perl/5.24.1 /usr/local/share/perl/5.24.1 /usr/lib/i386-linux-gnu/perl5/5.24 /usr/share/perl5 /usr/lib/i386-linux-gnu/perl/5.24 /usr/share/perl/5.24 /usr/local/lib/site_perl /usr/lib/i386-linux-gnu/perl-base) at
> /usr/share/ledgersmb/setup.pl line 8.
> 

This was found on a current version of Debian Testing ('Stretch'). using the Debian package for LedgerSMB v1.4.34-2.

jame:
A work around is that there is a line in the /etc/perl/sitecustomize.pl file on Debian systems that can be commented out in order to **not** remove the "." (current directory)  from the Perl @INC path.

jame:
A bug has been submitted in Debian and forwarded here;  reference:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=844276

jame:
And there may not be work around on any system (not just Debian) with Perl v5.24, as I understand that not having the current directory in @INC is compiled into the interpreter.

jame:
One other work around can be to run it under starman instead of apache (like how version v1.5 of LedgerSMB is going to be), which can be done as simply as by changing to the /usr/share/legersmb directory and then running something like the following:

> starman -I. tools/starman.psgi

ehuelsmann:
Fixed in all versions with community maintenance (1.4+) and for both psgi as well as cgi versions.

Released in 1.4.39 and 1.5.3.

jame:
It was thought that this issue had been resolved with changes implemented in LegderSMB v1.4.39/1.4.40, but when I tested a new package for v1.4.40 on a Debian Stretch ("Testing") system, which now has Perl v5.24 installed, I found that the error is still or again present.

jame:
Removed the **Has Workaround** tag because the one known main workaround does not work in Debian Stretch with Perl v5.24.
[...]

PR #2609, opened 2017-02-07
* Follow-up to CVE-2016-1238: '.' no longer in @INC

ehuelsmann:


ehuelsmann:
Resolves #2275 for 1.5.

PR #2614, opened 2017-02-09
Use Sysconfig to get the proper database location

ylavoie:


xt/41-coaload.t
```diff
@@ -1,6 +1,7 @@
 use Test::More;
 use strict;
 use warnings;
+use LedgerSMB::Sysconfig;
```
ehuelsmann: Can you explain this change? This file does not depend on Sysconfig.

ylavoie: This isn't completed. We assume that the database is always on the LSMB machine. Environment variables used for tests specify the database name, but we should get PG address from Sysconfig.

ehuelsmann: Hmm. I'm not sure I understand the remark that we assume that the LSMB machine is local. We documented as much in https://github.com/ledgersmb/LedgerSMB/tree/master/t#environment-variables : we expect the tests to be passed the database configuration through environment variables. If you want a non-local database, simply set PGHOST and PGPORT.

sbts: @ehuelsmann actually, the required PGHOST and PGPORT variables aren't documented there.
and as we've just been discussing in #ledgersmb there was no default value applied to either of those env vars which broke running with no ledgersmb.conf file

I'd say that #2635 will supersede this PR as it correctly sets defaults from and the env-vars from ledgersmb.conf

#2635 doesn't however add PGHOST and PGPORT to  https://github.com/ledgersmb/LedgerSMB/tree/master/t#environment-variables

ehuelsmann: Ok. I'll add PGHOST and PGPORT in that section and, after merging #2635, close this one.

ehuelsmann:
Superseeded by #2634

PR #2619, opened 2017-02-10
Silence Sysconfig redefines

ylavoie:


ehuelsmann:
I made the same change, but with a much stricter scope. Closing as done. Thanks!

PR #2647, opened 2017-03-01
Removing log warnings aout removed js files not found

einhverfr:
These js files no longer exist which makes debugging the setup.pl a but harder.


einhverfr:
These files are already removed from 1.5

PR #2651, opened 2017-03-01
fixed a bizarre bug on CentOS

einhverfr:
This may be due to some strange scoping bug?

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -528,9 +528,9 @@ sub load_templates {
     _init_db($request);
     my $dbh = $request->{dbh};
     opendir(DIR, $dir);
-    while (readdir(DIR)){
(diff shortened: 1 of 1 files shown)


ehuelsmann:
Hi Chris,
Can you provide some details to the bug? What problem were you seeing that this resolves? (Both for when we find new problems traced back to this commit as well as when we're helping others; maybe we see the same symptoms?)

Thanks!

Erik.

einhverfr:
On CentOS with Perl 5.10 the setup.pl seems to run through completely but the templates never load..  Try to load from setup.pl and they don't load again.

Apparently the $_ was remaining undefined in the while loop and an explicit assignment fixed the problem.

einhverfr:
A side point is....

Using $_ in a loop like that is going to be fragile anyways so maybe there are other gotchas or timebombs lurking there too.

PR #2652, opened 2017-03-01
fixed a bizarre bug on CentOS

einhverfr:
This may be due to some strange scoping bug?


einhverfr:
sorry did not want to merge it yet, but will monitor tests just in case I broke the build :-P

PR #2660, opened 2017-03-03
Master no inv entity

ehuelsmann:


Changelog
```diff
@@ -9,6 +9,7 @@ Code cleanup
 * Removed code duplication in
   - Template handling framework (Yves L)
 * Remove unused variables and unreachable code (Nick P)
+* Removed assumptions about being a webapp from libraries (Erik H)
```
sbts: Perhaps you need to also add an entry for the fixes in this PR ;-)

lib/LedgerSMB/Scripts/asset.pm
```diff
@@ -912,6 +912,39 @@ CSV formats.
 
 =cut
 
+sub _import_file {
```
sbts: is this the best naming for this function?
It appears the function is specific to importing a CSV file, so perhaps the function name should reflect that

PR #2711, opened 2017-04-09
* Clean up parallel test experiment remnants

ehuelsmann:


ylavoie:
ForkManager was no remnant. It it optional for Test::Class::Moose but removing it prevent parallel runs.

PR #2722, opened 2017-04-17
Extend Perl::Critic tests

nick-prater:
* Added some of the tests recommended by CERT
* Replace wildcard exclusions with specific tests


nick-prater:
This patch contains a duplicate policy. Closing to submit a revised patch.

PR #2734, opened 2017-04-18
T debugpane

ylavoie:
Fixes for the debug pane


tools/starman-development.psgi
```diff
@@ -57,3 +62,16 @@ builder {
 };
 
 # -*- perl-mode -*-
+
```
ehuelsmann: merging the entire PR, but I think this function can be removed if we can reduce the list of directories being monitored.

Issue #815, opened 2015-09-11
More preferences (preference country)


ehuelsmann:
As per Rob Ransbottom's suggestion on the mailing list:

``` plain
It would be good to have a default country for forms per user preference.
```

Note that the United States is all the way down on most drop-downs to understand the importance of his request.

einhverfr:
What other preferences do we need?

einhverfr:
From discussion, default printer, default format, and default attachment for dropdown

einhverfr:
See also https://sourceforge.net/p/ledger-smb/feature-requests/310/

pbolger:
It'd also be good to be able to limit the visible countries in the country dropdowns. Many users won't ever select Afghanistan (for example) so having to go past it on the list is an unecessary distraction.

ehuelsmann:
@pbolger I was going to react that you can reduce the list from the ````System```` menu, but as it turns out I'm mixing Country listing with Languages listing. The latter allows removal.

ehuelsmann:
@ylavoie I guess the enhancement we were talking about today (reduction of the number of countries) has an issue filed already.

ehuelsmann:
User-level preferences should be made more flexible before we can implement this one => depends #776

PR #2739, opened 2017-04-21
Localized database migration

ylavoie:
1- Complete the country list
2- Keep the selected country selected at COA available for the new_user creation
3- Add default date format by locale to present a suitable default format


sql/changes/1.6/country_date_format.sql
```diff
@@ -0,0 +1,267 @@
+ALTER TABLE country
+  ADD date_format TEXT;
```
ehuelsmann: Hi, wondering why we need this. Dojo knows all this, correct? If so, can't we leverage that?

ylavoie: 1- Dojo must know at start of application which languages will be used
2- Even if Dojo can display, convert, calculate on dates, it doesn't seem to have a way to show a date format by country of locale.

ylavoie:
This PR is probably only an interim solution, for those fields should be common and available by other means. Nevertheless, it addresses part of #815.

UI/setup/select_coa.html
```diff
@@ -24,11 +24,19 @@
            label = text('Chart of accounts')
     };
 ELSE;
+    lang = '';
+    IF languages;
+        lang = languages.pop;
+        IF lang.search('-');
+            lang = lang.split('-').1;
+        END;
+    END;
```
sbts: For the record, this is the code that angular JS uses for the task.
```javascript
var getFirstBrowserLanguage = function () {
    var nav = window.navigator,
    browserLanguagePropertyKeys = ['language', 'browserLanguage', 'systemLanguage', 'userLanguage'],
    i,
    language;

    // support for HTML 5.1 "navigator.languages"
    if (Array.isArray(nav.languages)) {
      for (i = 0; i < nav.languages.length; i++) {
        language = nav.languages[i];
        if (language && language.length) {
          return language;
        }
      }
    }

    // support for other well known properties in browsers
    for (i = 0; i < browserLanguagePropertyKeys.length; i++) {
      language = nav[browserLanguagePropertyKeys[i]];
      if (language && language.length) {
        return language;
```

sbts: As an additional layer.
Accept-Language header only viewable server side provides a prioritised list of Lang's preferred by the users browser setup

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -322,7 +320,6 @@ sub copy_db {
     my ($request) = @_;
     my ($reauth, $database) = _get_database($request);
     return $reauth if $reauth;
-
```
sbts: From a readability perspective Id like to keep the blank line after the returns. Helps to visually identify them

UI/setup/select_coa.html
ylavoie: Done with HTTP::AcceptLanguage instead in LedgerSMB.psgi

sbts: I agree, that HTTP::AcceptLanguage will do the job, but it then makes our code dependant on being delivered via an unmolested browser <==> server interface.
There are security helpers out there (especially in today's world of Metadata retention) that strip information like this from the Headers.
Hence my suggestion to also include it in the JS
We may want to compare the 2 results and prompt the user to select if they differ

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Agreed but not involved in this PR

cpanfile
```diff
@@ -103,6 +104,7 @@ feature 'debug', "Debug pane" =>
 # will be included, so put our testing requirements in develop...
 on 'develop' => sub {
     requires 'App::Prove', '3.36';
+    requires 'File::Find::Rule::Perl';
```
ehuelsmann: @ylavoie where was this extra dependency introduced? I can't find it in the diff...

cpanfile
```diff
@@ -130,4 +132,5 @@ on 'develop' => sub {
     requires 'Weasel', '0.11';
     requires 'Weasel::Driver::Selenium2', '0.05';
     requires 'Weasel::Widgets::Dojo';
+    requires 'Weasel::FindExpanders';
```
ehuelsmann: @ylavoie Where was this extra dependency introduced? I can't find it in the diff...

t/.pherkin.yaml
```diff
@@ -23,7 +23,7 @@ default:
              retry_timeout: 30
              driver:
                 drv_name: Weasel::Driver::Selenium2
-                wait_timeout: 3000
+                wait_timeout: 31000
```
ehuelsmann: What's the reason to increase the wait time-out? Does it need to be increased by a factor 10? I know about the Dojo components timing out every now and then, but there were fixes for that on master recently. Maybe we don't need this increase anymore?

sql/Pg-database.sql
```diff
@@ -271,9 +271,11 @@ INSERT INTO country(short_name,name) VALUES ('BG','Bulgaria');
 INSERT INTO country(short_name,name) VALUES ('BH','Bahrain');
 INSERT INTO country(short_name,name) VALUES ('BI','Burundi');
 INSERT INTO country(short_name,name) VALUES ('BJ','Benin');
+INSERT INTO country(short_name,name) VALUES ('BL','St. Barthelemy');
```
ehuelsmann: With the introduction of the `sql/changes/` schema management, I consider this file frozen. Even more, I think we should try and remove all data initialization from the schema creation file, because, well, people way well want to *just* create the schema (or *just* load the data into the schema)...
The step to remove the data from the schema isn't the scope of this change, but *not* changing the schema file is probably a good thing. We should add the above explanation to the top of the schema file to help other developers.

ehuelsmann: How do these additional country definitions get into existing databases? Pg-database.sql isn't run on existing databases...

sql/changes/1.6/country_date_format.sql
ehuelsmann: I know Dojo must know from the start of the application which language will be used. We can tell it which language it should use or it can detect itself and load the requested locale data from the server.
If it detects the locale itself, it will also select the appropriate date setting. However, we don't allow Dojo to use its own detection: we provide our own override in the page header.

xt/66-cucumber/01-basic/setup.feature
```diff
@@ -171,9 +171,9 @@ Scenario: Create database with ampersand in the name
       | First Name         | A                |
       | Last name          | Dmin             |
       | Employee Number    | 00000001         |
+      | Country            | United States    |
       | Date of Birth      | 09/01/2006       |
       | Tax ID/SSN         | 00000002         |
-      | Country            | United States    |
```
ehuelsmann: Can we create a test with a localized version as well? Maybe a French or German version?

That'll guarantee our support stays in place.

ehuelsmann:
This is where the NLS information for date formats, currencies  and number formats is in Dojo: https://github.com/dojo/dojo/tree/master/cldr/nls

When dojo is packaged with our dist, we package this too, to my knowledge.

sbts:
@ehuelsmann I have a script almost done that should allow us to have both "sourced" country SQL inserts and translated country names.
There are 2 sources, one for country data, and the other for country name translations.
Both appear to be long running, regularly updated, open projects on github that pull their data from multiple sources so should be perfect for us.

I'll get that completed (mainly testing left from memory) once I'm back on deck next week

t/.pherkin.yaml
ylavoie: Local problems, It should not have made it through the PR

xt/66-cucumber/01-basic/setup.feature
ylavoie: Done

cpanfile
ylavoie: Leftover. Removing

UI/js-src/lsmb/PublishSelect.js
```diff
@@ -17,6 +17,11 @@ define(["dojo/_base/declare",
                               self.publish(targetValue);
                           })
                    );
+               },
+               startup: function() {
```
ehuelsmann: Why do you need this change? Most widgets don't need a startup function and I'm not seeing anything happen here which the regular startup function isn't already doing...

ehuelsmann:
This commit changes the dojo, dijit and util submodules. A change like that *must* be outside the scope of this change. Please make sure the PR is submitted cleanly.

UI/js-src/lsmb/DateTextBox.js
```diff
@@ -24,8 +24,6 @@ define([
                params.placeholder = lsmbConfig.dateformat;
             }
             // end of 'old code' support block
-
-            this.inherited(arguments);
```
ehuelsmann: Why should the parent class of this widget *not* be initialized?

ylavoie:
Dojo changes will be removed from this PR

xt/01.2-deps.t
```diff
@@ -34,8 +34,10 @@ find(\&collect, 'lib/', 'old/bin/', 'old/lib/');
 
 push @on_disk, 'tools/starman.psgi';
 
+#TODO: Implement Template plugins <?lsmb USE plugin ?>
 ok_dependencies($file, \@on_disk,
                 phases => 'runtime',
                 ignores => [ 'Image::Size', 'LaTeX::Driver',
+                             'Template::Plugin::JSON',
```
ehuelsmann: If Template::Plugin::JSON isn't used anywhere, it would be good to document in `cpanfile` why it's required. (Is it still?)

sql/changes/1.6/country_date_format.sql
ehuelsmann: I still don't like this file. It'll be subject to aging and bitrot. And what's more: we can't change it after the fact, because that changes the sha sum.

sql/upgrade/sl2.8-1.5.sql
```diff
@@ -724,6 +724,7 @@ SELECT setval('id', max(id)) FROM transactions;
  SELECT setval('cr_report_line_id_seq', max(id)) FROM cr_report_line;
  SELECT setval('business_unit_id_seq', max(id)) FROM business_unit;
 
+INSERT INTO defaults(setting_key,value) VALUES('default_country', :default_country);
```
ehuelsmann: Isn't this unrelated to the change at hand?

sql/upgrade/sl3.0-1.5.sql
```diff
@@ -1006,6 +1006,7 @@ SELECT setval('business_unit_id_seq', max(id)) FROM business_unit;
 --    SELECT MAX(CAST(???number AS NUMERIC))+1 FROM SL30.??? WHERE ???number ~ '^[0-9]+$'
 --) WHERE setting_key = 'paynumber';
 
+INSERT INTO defaults(setting_key,value) VALUES('default_country', :default_country);
```
ehuelsmann: Isn't this unrelated to the change at hand?

sql/upgrade/sl3.0-1.6.sql
```diff
@@ -1398,6 +1398,7 @@ SELECT setval('business_unit_id_seq', max(id)) FROM business_unit;
 --    SELECT MAX(CAST(???number AS NUMERIC))+1 FROM SL30.??? WHERE ???number ~ '^[0-9]+$'
 --) WHERE setting_key = 'paynumber';
 
+INSERT INTO defaults(setting_key,value) VALUES('default_country', :default_country);
```
ehuelsmann: Unrelated?

xt/01.2-deps.t
```diff
@@ -34,8 +34,10 @@ find(\&collect, 'lib/', 'old/bin/', 'old/lib/');
 
 push @on_disk, 'tools/starman.psgi';
 
+#TODO: Implement Template plugins <?lsmb USE plugin ?>
```
ehuelsmann: Why are you submitting a PR that's got TODOs in them? Is it half-done?

xt/lib/PageObject/Setup/CreateUser.pm
```diff
@@ -20,7 +20,7 @@ __PACKAGE__->self_register(
 
 my @fields = ("Username", "Password", "Create new user", "Import existing user",
               "Salutation", "First Name", "Last name", "Employee Number",
-              "Date of Birth", "Tax ID/SSN", "Country", "Assign Permissions");
+              "Country", "Date of Birth", "Tax ID/SSN", "Assign Permissions");
```
ehuelsmann: Ideally, I'd like to see a test which implements filling out the country first and then the DOB *and* a test which fills out the DOB first and then the country. The inverted order shouldn't break anything.

sql/changes/1.6/country_date_format.sql
```diff
@@ -0,0 +1,281 @@
...
+INSERT INTO country(short_name,name) VALUES ('GG','Guernsey');
+INSERT INTO country(short_name,name) VALUES ('IC','Canary Islands');
+INSERT INTO country(short_name,name) VALUES ('ME','Montenegro');
+INSERT INTO country(short_name,name) VALUES ('MF','St. Martin');
+INSERT INTO country(short_name,name) VALUES ('SS','South Sudan');
+INSERT INTO country(short_name,name) VALUES ('SX','Sint Maarten');
+INSERT INTO country(short_name,name) VALUES ('TL','Timor-Leste');
+INSERT INTO country(short_name,name) VALUES ('XK','Kosovo');
+
+--UPDATE country SET date_format = '' WHERE short_name = 'AC'; -- Ascension Island
```
ehuelsmann: Why do we have all these commented-out lines in this file?

UI/js-src/lsmb/PublishSelect.js
ylavoie: Dojo decided that a transition from undefined to defined should not be published. Here we have dependent fields which clearly need to stay synced. It must be in startup because we rely on postCreate to initialize fields, so we have to wait for them to be properly setup before triggering an update.

ehuelsmann: Actually, Dojo's API documentation explicitly states that the locale value can *only* be set before Dojo-boot (at page load time). So, I think this approach is asking for trouble down the line: not a single widget will detect the change.

However, none of that explains why we need a "startup" function? The function is only called when the widget is instantiated, after all...

xt/01.2-deps.t
ylavoie: It is used by new_user.html but we do not currently recognize Template plugins in the dependencies.

xt/01.2-deps.t
ylavoie: Our dependency module never detected Template plugins. I put the reference manually in the list and added the TODO comment to upgrade the module. I am just trying to stay focused on the initial modification, not more.

UI/js-src/lsmb/PublishSelect.js
ylavoie: I do not change locale. I display a value dependent on another. I a country is selected, then the date placeholder has to change accordingly.

sql/upgrade/sl2.8-1.5.sql
ylavoie: On the contrary, we need to stash this value if we want to be able to use it as default selection for new users, based on the hypothesis that 99% of the time, the users will be in the same country as the company. So it can serve as a default for the initial user creation.

sql/upgrade/sl3.0-1.5.sql
ylavoie: Nope, it is.

sql/upgrade/sl3.0-1.6.sql
ylavoie: Nope

UI/js-src/lsmb/DateTextBox.js
ylavoie: Parent constructors are called automatically and we must not inherit within a constructor. Dojo reports an error with child classes.

UI/setup/select_coa.html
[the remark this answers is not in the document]
ehuelsmann: When people strip that information, they can't but expect to be served the default...

UI/setup/select_coa.html
```diff
@@ -24,11 +24,19 @@ <h2><?lsmb text('Database Management Console') ?></h2>
            label = text('Chart of accounts')
     };
 ELSE;
+    lang = '';
+    IF languages;
+        lang = languages.pop;
+        IF lang.search('-');
```
ehuelsmann: What's this code doing?? What problem is this solving? 'en-US' and 'en' are both in our language set, no? If not, why not? And why would this then be the best solution to that problem?

UI/setup/template_info.html
```diff
@@ -8,6 +8,11 @@ <h2><?lsmb text('Database Management Console') ?></h2>
 <form data-dojo-type="lsmb/SimpleForm" action="setup.pl" method="post">
 <div class="listtop"><?lsmb text('Select Templates to Load') ?></div>
 <?lsmb
+INCLUDE input element_data = {
```
ehuelsmann: Why do you want to retain the country of the CoA? The CoA has already been loaded...

PR #2753, opened 2017-04-22
Enable Perl::Critic RequireCheckedOpen Policy

nick-prater:


ehuelsmann:
@nick-prater agreed with @sbts ; maybe adding the OS error (`$!`) helps to clarify/debug?

lib/LedgerSMB/File.pm
```diff
@@ -247,9 +247,12 @@ sub get_for_template{
...
-                      args => [$args->{ref_key}, $args->{file_class}]
-     );
+        funcname => 'file__get_for_template',
+        args => [
+            $args->{ref_key},
+            $args->{file_class},
+        ],
+    );
     if ( -d $LedgerSMB::Sysconfig::tempdir . '/' . $$){
         die 'directory exists';
```
sbts: perhaps we can expand this error to include the dirname that exists

nick-prater: I'd rather refactor this and the associated cleanup to use File::Temp, but for now will expand the error.

lib/LedgerSMB/Template/DB.pm
```diff
@@ -139,7 +139,8 @@ sub get_from_file {
     $fname = $2;
     my ($template_name, $format) = split /\./, $fname;
     my $content = '';
-    open my $fh, '<', $path;
+    open my $fh, '<', $path
+        or die "Failed to open template path";
```
sbts: can we include the file name ($path) in the die?

PR #2754, opened 2017-04-23
Master feature skip lsmbisrunning on demand

sbts:
add skip early capability to Is_LSMB_running.sh

ensures the test is run in xt/60 regardless,
but allows it to be skipped on a per reop or per owner basis during travis setup stage


.travis.yml
```diff
@@ -129,7 +129,7 @@ before_script:
   - #PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES" plackup -Ilib --port 5001 --server HTTP::Server::PSGI tools/starman.psgi >/dev/null 2>/dev/null &
   - PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES" plackup -Ilib --port 5001 --server HTTP::Server::PSGI tools/starman.psgi --access_log /tmp/plackup-access.log 2> /tmp/plackup-error.log &
   - sleep 5  # give starman a chance to start
```
ehuelsmann: Since this sleep remains in place, I really, really don't like the added complexity of the solution (albeit less invasive than Yves's proposal).
I really do not think we need to optimize for this 5s sleep. Instead, we need to look at optimizing recurring sleeps or doing other stuff during those recurring  sleeps (like in the web framework).

sbts: Ok, I'll look at how we may be able to dynamically wait for starman/plack

PR #2760, opened 2017-04-24
Enable critic ProhibitIndirectSyntax Policy

nick-prater:
Instead of

    my $o = new Form();

use

    my $o = Form->new();

See: https://metacpan.org/source/SZABGAB/CGI-Simple-1.115/lib/CGI/Simple.pm

> Perl needs to make a number of assumptions at compile time to disambiguate the first form, so it tends to be fragile and to produce hard-to-track-down bugs.

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -389,45 +389,47 @@ sub run_backup {
     my $backupfile;
     my $mimetype;
 
-    if ($request->{backup} eq 'roles'){
-       my @t = localtime(time);
-       $t[4]++;
-       $t[5] += 1900;
-       $t[3] = substr( "0$t[3]", -2 );
-       $t[4] = substr( "0$t[4]", -2 );
-       my $date = "$t[5]-$t[4]-$t[3]";
+    if ($request->{backup} eq 'roles') {
+        my @t = localtime(time);
+        $t[4]++;
+        $t[5] += 1900;
+        $t[3] = substr( "0$t[3]", -2 );
+        $t[4] = substr( "0$t[4]", -2 );
+        my $date = "$t[5]-$t[4]-$t[3]";
 
-       $backupfile = $database->backup_globals(
-                      tempdir => $LedgerSMB::Sysconfig::backupdir,
-                         file => "roles_${date}.sql"
-       );
-       $mimetype   = 'text/x-sql';
-    } elsif ($request->{backup} eq 'db'){
-       $backupfile = $database->backup;
-       $mimetype   = 'application/octet-stream';
-    } else {
+        $backupfile = $database->backup_globals(
+            tempdir => $LedgerSMB::Sysconfig::backupdir,
+            file => "roles_${date}.sql"
+        );
+        $mimetype = 'text/x-sql';
+    }
+    elsif ($request->{backup} eq 'db') {
+        $backupfile = $database->backup;
+        $mimetype   = 'application/octet-stream';
+    }
+    else {
         die $request->{_locale}->text('Invalid backup request');
     }
 
     $backupfile or
         die $request->{_locale}->text('Error creating backup file');
 
-    if ($request->{backup_type} eq 'email'){
+    if ($request->{backup_type} eq 'email') {
         # suppress warning of single usage of $LedgerSMB::Sysconfig::...
         no warnings 'once';
 
         my $csettings = $LedgerSMB::Company_Config::settings;
         my $mail = LedgerSMB::Mailer->new(
-            from          => $LedgerSMB::Sysconfig::backup_email_from,
-            to            => $request->{email},
-            subject       => "Email of Backup",
-            message       => 'The Backup is Attached',
+            from     => $LedgerSMB::Sysconfig::backup_email_from,
+            to       => $request->{email},
+            subject  => "Email of Backup",
+            message  => 'The Backup is Attached',
         );
         $mail->attach(
             mimetype => $mimetype,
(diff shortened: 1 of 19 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -389,45 +389,47 @@ sub run_backup {
...
+            file => "roles_${date}.sql"
+        );
+        $mimetype = 'text/x-sql';
+    }
+    elsif ($request->{backup} eq 'db') {
+        $backupfile = $database->backup;
+        $mimetype   = 'application/octet-stream';
+    }
+    else {
         die $request->{_locale}->text('Invalid backup request');
```
sbts: Not essential, but perhaps report `$request->{backup}` in this error message.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -389,45 +389,47 @@ sub run_backup {
...
     }
 
     $backupfile or
         die $request->{_locale}->text('Error creating backup file');
```
sbts: Would be preferable to report `$backupfile` in this error

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -389,45 +389,47 @@ sub run_backup {
...
 
-    if ($request->{backup_type} eq 'email'){
+    if ($request->{backup_type} eq 'email') {
         # suppress warning of single usage of $LedgerSMB::Sysconfig::...
         no warnings 'once';
```
sbts: Probably don't want a blank line after this to help group it with the statement it affects.
Ideally wrap the 2 statements in something like....
```plain
# ##########
# suppress warning .......
no warnings 'once';
assignment to suppress warnings for;
# ##########
```

@ehuelsmann However, I wonder why we need to suppress the warnings here.
If we really do need to suppress them, then I'd like to see a comment describing the reason.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -389,45 +389,47 @@ sub run_backup {
...
-        my $mail = new LedgerSMB::Mailer(
-            from          => $LedgerSMB::Sysconfig::backup_email_from,
-            to            => $request->{email},
-            subject       => "Email of Backup",
-            message       => 'The Backup is Attached',
-            );
+        my $mail = LedgerSMB::Mailer->new(
+            from     => $LedgerSMB::Sysconfig::backup_email_from,
+            to       => $request->{email},
+            subject  => "Email of Backup",
```
sbts: can we insert a human readable date into the email subject?
eg:
> Email of Backup for 2nd March 2017

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -389,45 +389,47 @@ sub run_backup {
...
+            message  => 'The Backup is Attached',
+        );
         $mail->attach(
             mimetype => $mimetype,
             filename => 'ledgersmb-backup.sqlc',
```
sbts: It would be good to datestamp the filename here in yymmdd format, so something like....
`ledgersmb-backup-20170424.sqlc

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -436,19 +438,25 @@ sub run_backup {
...
         return $template->render_to_psgi($request);
-    } elsif ($request->{backup_type} eq 'browser') {
+    }
+    elsif ($request->{backup_type} eq 'browser') {
         my $bak;
         open $bak, '<', $backupfile
             or die "Failed to open temporary backup file $backupfile : $!";
         unlink $backupfile; # remove the file after it gets closed
 
         my $attachment_name = 'ledgersmb-backup-' . time . '.sqlc';
```
sbts: Not certain what format `time` is here, but it should match the timestamp used in the filename above.
Probably `yymmdd-hhMMss`

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -436,19 +438,25 @@ sub run_backup {
...
-        return [ 200,
-                 [ 'Content-Type' => $mimetype,
-                   'Content-Disposition' =>
-                        "attachment; filename=\"$attachment_name\"" ],
-                 $bak ];  # return the file-handle
-     } else {
+        return [
+            200,
```
sbts: coding style question, @ehuelsmann I feel `return [ 200,` is more readable than putting the `200` on the second line, But I'm happy to go with your preference

sbts:
I've added some comments, some of them are for possible changes that are slightly out of scope for this PR, but probably bring benefits if we implement them while touching these sections of the code.
Ideally addressing the different types of changes in grouped commits

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: @sbts the `no warnings 'once'` applies to the entire block. The reason we currently need that is because the file refers to that variable in a different namespace exactly once. This is usually the case for Sysconfig:: variables.
The assumption behind the 'once' warnings is that using a variable once is probably a typo.

Now, with the Sysconfig refactoring, where now everything is hidden behind a functional interface, I'm thinking we can probably replace this with a function call instead.

That's outside the scope of this PR, IMO.

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: That's an idea (even better in case of saving to disk, because the e-mail will be dated, but on disk, that's harder to trace back).

Outside the scope of this PR, IMO.

PR #2780, opened 2017-05-02
Remove Form module unused method arguments

nick-prater:
A number of methods in Form.pm have unused arguments. This patch removes them and updates their callers to them to simplify code.

### DIFF
--- a/old/lib/LedgerSMB/Form.pm
+++ b/old/lib/LedgerSMB/Form.pm
@@ -1623,7 +1623,7 @@ sub save_exchangerate {
     );
 }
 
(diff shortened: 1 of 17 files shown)


old/lib/LedgerSMB/Form.pm
```diff
@@ -1685,20 +1685,21 @@ sub check_exchangerate {
     $exchangerate;
 }
 
-=item $form->add_shipto($dbh, $id);
+=item $form->add_shipto($id, $is_oe);
```
ehuelsmann: Because the `$is_oe` parameter is optional, I'd propose we document it as:

```perl
=item $form->add_shipto($id [, $is_oe]);
```

PR #2786, opened 2017-05-05
UI Speed improvements

ylavoie:
Improve page loading by using a much quicker, but limited, html escaping module and move detection of available file format in configuration instead of page loading.
This commit reduces serving the login page on a fresh plackup by ~66%v

### DIFF
--- a/lib/LedgerSMB/Sysconfig.pm
+++ b/lib/LedgerSMB/Sysconfig.pm
@@ -207,7 +207,6 @@ def 'cache_templates',
     default => 0,
     doc => qq||;
 
-
 ### SECTION  ---   paths
 
 def 'pathsep',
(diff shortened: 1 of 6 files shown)


lib/LedgerSMB/Sysconfig.pm
```diff
@@ -504,6 +520,27 @@ sub check_permissions {
...
+###TODO-LOCALIZE-DOLLAR-AT
+our $latex = 0;
+
+
+sub override_defaults {
+    # Check Latex
+    $latex = eval {require Template::Plugin::Latex; 1;};
+
+    # Check availability and loadability
+    $LedgerSMB::Sysconfig::template_latex = eval {require LedgerSMB::Template::LaTeX; 1}
```
sbts: We should do a version check here and raise a sane error (on the console) if the installed version is too old.
Required version is listed in cpanfile but we can't obtain that directly as cpanfile is not guaranteed to be present (eg: system package install)

ylavoie: I think we should, but this is broader than the scope of this PR, as it implies moving cpanfile checks to runtime and should be done for all modules.

ehuelsmann:
Actually, there cannot be a case of "moving" here, because `cpanfile`
documents our dependencies and their versions, for the purpose of build
systems.
The version checks in the modules don't work for the build and installation
systems, but are there to detect installation issues at load time.
In other words, the version checks are additional.

lib/LedgerSMB/Sysconfig.pm
```diff
@@ -525,6 +520,27 @@ sub check_permissions {
     }
 }
 
+# if you have latex installed set to 1
+###TODO-LOCALIZE-DOLLAR-AT
+our $latex = 0;
+
+
+sub override_defaults {
+    # Check Latex
```
ehuelsmann: This line needs a `local $@` line before it to make sure the error value gets correctly restored.

ehuelsmann: Heh The following line*

PR #2799, opened 2017-05-13
Dojo menus

ylavoie:
Start using Dojo for menus


lib/LedgerSMB/DBObject/Menu.pm
```diff
@@ -65,6 +66,22 @@ sub generate_section {
     return @{$self->{menu_items}};
 }
 
+=item generate_item()
```
ehuelsmann: My thoughts were to generate the menu in a single request/response combination. If we use that approach, this function (and its associated SQL implementation below) are superfluous and probably better removed.

UI/menu/expanding.css
```diff
@@ -8,3 +8,42 @@ li.menu_closed > ul {
 li.menu_open > ul {
 	display: block;
 }
+
```
ehuelsmann: Since all CSS is integrated into `UI/css/ledgersmb.css` and `UI/css/system/global.css`, I don't think that changing this file matters; more specifically, it probably needs to be removed?

sql/modules/Menu.sql
```diff
@@ -12,13 +12,14 @@ CREATE TYPE menu_item AS (
    label varchar,
    path varchar,
    parent int,
-   args text[]
+   args text[],
```
ehuelsmann: Changing a type doesn't work on existing databases. You'll need to `DROP CASCADE` it first (`IF EXISTS`, of course because otherwise it doesn't work on new databases).

UI/js-src/lsmb/menus/MenuSelector.js
```diff
@@ -0,0 +1,26 @@
+define([
```
ehuelsmann: Is this file used anywhere?

UI/js-src/lsmb/menus/MenuRestStore.js
```diff
@@ -0,0 +1,119 @@
+define(["dojo/_base/declare",
+    "dojo/store/JsonRest", "dojo/store/Observable",
+    "dojo/store/Memory", "lsmb/menus/Cache",
+    "dijit/Tree", "dijit/tree/ObjectStoreModel",
+    "dijit/Menu", "dijit/MenuSeparator", "dijit/PopupMenuItem",
+    "dojo/when", "dojo/dom", "dojo/ready"
+], function(declare, JsonRest, Observable,
+    Memory, lsmbStoreCache,
```
ehuelsmann: the `lsmbStoreCache` variable doesn't follow the naming of the other variables. That's confusing.

UI/js-src/lsmb/menus/MenuRestStore.js
```diff
@@ -0,0 +1,119 @@
...
+        // to query the children of a node
+        var restStore = new JsonRest({
+            target:      "menu.pl?action=menuitems_json",
+            idProperty: "id",
+            _getTarget: function(id){
+                return this.target;
+            },
+        });
+        var memoryStore = new Memory({idProperty: "id"});
+        var store = new lsmbStoreCache(restStore, memoryStore);
```
ehuelsmann: Why can't we use the same model here as we used for the accounts and parts: simply initialize a MemoryStore from the JSON response from the server?
That eliminates the need to have a caching store.

UI/js-src/lsmb/menus/Cache.js
```diff
@@ -0,0 +1,65 @@
+define(["dojo/_base/lang","dojo/when" /*=====, "dojo/_base/declare", "dojo/store/api/Store" =====*/],
+function(lang, when /*=====, declare, Store =====*/){
+
+// module:
+//		lsmb/menus/Cache
+
+var Cache = function(masterStore, cachingStore, options){
```
ehuelsmann: Why do we need our own Cache store? Isn't the Dojo cache store good enough? If not, please include documentation why not.

UI/js-src/lsmb/menus/Cache.js
```diff
@@ -0,0 +1,65 @@
...
+                            cachingStore.put(child, {id: child.id});
+                        });
+                        result = cachingStore.get(id);
+                    }
+                    return result;
+                });
+            });
+        },
+/*
+        add: function(object, directives){
```
ehuelsmann: If this commented out code needs to stick around, please document why. If not, please remove it.

lib/LedgerSMB/DBObject/Menu.pm
ylavoie: I do, but I allow pulling only one item if necessary.

UI/menu/expanding.css
ylavoie: No, main.html explicitly requests this css for menu. We may decide to regroup it later.

UI/js-src/lsmb/menus/MenuSelector.js
ylavoie: No, remant from account. Removed

UI/js-src/lsmb/menus/MenuRestStore.js
ylavoie: Renamed.

UI/js-src/lsmb/menus/Cache.js
ylavoie: Because the actual Dojo cache doesn't return a promise, forcing the user to manage the cache externaly. This one returns a promise, so sync or async are transparents.

UI/js-src/lsmb/menus/Cache.js
ylavoie: Some will be uncommented in the Preferred menu PR, which is immediately following. The remainder will get removed there.

sql/changes/1.6/menu_item.sql
```diff
@@ -0,0 +1,11 @@
+DROP TYPE IF EXISTS menu_item CASCADE;
+CREATE TYPE menu_item AS (
```
ehuelsmann: I'm not sure I understand this change: we were returning a flat list where each collection element refers to its parent by `id`. What - on the server side - is going to populate the `children`? And if we return an array of children, then from the server's perspective, things are no longer flat, because you need recursive/additional queries to determine which are the children...

lib/LedgerSMB/DBObject/Menu.pm
ehuelsmann: Ok. Are there any compelling scenarios to add this extra bit of code? Is this an optimization? If so, is it an optimization of response time, network data, developer time?

Given the amount of data to be sent over (218 items max), I'm thinking we'd be wise to optimize for developer time here and add API entry points only when we need them. My point: is this additional code currently being tested with unit tests? (If not, removing the code is quicker than implementing unit tests, I imagine?)

sql/modules/Menu.sql
ylavoie: Fixed

UI/js-src/lsmb/menus/Cache.js
ylavoie: Fixed

sql/changes/1.6/menu_item.sql
ylavoie: The same query also returns the children to prevent recomputing in the UI. We need to go through them in SQL anyway.

ylavoie:
Too many errors. Rebasing PR into a more understandable version. Closing.

PR #2801, opened 2017-05-14
1.5 Fix #2800 can't delete unused account after books closed

### LINKED ISSUE #2800: Can't Delete unused account after Books Closed
PR to fix this will follow shortly...

Looking at Chart Of Accounts:
 * Accounts that have never been used have a Delete button which should remove them.
 * But once books have been closed, such as at year-end, Clicking Delete gives an error:
  `ERROR: update or delete on table "account" violates foreign key constraint "account_checkpoint_account_id_fkey" on table "account_checkpoint"`

Some discussion here:
https://matrix.to/#/!qyoLumPqusaXqFJNyK:matrix.org/$1494605749314599xQmTV:matrix.org

nick-prater:
Ensures that associated zero-value records in account_checkpoint are deleted before the account, to avoid foreign-key errors.


sql/changes/LOADORDER
```diff
@@ -44,4 +44,5 @@
 1.5/issue-2380-modify-tax_extended_entry_id-constraint.sql
 1.5/issue-2278-modify-acc_trans-index.sql
 1.5/template_menu3.sql
+1.5/issue-2800-cant-delete-unused-account-after-books-closed.sql
```
ehuelsmann: No need to create an `sql/changes/` item for this one: modules are always reloaded.
`sql/changes/` only comes in when changing the schema definition (adding/removing indexes, adding/changing/removing columns and tables, etc).

PR #2805, opened 2017-05-14
Enable Perl::Critic tests on files in lib/LedgerSMB/Auth/

nick-prater:
For some historic reason, files in lib/LedgerSMB/Auth/ were excluded
from tests. Only one change needed for them to pass the tests -
adding `use namespace::autoclean`.


ehuelsmann:
yup. Didn't use to be true modules (now they are, since a week).

PR #2806, opened 2017-05-15
Remove general Module::Runtime dependency

nick-prater:
After discussion (https://matrix.to/#/!qyoLumPqusaXqFJNyK:matrix.org/$14947921991651761iyRKA:matrix.org) using Perl core module Module::Load rather than non-core Module::Runtime to dynamically load modules.

Cannot completely remove dependency on Module::Runtime as it is used for X12 EDI support. Already the cpanfile depends on Module::Runtime only if the edi feature is enabled, so doesn't need to change.

### DIFF
--- a/lib/LedgerSMB/Auth.pm
+++ b/lib/LedgerSMB/Auth.pm
@@ -54,16 +54,15 @@ use strict;
 use warnings;
 
 use LedgerSMB::Sysconfig;
-use Module::Runtime qw(use_module);
-
+use Module::Load;
 
 my $plugin = 'LedgerSMB::Auth::' . LedgerSMB::Sysconfig::auth;
-use_module($plugin) or die "Can't locate Auth parser plugin $plugin";
+load $plugin;
 
 sub factory {
     my ($psgi_env) = @_;
 
-    return use_module($plugin)->new(env => $psgi_env);
+    return $plugin->new(env => $psgi_env);
 }
 
 
--- a/lib/LedgerSMB/PSGI.pm
+++ b/lib/LedgerSMB/PSGI.pm
@@ -17,9 +17,8 @@ use warnings;
 use LedgerSMB;
 use LedgerSMB::App_State;
 use LedgerSMB::Auth;
-
+use Module::Load;
 use CGI::Emulate::PSGI;
-use Module::Runtime qw/ use_module /;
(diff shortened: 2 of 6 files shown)


ehuelsmann:
While trying to fix the issues that make this PR fail, I found: https://rt.cpan.org/Ticket/Display.html?id=106128 

As we're dealing with a web server environment, we probably need to carefully consider the arguments presented there.

Also, see my current intermediate result: https://travis-ci.org/ehuelsmann/LedgerSMB/builds/232580120
It seems like the variant of this module distributed with 5.18 and earlier (module version 0.18 and earlier) are broken.

ehuelsmann:
We (@nick_prater and I) agreed that instead of removing Module::Runtime dependency, we'll switch completely *to* it, away from Module::Load due to security concerns.

sbts:
@ehuelsmann 
I've just had a look at both the the links you posted above and the diff for the PR.

In your last comment you say....
> away from Module::Load due to security concerns.

However the diff shows a complete switch __to__ Module::Load

Also reading the CWE links etc I'd have to say this PR, as it is, is incomplete.
It doesn't validate the value passed to ::Load in any way.

I'd think we should use realpath() and then ensure the resultant path is at least somewhere within our currently set module load path.
ie: system module, local::lib module, our tree module.

If, as with
```perl
my $plugin = 'LedgerSMB::Auth::' . LedgerSMB::Sysconfig::auth;
```
We are going to pass an actual module name rather than a path, we should validate that the resultant module name conforms to module naming conventions and can't be pathlike.

I'd go as far as to suggest we should wrap Module::Load with our own function that does this and then use critic to blacklist Module::Load outside that function.
Or at least have a validation function and make it mandatory for that function to be used on the line immediately prior to the use of Load and have a test to ensure that is the case.

sbts:
I'll re-open pending your clarification and possible immediate reclosure.

ehuelsmann:
I closed this PR because it moves *to* Module:: Load. After fixing this PR
and reading cpan rt tickets, @nick-prater and I discussed the PR and my
fixes as well as the security concerns raised in the CPAN rt. W decided
this was moving the wrong direction: we need to move *to* Module:: Runtime
instead.

ehuelsmann:
Closing again, dis ussed with @sbts

PR #2808, opened 2017-05-15
Fr1  -- RequireFinalReturn

ziuq:
Silencing the require final return criticism.  I was a believer in this critique before, now I believe more deeply.

Added a lot of  returns.   Removed one or two function prototypes and a few trailing spaces.

In 9e69fc32, removed  duplicated bulk from the branches of a condition.

This has passed local tests and TravisCI.

Rob

### DIFF
--- a/lib/LedgerSMB/File.pm
+++ b/lib/LedgerSMB/File.pm
@@ -220,7 +220,8 @@ Auto-detects the type of the file.  Not yet implemented
 
 sub detect_type {
     my ($self) = @_;
-    return $logger->warn("Stub LedgerSMB::File::detect_type\n");
+    $logger->warn("Stub LedgerSMB::File::detect_type\n");
+    return;
 };
 
 =item get
@@ -307,7 +308,8 @@ sub DEMOLISH {
        unlink $self->{file_path} . '/' . $file;
    }
    closedir (TMP);
-   return rmdir $self->{file_path};
+   rmdir $self->{file_path};
+   return;
 }
 
 =item list({ref_key => int, file_class => int})
--- a/lib/LedgerSMB/File.pm
+++ b/lib/LedgerSMB/File.pm
@@ -208,7 +208,7 @@ sub set_mime_type {
     $self->mime_type_text($mime_type);
     my ($ref) = $self->call_procedure(funcname => 'file__mime_type_text',
          args => [undef, $self->mime_type_text]);
-    $self->mime_type_id($ref->{id});
+    return $self->mime_type_id($ref->{id});
 
 }
 
@@ -220,7 +220,7 @@ Auto-detects the type of the file.  Not yet implemented
 
 sub detect_type {
     my ($self) = @_;
-    $logger->warn("Stub LedgerSMB::File::detect_type\n");
+    return $logger->warn("Stub LedgerSMB::File::detect_type\n");
 };
 
 =item get
@@ -233,6 +233,7 @@ sub get {
     my ($self) = @_;
     my ($ref) = $self->call_dbmethod(funcname => 'file__get');
     $self->{$_} = $ref->{$_} for keys %$ref;
+    return;
 }
 
 =item get_for_template({ref_key => int, file_class => int})
@@ -306,7 +307,7 @@ sub DEMOLISH {
        unlink $self->{file_path} . '/' . $file;
    }
    closedir (TMP);
-   rmdir $self->{file_path};
+   return rmdir $self->{file_path};
 }
 
 =item list({ref_key => int, file_class => int})
--- a/lib/LedgerSMB/Report/Dates.pm
+++ b/lib/LedgerSMB/Report/Dates.pm
@@ -216,7 +216,7 @@ sub _get_to_date {
     if ($self->interval eq 'month'){
        $date->add(months => 1);
     } elsif ($self->interval eq 'quarter'){
-       $date->add(months => 3);
+       $date->add(months => 3);         ## no critic ( ProhibitMagicNumbers)
     } elsif ($self->interval eq 'year'){
        $date->add(years => 1);
     }
@@ -232,6 +232,7 @@ sub _set_lazy_dates {
               $self->to_date;
               $self->date_from;
               $self->date_to;
+              return;
 }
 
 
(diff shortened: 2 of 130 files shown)


lib/LedgerSMB/File.pm
```diff
@@ -306,7 +307,7 @@ sub DEMOLISH {
        unlink $self->{file_path} . '/' . $file;
    }
    closedir (TMP);
-   rmdir $self->{file_path};
+   return rmdir $self->{file_path};
```
ehuelsmann: this probably didn't mean to return the rmdir value, but 'undef' instead.

lib/LedgerSMB/Report/Dates.pm
```diff
@@ -216,7 +216,7 @@ sub _get_to_date {
     if ($self->interval eq 'month'){
        $date->add(months => 1);
     } elsif ($self->interval eq 'quarter'){
-       $date->add(months => 3);
+       $date->add(months => 3);         ## no critic ( ProhibitMagicNumbers)
```
ehuelsmann: Hmm. Where did the ProhibitMagicNumbers come from?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -248,15 +248,15 @@ sub login {
...
+    return $template->render_to_psgi($request);
 }
 
 =item sanity_checks
 Checks for common setup issues and errors if admin tasks cannot be completed/
 
 =cut
 
-sub sanity_checks {
+sub sanity_checks { ## no critic ( RequireFinalReturn )
```
ehuelsmann: I can see that we want to return 'undef' (or a simple  final `return;`) here: this seems to be more of a `procedure` (Pascal terms: not meant to have a return value) than a `function` (explicit return value)

lib/LedgerSMB/Template.pm
```diff
@@ -619,7 +620,7 @@ sub _http_output {
     print $data;
     # change global resource back asap
     binmode (STDOUT, ':utf8');
-    $logger->trace("end print to STDOUT");
+    return $logger->trace("end print to STDOUT");
```
ehuelsmann: There's an empty `return;` meant to be closing this sub.

lib/LedgerSMB/Template.pm
```diff
@@ -693,7 +696,7 @@ sub _email_output {
             @attachment,
         );
     }
-    $mail->send;
+    return $mail->send;
```
ehuelsmann: There's an empty `return;` meant to be here, I think.

lib/LedgerSMB/Template.pm
```diff
@@ -718,7 +721,7 @@ sub _lpr_output {
     }
 
     close $pipe;
-    close $file;
+    return close $file;
```
ehuelsmann: Same empty `return;` here too, I think.

lib/LedgerSMB/Template/ODS.pm
```diff
@@ -147,11 +147,12 @@ sub _worksheet_handler {
     } else {
         $sheet = $ods->appendTable($_->{att}->{name}, $rows, $columns);
     }
+    return;
 }
 
 sub _row_handler {
     $rowcount++;
-    $currcol = 0;
+    return $currcol = 0;
```
ehuelsmann: This callback and the ones below are probably expected to return undef: `return;`

lib/LedgerSMB/Template/XLS.pm
```diff
@@ -67,25 +67,25 @@ my $currcol;
 sub _worksheet_handler {
     $rowcount = -1;
     $currcol = 0;
-    $_->set_att(type => 'worksheet');
+    return $_->set_att(type => 'worksheet');
```
ehuelsmann: These handlers probably should be returning undef (`return;`)

lib/LedgerSMB/Template/XLSX.pm
```diff
@@ -67,25 +67,25 @@ my $currcol;
 sub _worksheet_handler {
     $rowcount = -1;
     $currcol = 0;
-    $_->set_att(type => 'worksheet');
+    return $_->set_att(type => 'worksheet');
```
ehuelsmann: These handlers probably should be returning undef (`return;`)

lib/LedgerSMB/File.pm
ziuq: In the face of 25% test coverage, usually no doc of the return values, and no response on lsmb.dev, I mostly preserved the actual logic of the code, rather than go with my guess of the author's intention.  20% bad guesses * 20% bad luck on those * 25% covered leaves a couple of unnoticed bugs alive in code.  

I varied from conservative to very conservative on risking logic changes.    I did change loop-condition-returns and where an ending conditional gave very disparate final values.

So I agree with all your 'probably meant "return;"', but used a more conservative standard.
I have no problem with redoing any or all of these on your say so.

ehuelsmann: Please don't take my review as something negative. It's definitely not something ad-hominem: the feedback I'm providing merely serves to improve the change we'll be merging to the master branch. That might include changes which - while staring us in the face due to this change - should be fixed one time or another. It's more practical to do that right now with the PR pending, because now they can't be forgotten.

In short: yes, I'd like you to make the changes.

lib/LedgerSMB/Report/Dates.pm
ziuq: Sorry about that.  I meant to mention it in the comment above.  At one point, I wanted something to push to Github this was it. 

Do you think that in context this is not very magical and so no-critic works?  MONTHS_PER_QUARTER is more correct.  Which is preferred,  Readonly or constant or
something else?

lib/LedgerSMB/File.pm
ziuq: No offense taken.  I was asking on lsmb dev for guidance on this, I will certainly take it now.

All your mentioned changes are up for review, plus a few cousins that show a strong family resemblance..

ehuelsmann:
Thanks for the contribution! Merging.

PR #2815, opened 2017-05-19
Removing warnings with PGObject2

einhverfr:
This is a minimal shim to avoid warnings (and call the correct APIs) for PGObject 2 and type registration.

The major incompatibilities in PGObject 2 do *not* affect LedgerSMB and appropriate interfaces have been provided to allow it to run.  However since these will be removed in the future at some point they throw warnings when called (at startup).  This results in several pages warnings.  No additional warnings are shows even without this patch while operating the software however, at least as far as I can see.

One extremely important thing to note here is that the first version of PGObject released was 1.0, so the only versions we have to worry about are those which start with a 1.  This logic has been ported so far into all of the type modules and all have been released except BigFloat (where PGObject has broken something in 1.x but this has been fixed in 2.x pending release and Bytestream (where I am waiting for author review).

### DIFF
--- a/lib/LedgerSMB/PGDate.pm
+++ b/lib/LedgerSMB/PGDate.pm
@@ -13,9 +13,17 @@ use base qw(PGObject::Type::DateTime);
 use strict;
 use warnings;
 
-PGObject->register_type(pg_type => $_,
+if ($PGObject::VERSION =~ /^1\./){
+    PGObject->register_type(pg_type => $_,
                                   perl_class => __PACKAGE__)
-   for ('date');
+    for ('date');
+} else {
+    PGObject::Type::Registry->register_type(
+                                  registry => 'default',
+                                    dbtype => $_,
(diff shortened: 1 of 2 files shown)


einhverfr:
At this point it is important that this branch be held briefly while we discuss.  I am putting on hold until later.  However this represents the only changes needed to support PGObject 2 fully.  Everything else is PGObject's responsibility.

einhverfr:
ok I am happy with this at any time it is to be merged in.

einhverfr:
builds passing with pgobject2 as well now.

einhverfr:
The reason to take this approach and a better one in master is because I am not 100% sure of the history of the register->() function associated with the type handlers.  It's possible (though not certain) that these may have problems in old versions with sub classes.    However, these would be old versions (significantly older than 1.5) so I am not as worried about 1.6

PR #2817, opened 2017-05-20
Remove unused LedgerSMB::DBObject::Part

einhverfr:
No longer seems used.  Removing.


ehuelsmann:
@einhverfr could you send me instructions how to update the model.tex?

einhverfr:
@ehuelsmann I didn't remember so I just deleted the lines manually.  I had a script at one point that generated it.  I would be fine with deleting the file since it is just what comes from perldoc anyway.

PR #2819, opened 2017-05-20
Moving lsmb to type registry for 'ledgersmb' on pgobject

einhverfr:
Master changes.  I am not 100% sure about minimum PGObject::Type::* versions needed for this but anything within the last couple years ought to be recent enough.  May want to specify that at some point.


lib/LedgerSMB/PGNumber.pm
```diff
@@ -8,15 +8,14 @@ LedgerSMB::PGNumber - Number handling and serialization to database
...
+use parent qw(PGObject::Type::BigFloat);
 use strict;
 use warnings;
 use Number::Format;
 use LedgerSMB::Setting;
 
-PGObject->register_type(pg_type => $_,
-                                  perl_class => __PACKAGE__)
-   for ('float4', 'float8', 'double precision', 'float', 'numeric');
+LedgerSMB::PGNumber->register_type(registry => 'ledgersmb',
```
ehuelsmann: Why do you use `LedgerSMB::PGNumber` here and `__PACKAGE__` in PGDate?

lib/LedgerSMB/PSGI.pm
```diff
@@ -14,6 +14,10 @@ LedgerSMB::PSGI - PSGI application routines for LedgerSMB
 use strict;
 use warnings;
 
+use PGObject;
+
+BEGIN { PGObject->new_registry('ledgersmb') }
```
ehuelsmann: Would the new_registry not be a nice argument to the 'use PGObject' call? Seeing `BEGIN` blocks triggers questions :-)

einhverfr: that's a good idea.  Then we'd need to specify that as a minimum version, I suppose :-)

einhverfr: but for 1.6 I don't see a problem.  Will create a 2.1 today with this feature and then amend.

.travis.yml
```diff
@@ -137,7 +137,8 @@ before_script:
   - ((phantomjs --webdriver=4422 2>/dev/null >/dev/null &) && echo "PhantomJS started succesfully") || (echo "Failed to start PhantomJS!" ; exit 1)
   - #PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES" plackup -Ilib --port 5001 --server HTTP::Server::PSGI tools/starman.psgi >/dev/null 2>/dev/null &
   - PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES" plackup -Ilib --port 5001 --server HTTP::Server::PSGI tools/starman.psgi --access_log /tmp/plackup-access.log 2> /tmp/plackup-error.log &
-  - #sleep 5  # give starman a chance to start # we now have better handling for this in Is_LSMB_running.sh
+  - sleep 5  # give starman a chance to start # we now have better handling for this in Is_LSMB_running.sh
+  - cat /tmp/plackup-error.log
```
sbts: @einhverfr can you please drop in to #ledgersmb to discuss this.
The sleep should no longer be required due to changes in "Is_LSMB_Running.sh" 

However, the failure you saw in testing didn't dump logfiles as expected so we need to investigate that

einhverfr:
PGObject2 is out now.  So plan is to test first and make srue everything is green on my repo, then clear caches and re-test as a pr (and remove on-hold flag at that point).

PR #2824, opened 2017-05-21
Moving Setting.pm to old

einhverfr:
I looked at trying to rewrite this but we are using it everywhere and that is a larger job than I wanted to do quickly.  So moving it as a way to flag the need to rewrite it.  This is, unfortunately, a fairly central piece of logic so rewriting will have to be done carefully.


einhverfr:
How did coverage increase on a commit with no code?

ehuelsmann:
@einhverfr yes, that's problematic. Apparently there are timing issues with the way coverage is being measured. Sometimes line counts differ slightly, apparently missing a line every now and then.

PR #2827, opened 2017-05-22
Minimal removal of PGObject2 warnings

einhverfr:
Does nothing but moves to the correct api.


lib/LedgerSMB/PGNumber.pm
```diff
@@ -14,9 +14,8 @@ use warnings;
 use Number::Format;
 use LedgerSMB::Setting;
 
-PGObject->register_type(pg_type => $_,
-                                  perl_class => __PACKAGE__)
-   for ('float4', 'float8', 'double precision', 'float', 'numeric');
+__PACKAGE__->register(registry => 'default',
+    dbtype => [qw(float4 float8 float numeric), 'double precision']);
```
ehuelsmann: is the 'dbtype' PGObject1 compatible?

ehuelsmann: my point being: if it's not, shouldn't we set a lower limit on the PGObject supported versions?

einhverfr: It's actually a good catch, and it raises a different question.  dbtypes is the wrong keyword when going through the type module (types is used instead).  In the case of PGNumber we register more types than default.  I am not sure  what the correct behavior we want is for PGDate though, where we register fewer.

Without passing the types keyword, we register the default types which is why our tests didn't catch it.

einhverfr:
I have added min versions where they were missing.  Also bumped the required PGObject.pm version to 2.0 so we can more readily move to a registry shortly (but will not attempt that until thsi is merged).

PR #2616, opened 2017-02-09
Parallel tests

ylavoie:

PR #2828, opened 2017-05-22
Remove spurious testrules.yml from root

ylavoie:
Testrules.yml needs to be in t. Remove leftover from #2616


ehuelsmann:
Why does this PR contain dojo changes?

ylavoie:
This was submitted with Dojo 1.12.1

PR #2833, opened 2017-05-22
  Nm1

ziuq:
Problems with Pod coverage on Magic.pm on Perls 5.18, 5.16, and 5.14.   I am just not seeing it.

Reworked  negativising code to bool and unary minus vs. multiplying negative one.

Named many magicals.

I will be not be adding to this pull request.   The response to these changes will guide how I
do the rest.


lib/LedgerSMB/Template/XLS.pm
```diff
@@ -65,8 +65,9 @@ my $rowcount;
 my $currcol;
 
 sub _worksheet_handler {
-    $rowcount = -1;
     $currcol = 0;
+    $rowcount = 0;
+    -- $rowcount;
```
ehuelsmann: @ylavoie I think the code can be completely removed; the rowcount isn't used anywhere, is it?

lib/LedgerSMB/Template/XLSX.pm
```diff
@@ -65,7 +65,8 @@ my $rowcount;
 my $currcol;
 
 sub _worksheet_handler {
-    $rowcount = -1;
+    $rowcount = 0;
+    -- $rowcount;
```
ehuelsmann: @ylavoie I think the code can be completely removed; the rowcount isn't used anywhere, is it?

lib/LedgerSMB/Tax.pm
```diff
@@ -47,7 +47,7 @@ sub init_taxes {
     my @accounts = split / /, $taxaccounts;
     if ( defined $taxaccounts2 ) {
         #my @tmpaccounts = @accounts;#unused var
-        $#accounts = -1;# empty @accounts,@accounts=();
+        @accounts=(); # empty @accounts
```
sbts: @ehuelsmann can you double check the original code and new here.
I'm wondering if there may perhaps have been a good reason it was done the way it was (after all the comment show its equivalent to the new code so why didn't the code it's self get changed instead of adding the comment)

lib/LedgerSMB/Template/LaTeX.pm
```diff
@@ -57,13 +57,14 @@ use strict;
...
+use LedgerSMB::Magic qw( UNI_aring UNI_Aring );
 use Log::Log4perl;
 use TeX::Encode::charmap;
 use TeX::Encode;
 
 BEGIN {
-    delete $TeX::Encode::charmap::ACCENTED_CHARS{chr(0x00c5)};
-    delete $TeX::Encode::charmap::ACCENTED_CHARS{chr(0x00e5)};
+    delete $TeX::Encode::charmap::ACCENTED_CHARS{chr(UNI_Aring)};
+    delete $TeX::Encode::charmap::ACCENTED_CHARS{chr(UNI_aring)};
```
sbts: Hmm. Why are we deleting these specific chars?
At a minimum we need some comments as to why it's necessary.
@ehuelsmann and @eihnverfr can one of you cast some light on this?

lib/LedgerSMB/X12/EDI850.pm
```diff
@@ -66,25 +66,7 @@ sub _order {
...
-                   $elements[9],
-                   $elements[10],
-                   $elements[11],
-                   $elements[12],
-                   $elements[13],
-                   $elements[14],
-                   $elements[15],
-                   $elements[16],
-                );
+                my @new_elements = @elements[0 ..4, 7, 8, 5, 6, 9 .. 16 ];
```
sbts: To be honest I prefer the original form of this. 
But it does need some comments on each line explaining what each field actually is

old/lib/LedgerSMB/DBObject/Admin.pm
```diff
@@ -29,6 +29,7 @@ User/group management for LedgerSMB
 
 use base qw(LedgerSMB::PGOld);
 
+use LedgerSMB::Magic qw( EC_EMPLOYEE );
```
sbts: Why do we need to enable this constant if there is no change to use it?

old/lib/LedgerSMB/DBObject/Reconciliation.pm
```diff
@@ -401,8 +402,9 @@ sub get {
 
     for my $line (@{$self->{report_lines}}){
         if ($line->{cleared}){
-            $our_balance += ($neg * $line->{our_balance});
-            $self->{cleared_total} += ($neg * $line->{our_balance});
+            $our_balance += $pos ? $line->{our_balance} : -$line->{our_balance};
+            $self->{cleared_total} += $pos ? $line->{our_balance} : -$line->{our_balance};
```
sbts: Comments as before

old/lib/LedgerSMB/DBObject/Reconciliation.pm
```diff
@@ -377,10 +378,10 @@ sub get {
     my %report_days = map { $_->{id} => $_->{days} } @{$db_report_days};
     ($ref) = $self->call_dbmethod(funcname=>'account_get',
                                 args => { id => $self->{chart_id} });
-    my $neg = 1;
+    my $pos = 1;
     if (defined $self->{account_info}->{category}   # Report may be empty
     and $self->{account_info}->{category} =~ /(A|E)/){
-        $neg = -1;
+        $pos = 0;
```
sbts: As suggested previously we really should look at factoring this sort of behaviour put to a single function if possible

old/lib/LedgerSMB/DBObject/Reconciliation.pm
```diff
@@ -433,8 +435,8 @@ sub get {
     }
     $self->{format_amount} = sub { return $self->format_amount(@_); };
     if ($self->{account_info}->{category} =~ /(A|E)/){
-       $self->{our_total} *= -1;
-       return $self->{mismatch_their_total} *= -1;
+       $self->{our_total} = -$self->{our_total};
+       return $self->{mismatch_their_total} = -$self->{mismatch_their_total};
```
sbts: Comments as before

lib/LedgerSMB/Magic.pm
```diff
@@ -0,0 +1,388 @@
...
+    RC_IMPORT           => 3,
+    RC_PARTIAL_DISPOSAL => 4,
+
+    # codes
+
+    # EDI
+    EDI_PATHNAME_MAX => 180,    # default max length of EDI pathname
+                                # code says 180, perldoc says 179
+
+    # HTTP  These can be taken from HTTP::Status
```
ziuq: For six items I'd just leave them.  But I did not consider this as their final resting place; but just as
a place to put magic until they are all, or mostly, reaped.  My thought  was to sort it out after it is all
together.

lib/LedgerSMB/Entity/Note.pm
```diff
@@ -8,7 +8,7 @@ employees, etc.
   @notes = LedgerSMB::Entity::Bank->list($entity_id, [$credit_id]);
   $note->add;
 
-=head1 DESCRIPTION
+=head1 DESCRIPTION
```
ziuq: Really, I had not noticed.  Ha!

I have a perltidyrc that seems to fit the newer code formating rules.   Even if it is just what folk like, it 
seems like torture to pour a lot of reformats on top of all these refactorings.

lib/LedgerSMB/Magic.pm
```diff
@@ -0,0 +1,388 @@
...
+    HTTP_SEE_OTHER             => 303,
+    HTTP_BAD_REQUEST           => 400,
+    HTTP_UNAUTHORIZED          => 401,
+    HTTP_INTERNAL_SERVER_ERROR => 500,
+
+    HTTP_454 => 454,
+
+    # Unicode
+    UNI_Aring => 0x00c5,
+    UNI_aring => 0x00e5,
```
ziuq: I don't understand your point.   I did not explore why 0x00c5 and 0x00e5 character codes were in the codebase.

lib/LedgerSMB/Magic.pm
```diff
@@ -0,0 +1,388 @@
...
+=head3  Unicode
+
+    UNI_Aring               0x00c5
+    UNI_aring               0x00e5
+
+=head1 BUGS
+
+Are more organized.
+
+
```
ziuq: I expect to reorganize the pod when the rest of the constants are in it.

lib/LedgerSMB/Report/Listings/TemplateTrans.pm
```diff
@@ -110,18 +110,14 @@ sub name {
 
 =cut
 
-my %jtype = (
-    1 => 'gl',
-    2 => 'ar',
-    3 => 'ap',
-    );
+my @jtype = ( undef, 'gl', 'ar', 'ap' );    # XXX magic number map.  From where is the magic? --rir
```
ziuq: I agree.  `XXX` is a marker for severely damaged code. 

From whence is this magic?     Or if it is ad hoc,  what shall we call it?

lib/LedgerSMB/Budget.pm
```diff
@@ -246,7 +246,7 @@ sub from_input {
              $line->{amount} =   $input->{"credit_$rownum"}
                                - $input->{"debit_$rownum"};
              $line->{credit} = $line->{amount} if $line->{amount} > 0;
-             $line->{debit}  = $line->{amount} * -1 if $line->{amount} < 0;
+             $line->{debit}  = -$line->{amount} if $line->{amount} < 0;
```
ziuq: Sorry.  I viewed the task as removing magic and as just one task.  '-1' is considered magical and so it got tossed.

lib/LedgerSMB/Report/PNL.pm
```diff
@@ -243,11 +243,11 @@ sub run_report {
...
-            my $sign = (($contra) ? -1 : 1)
-                * ((($cat eq 'A') || ($cat eq 'E')) ? -1 : 1);
+            # XXX if $contra is constrained to 0, 1, undef, and '' then no bangbang
+            my $is_positive = !!$contra == ( $cat eq 'A' or $cat eq 'E' );
 
-            $self->cell_value($row_id, $col_id, $sign * $value)
-                if $sign < 0;
+            $self->cell_value($row_id, $col_id, -$value)
+                if not $is_positive;
         }
```
ziuq: It might be more difficult to isolate this into a function for general use than we would hope.

In this specific code, the signedness of `$value` depends on whether `cell_value` is expressing a credit or a debit.
The design choice to use contra was unhelpful in this.

lib/LedgerSMB/Report/GL.pm
```diff
@@ -333,7 +333,7 @@ sub run_report{
     my @rows = $self->call_dbmethod(funcname => 'report__gl');
     for my $ref(@rows){
         if ($ref->{amount} < 0){
-            $ref->{debits} = $ref->{amount} * -1;
+            $ref->{debits} =  -$ref->{amount};
```
ziuq: `-$ref->{amount}`  inverts the sign of `$ref->{amount}`. 

I think the direct unary minus alone is clearer.

My fonts are bigger than most peoples.   `0 - $stuff` also bears consideration.

old/lib/LedgerSMB/DBObject/Admin.pm
ziuq: I am too young to be in the old library.

lib/LedgerSMB/X12/EDI850.pm
ziuq: Better to have identifiers do the explaining.

lib/LedgerSMB/Magic.pm
ehuelsmann: Ok. It seems like we're already depending (although currently indirectly) on HTTP::Status. I'd be fine with replacing these magic numbers with the HTTP Status values directly instead of going through the process of centralizing them here first.

lib/LedgerSMB/Magic.pm
ehuelsmann: @sbts I think we might want to replace the original code where these character references appear with the `\N` syntax:

`\N{LATIN CAPITAL LETTER A WITH RING ABOVE}` (0x00c5)  
and  
`\N{LATIN SMALL LETTER A WITH RING ABOVE}` (0x00e5)

lib/LedgerSMB/Scripts/budgets.pm
```diff
@@ -179,7 +179,7 @@ sub view_budget {
         my $row = {};
         $row->{description} = $line->{description};
         if ($line->{amount} < 0 ) {
-            $row->{debit} = $line->{amount} * -1;
+            $row->{debit} = -$line->{amount};
```
ehuelsmann: Here, the intent of the code is to make the value positive. I'd say that the `abs()` (https://perldoc.perl.org/functions/abs.html) function should do exactly that.

ehuelsmann:
Looking at the review as a whole, my first reaction is to say that we need to allow `-1` as a non-magical value too. `Perl::Critic` has a facility to do that.

My second thought (unrelated to this PR in particular, more a thought in general) is that we have so much sign-changing going on all over the place, that we probably need better input parsing: something to parse a value into a debit/credit based on the type of transaction. Similar for amount printing: have a routine which determines whether or not an amount needs to change sign when presented as an amount relating to a specific account and/or specific transaction type.

lib/LedgerSMB/Budget.pm
ziuq: Well, I don't have problem with this pull request being refused and I splitting it up for you.  I don't know how that is done (the rejection).    In your place, I'd probably be lazy and want to deal with it just once.

 I do not yet have the view that a PR is the unit of change, rather I think of commits as the atomic tracking unit.  I have loads to learn about git.  So I don't understand how the bug tracking will be easier.  

It is disheartening to find that the critiques I that I took to satisfy are not generally accepted by the team.   I have been on the other side where I have felt "noone told me, I did not agree to that."  In that way, I'm just caught in the middle.

It is heartening that  these mostly mind numbingly boring changes, bring some scrutiny to crufty code; for instance: you picking up that a megabyte is the wrong size.

lib/LedgerSMB/Scripts/budgets.pm
ziuq: I think `abs` is a bad choice here.  It is correct, but the generalized credit/debit problem running through the code is reversing the sign when required.

lib/LedgerSMB/Report/GL.pm
sbts: Sorry my question about what the `-$ref` is negating was somewhat rhetorical.
It may not be a problem in Perl, but in many languages, the exact behavior of that construct depends on many things, and the general recommendation is that brackets be used to ensure the correct scoping.

lib/LedgerSMB/Report/Listings/TemplateTrans.pm
sbts: Hmm, Rather than xxx, please use "fixme: " or "todo: " with an explanation of what the issue is and what may need to be done to resolve it.
That way some of the automated tooling around can easily identify things that need to be addressed.
The use of "xxx" would simply just get overlooked forever.

As to the source of the magic, it's possible that it comes from very old legacy constructs that have resulted in these constant numbers being stored in the DB, @ehuelsmann may be able to cast more light on that, and the best way to handle it as well

sbts:
I have to agree with @ehuelsmann , I think allowing -1 in critic makes a lot of sense, and maintains clarity.
Also, I agree that we need some sort of centralised code to manage sign changes, there are so many places we repeat that process, but with differences in how the code is written despite the fact that the intent and functionality is identical.

@ziuq If you could please split this PR into at least 2 separate PR's, 
1. Changing magic numbers to constants (except where it involves -1 for negation)
This PR should also enable the use of -1 in critic config as suggested by ehuelsmann
2. proposed negation changes

That should allow the first PR to be reviewed and merged quickly and easily.
Then we can work out how to manage the required/desired changes to negation.

lib/LedgerSMB/Scripts/recon.pm
```diff
@@ -274,10 +275,13 @@ sub _display_report {
 
     $recon->{zero_string} = LedgerSMB::PGNumber->from_input(0)->to_output(money => 1);
 
-    $recon->{statement_gl_calc} = $neg_factor *
-                                    ($recon->{their_total}
+    $recon->{statement_gl_calc} = $is_positive
+                                  ?  ($recon->{their_total}
```
sbts: @ehuelsmann can you please check this one as well

sbts:
@ziuq I hope we aren't discouraging you with the comments on this PR.
The bulk of it is great and exactly the sort of result we were hoping to get with extending the critic tests.

It really is just the risks associated with making changes to the financial logic (negation code in this case) that has me being very picky

lib/LedgerSMB/Magic.pm
ziuq: Great, that will be my first move out of the ascii world.  The baby takes its first steps.

lib/LedgerSMB/Scripts/recon.pm
ziuq: I'll recheck.   I'd be surprised if I didn't run a truth table through the new and old code.

lib/LedgerSMB/Magic.pm
ziuq: And  I just find it mechanistically easier to have them in one place.  As more magic is resolved, there is one place to look for dupes or collisions. With codes, it will be efficient to add `use this` and `use that` at once.
I am surprised no one has commented on the actual naming of these constants,  as an outsider OEC_xxxxx seems only okay.  I expected to be changing these.

There is a boundary issue with the EDI_PATHNAME_MAX.   ?

lib/LedgerSMB/Report/Listings/TemplateTrans.pm
ziuq: My guess is that it equates to journal types in the database, but the hashed list was too short.
That seems pretty obvious to me now.  But I'd rather be overcautious than...

ziuq:
Thanks.  No, I am not discouraged.   
I think there are still some of my questions open above.
Given Perl's C roots, I would sleep on demystifying -1.
As I indicated before, I have no problem reworking these changes.
I do not know the protocol to split a pull request.   I would expect to start a branch and continue by moving fixes from this branch  to the other just going forward on both.  Does that meet expectations here.

lib/LedgerSMB/Scripts/recon.pm
ziuq: I've rechecked for what its worth.

I did NOT run a table through this passage, it was the PNL passage above.

lib/LedgerSMB/Template/XLSX.pm
ehuelsmann: @ziuq absent a response by @ylavoie, I think we can remove these variables entirely. They're not used and setting to zero end decrementing is just too ugly for words for a variable which isn't adding value.

(I'm referring to both `$rowcount` and `$currcol`.)

lib/LedgerSMB/Budget.pm
ehuelsmann: @ziuq absolutely agreed. It's definitely one of the things that the Critic tests bring to a code base: a good reason to go over the code once more with a different pair of eyes and intentions. It helps to coax out latent problems.

lib/LedgerSMB/Entity/Note.pm
ehuelsmann: It's my preference not to run perltidy on the code base (at least not yet): there will be a break in the 'blame' output, making it harder to trace line  changes back to their original commit.

lib/LedgerSMB/Magic.pm
ehuelsmann: As the code said 180, lets stick to 180. I have no way to test either...

lib/LedgerSMB/Report/Listings/TemplateTrans.pm
ehuelsmann: These constants correspond with the items at https://github.com/ledgersmb/LedgerSMB/blob/master/sql/Pg-database.sql#L1197-L1202

lib/LedgerSMB/Scripts/recon.pm
```diff
@@ -235,10 +236,10 @@ sub _display_report {
     for my $field (qw/ total_cleared_credits total_cleared_debits total_uncleared_credits total_uncleared_debits /) {
       $recon->{"$field"} = LedgerSMB::PGNumber->from_input(0);
     }
-    my $neg_factor = 1;
+    my $is_positive = 1;
     if ($recon->{account_info}->{category} =~ /(A|E)/){
```
ehuelsmann: It's correct that positive assets (and expenses) are accounted/stored as negative values.

lib/LedgerSMB/Entity/Note.pm
ziuq: Good to know,  there is the comment about reformating code at the end of changes.  And I was looking forward to that; I will diminish my expectations.  I don't think I'm fussy regarding format, but I find long lines difficult  and consistent indentation helps me a lot with finding scopes.  

Can one get stacked snapshots of blame?   Or  is there an easy way to view the commits of a file in reverse order through time?

Is the difficulty you mention just that there is an extra commit?

lib/LedgerSMB/Tax.pm
ehuelsmann: Running a quick test from the command line shows that the original works.

```bash
$ perl -MData::Dumper -e '@a=(1,2,3); print Dumper(\@a); $#a=-1; print Dumper(\@a);'
$VAR1 = [
          1,
          2,
          3
        ];
$VAR1 = [];
```

I have no idea why anybody would want to set the index of the last element. I'm all for the cleanup that was applied here.

lib/LedgerSMB/Report/Listings/TemplateTrans.pm
ziuq: Thanks, I use that.

lib/LedgerSMB/Template/LaTeX.pm
ehuelsmann: @einhverfr noticed that the character was (incorrectly?) encoded by TeX::Encode. So he removed the element from the hash. I think it's better if there had been a comment as to what situation and TeX::Encode version this problem applied to.

my preference would be that the new code used:

```perl
    delete $TeX::Encode::charmap::ACCENTED_CHARS{"\N{LATIN CAPITAL LETTER A WITH RING ABOVE}"};
    delete $TeX::Encode::charmap::ACCENTED_CHARS{"\N{LATIN SMALL LETTER A WITH RING ABOVE}"};
```

ehuelsmann:
All in all great work! Let me know which next-steps you see and want to persue.

lib/LedgerSMB/Scripts/recon.pm
```diff
@@ -274,10 +275,13 @@ sub _display_report {
...
-    $recon->{statement_gl_calc} = $neg_factor *
-                                    ($recon->{their_total}
+    $recon->{statement_gl_calc} = $is_positive
+                                  ?  ($recon->{their_total}
                                     + $recon->{outstanding_total}
-                                    + $recon->{mismatch_our_total});
+                                    + $recon->{mismatch_our_total})
+                                  : - ($recon->{their_total}
+                                    + $recon->{outstanding_total}
+                                    + $recon->{mismatch_our_total}) ;
```
ehuelsmann: I like this code as an example of how we might want `-1` *not* to be marked as magical: the consequence is code duplication if we don't. I'm afraid the result of this code duplication is that we end up with mismatching implementations.

lib/LedgerSMB/Report/Balance_Sheet.pm
```diff
@@ -255,11 +255,10 @@ sub run_report {
...
 
-            my $sign = (($contra) ? -1 : 1)
-                * ((($cat eq 'A') || ($cat eq 'E')) ? -1 : 1);
+            my $is_positive = !!$contra == ( $cat eq 'A' or $cat eq 'E' );
 
-            $self->cell_value($row_id, $col_id, $sign * $value)
-                if $sign < 0;
+            $self->cell_value($row_id, $col_id, -$value)
+                unless $is_positive;
         }
```
ziuq: There is a joke in the original.

sbts: @ziuq can you please explain, @ehuelsmann and I can't see the issue.

ziuq: There is not any issue.   I didn't mean to put you to work, I am sorry it I did.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -65,15 +66,15 @@ sub _get_database {
...
+    return [ HTTP_UNAUTHORIZED,
              [ 'WWW-Authenticate' => 'Basic realm="LedgerSMB"',
                'Content-Type' => 'text/text; charset=UTF-8' ],
              [ 'Please enter your credentials' ] ]
         if ! defined $creds->{password};
 
     # Ideally this regex should be configurable per instance, and possibly per admin user
     # for now we simply use a fixed regex. It will cover many if not most use cases.
-    return [ 454,
+    return [ HTTP_454,
```
ziuq: HTTP_UNAUTHORIZED?

lib/LedgerSMB/X12/EDI850.pm
ziuq: Though not authoritative, page 3 of http://www.1edisource.com/Portals/0/docs/850V5010.pdf seems to refer to the proper ANSI specification.  Our indices clearly refer to the header elements sequenced by the ISO prefixes.
Using the descriptions on that page to generate constant names seems somewhat helpful.

lib/LedgerSMB/Sysconfig.pm
```diff
@@ -163,7 +167,7 @@ def 'force_username_case',
 
 def 'max_post_size',
     section => 'main',
-    default => 1024 * 1024,
+    default => MEGABYTE,
```
ziuq: I suggest not  doing a no-critic; you are right, but the person who does not see that is not wrong.

Although the codebase has its odd quirks and warts at present it has four  no-critic directives (soon to be three).
I have suggested a fairly loose application of no-critic on Lsmb.  I was wrong; we should avoid `## no critic` like the plague.

sbts: @ziuq the problem I have with putting this constant declaration ANYWHERE else except in Sysconfig.pm it's self is quite simple.
The line `default => 4194304,` is, in it's self, the declaration of that constant.

All of the `default` values in this file are constants, (and as such we probably should fix the declaration's and or code taking those to reflect that.)

Declaring an intermediate constant does not in any way help readability nor maintainability of the codebase, in fact it hinders both.

As such I am extremely resistant to hiding this value behind a constant declaration.

@ehuelsmann needs to have the final say here, but I suspect he will agree that it should stay put.

I'm not certain, but if we can slightly alter the code here so that critic considers it a constant declaration (which it is) rather than use of a magic number, that would mean not needing "no-critic"

lib/LedgerSMB/Scripts/setup.pm
sbts: @ziuq Hmm, 401 is unauthorized.
I think we may need to ask @ehuelsmann to ensure he is still happy with the choice of 454.
I actually added this code some time ago after discussion with ehuelsmann.
454 was chosen so we could present some specific information back to the client (Prohibited User)
It is not an standards assigned Status Code, however, it's a code that is unlikely to be assigned any time in the forseable future (there are many unused 4xx codes free below this one)

This is the  [Official HTTP Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml#http-status-codes-1)

Looking at that, and the more verbose, but incomplete wikipedia [List of HTTP Status Codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) there really isn't a perfect match for our needs here.
However, if we were to change it, I'd suggest one of the "specifically cacheable" codes such as 404 which may assist with hardening against attacks.
This introduces issues we currently don't have a mechanism to resolve (in setup.pm) where the Client is unable to reliably display a useful error message to the User.

__NOTE: ANY CHANGE to this constant requires changes to be made in html and or .js files as well__

As for naming this constant, regardless of what number is used, I'd suggest
"HTTP_PROHIBITED_USERNAME" 
Even if we do later on change the actual response number this is descriptive and shouldn't need to be changed

lib/LedgerSMB/Scripts/login.pm
```diff
@@ -134,7 +135,7 @@ sub authenticate {
                      [ 'Content-Type' => 'text/plain; charset=utf-8' ],
```
sbts: We should change the preceding line so that the return string is a constant.
The constant declaration should be something like.
`use constant { HTTP_454_DB_NOT_EXIST => '454 Database Does Not Exist', };`

elsewhere I proposed adding `HTTP_PROHIBITED_USERNAME => '454'` in light of this additional use of 454 I think that should change to be....
`use constant { HTTP_454_PROHIBITED_USERNAME => '454 Prohibited Username', };`

Any other uses of 454 should then follow the same pattern, unique constants declared for each type of error message.

lib/LedgerSMB/Scripts/setup.pm
sbts: in light of other uses of 454 I think this should change to be....
use constant { HTTP_454_PROHIBITED_USERNAME => '454 Prohibited Username', };

As we use 454 to present error's directly to our client UI for display

lib/LedgerSMB/Sysconfig.pm
```diff
@@ -163,7 +167,7 @@ def 'force_username_case',
 
 def 'max_post_size',
     section => 'main',
-    default => 1024 * 1024,
+    default => MAX_POST_SIZE_DEFAULT,
```
sbts: @ziuq the problem I have with putting this constant declaration ANYWHERE else except in Sysconfig.pm it's self is quite simple.
The line `default => 4194304,` is, it's self, the declaration of that constant.

All of the default values in this file are constants, (and as such we probably should trye to make the declaration's behave as such.)

Declaring an intermediate constant does not in any way help readability nor maintainability of the codebase, in fact it hinders both.

As such I am extremely resistant to hiding this value behind a constant declaration.

@ehuelsmann needs to have the final say here, but I suspect he will agree that it should stay put.

I'm not certain, but if we can slightly alter the code here so that critic considers it a constant declaration (which it is) rather than use of a magic number, that would mean not needing "no-critic"

lib/LedgerSMB/Magic.pm
```diff
@@ -0,0 +1,389 @@
...
+=head1 SYNOPSIS
+
+ use LedgerSMB::Magic qw( EC_EMPLOYEE );
+
+ if ($entity_code == EC_EMPLOYEE) {
+     print "This entity  is an employee.";
+ }
+
+=head1 DESCRIPTION
```
sbts: Please move the POD sections for each group of constants to be immediately before each group.
That assists with consistency between the code and documentation in the long run.

lib/LedgerSMB/Magic.pm
```diff
@@ -0,0 +1,389 @@
...
+    RC_IMPORT           => 3,
+    RC_PARTIAL_DISPOSAL => 4,
+
+    # codes
+
+    # EDI
+    EDI_PATHNAME_MAX => 180,    # default max length of EDI pathname
+                                # TODO possible fencepost error XXX
+
+    # HTTP  These can be taken from HTTP::Status
```
sbts: As previously discussed with @ehuelsmann we should be using HTTP::Status rather than re-declaring constants ourselves.

The exception here would be HTTP_454 which will need a number of constants declared here as it's not a standards defined response code, and we use it to return multiple error types

lib/LedgerSMB/Magic.pm
```diff
@@ -0,0 +1,389 @@
...
+    JRNL_GJ => 1,
+    JRNL_AR => 2,
+    JRNL_AP => 3,
+    JRNL_CR => 4,
+    JRNL_CD => 5,
+
+    # our note classes
+    NC_ENTITY                => 1,
+    NC_INVOICE               => 2,
+    NC_ENTITY_CREDIT_ACCOUNT => 3,
```
sbts: can @ehuelsmann or @einhverfr identify what constant 4 should be called?
I think we should ensure there are no gaps in the sequence. If it's unused/invalid, lets at least have a commented out declaration showing so.

lib/LedgerSMB/Template/XLSX.pm
ziuq: @ehuelsmann if the vars are unnecessary, the subroutine is unnecessary, here (and in XLS.pm?).

lib/LedgerSMB/Scripts/setup.pm
ziuq: I will defer to greater wisdom for not having any HTTP experience.  The lack of an allowed username or of passwords/keys must have a standard answer.  It looks to me like a authentication/authorization issue.

lib/LedgerSMB/Report/PNL.pm
```diff
@@ -246,7 +246,7 @@ sub run_report {
             my $sign = (($contra) ? -1 : 1)
                 * ((($cat eq 'A') || ($cat eq 'E')) ? -1 : 1);
 
-            $self->cell_value($row_id, $col_id, $sign * $value)
+            $self->cell_value($row_id, $col_id, -$value)
```
ziuq: Yes.

lib/LedgerSMB/Scripts/asset.pm
```diff
@@ -24,7 +25,6 @@ use LedgerSMB::Report::Listings::Asset_Class;
 use LedgerSMB::Report::Listings::Asset;
 use strict;
 use warnings;
-
```
ziuq: Inadvertent.  The change is coming.

lib/LedgerSMB/Scripts/file.pm
```diff
@@ -34,15 +34,17 @@ use LedgerSMB::File::ECA;
...
-   6   => 'LedgerSMB::File::Internal',
-   7   => 'LedgerSMB::File::Incoming',
+   FC_TRANSACTION   => 'LedgerSMB::File::Transaction',
+   FC_ORDER   => 'LedgerSMB::File::Order',
+   FC_PART   => 'LedgerSMB::File::Part',
+   FC_ENTITY   => 'LedgerSMB::File::Entity',
+   FC_ECA   => 'LedgerSMB::File::ECA',
+   FC_INTERNAL   => 'LedgerSMB::File::Internal',
+   FC_INCOMING   => 'LedgerSMB::File::Incoming',
 };
```
ziuq: Done.  

Expect to keep prodding me on this one.   My practice is to always put edited files through perltidy.

lib/LedgerSMB/Scripts/login.pm
ziuq: A lsmb database not existing could be looked at as a page not found error.

lib/LedgerSMB/Magic.pm
ziuq: I thought I had indicated my agreement and my desire to clean up Magic.pm later for efficiency in the task.  I'd like to do that at the end of dealing with magic.  If that delay  is burdensome, let me know.  I was thinking of a second PR for the second half of magic numbers and finishing with Magic.pm--works for me but does it work for you?

http://www.1edisource.com/Portals/0/docs/850V5010.pdf gives descriptions of the 16 indices  of EDI items.  These could be the basis for constant names if we cannot find an authoritative source.  Last I knew, getting actual specifications was expensive.

lib/LedgerSMB/Magic.pm
ziuq: There is no number four in `note_class`.

lib/LedgerSMB/Magic.pm
ehuelsmann: Adding a second PR with more magic numbers later is completely fine. @sbts's point is that there can't be a code defined elsewhere for the 454 status code, because I "invented" it.

ehuelsmann: @ziuq I'm not sure the EDI code works as-is, today. We best establish that before diving into it deeper. If it doesn't take you too much time, feel free to define the 16 indices with constants.

lib/LedgerSMB/Report/Orders.pm
```diff
@@ -15,6 +15,7 @@ use namespace::autoclean;
 extends 'LedgerSMB::Report';
 with 'LedgerSMB::Report::Dates';
 use LedgerSMB::MooseTypes;
+use LedgerSMB::Magic qw( OEC_SALES_ORDER OEC_PURCHASE_ORDER OEC_QUOTATION OEC_RFQ );
 
 =head1 DESCRIPTION
```
sbts: @ziuq you can safely ignore this exploratory comment. @ehuelsmann and I will discuss later, but it may not be worth making a change here.

lib/LedgerSMB/Scripts/login.pm
sbts: @ziuq for now, leave all (not yet done) changes to 454 status codes.
I'll work on that later.

lib/LedgerSMB/Sysconfig.pm
ehuelsmann: I think we should interpret these defaults as the canonical definition of the value. Meaning we need `## no critic` here.

lib/LedgerSMB/Tax.pm
ehuelsmann: I think the replacement code (while I still think that `-1` isn't magical) is better than the original. Lets leave this one in.

ehuelsmann:
@ziuq thanks for your work on this PR! With these three items added:

* [ ] MAX_POST_SIZE_DEFAULT replaced with the number 4194304 and a ##no-critic
* [ ] POD moved close to the associated declarations
* [ ] placeholder added for NC_UNKNOWN_4

we can definitely merge this PR.

ehuelsmann:
I've taken the liberty to resolve the outstanding conflict.

ehuelsmann:
@ziuq the failures on Travis for Perl 5.18 through 5.14 are probably version related. Could you check and see if these can be easily addressed? I'd prefer not to drop those versions yet.

ziuq:
I did look about a bit to see if this is some known problem, with no hits (except 5.10 in 2008). 
I'll check but I think all my available hosts are running Perl 5.20 or newer.
We have pod coverage, that is not Perl version linked.  I'd suggest skipping the pertinent or all pod coverage tests on the earlier Perls.  
Checking that the pod is readable with 5.10 would be good.

ziuq:
I jump from Perl 5.8.4 to 5.20, so I am not going to spend time on the underlying issue.  I could rework magics numbers with Const::Fast or Readonly.

lib/LedgerSMB/Template/XLSX.pm
ehuelsmann: @ziuq yes. It's not necessary.

ziuq:
Things that might be addressed, none of much moment.

Master Lsmb::Template::LaTeX is probably fixed already,
undeclared $format around line 136.

Most of the Lsmb class codes in Magic.pm should be
gotten from Pg-database.sql.

20 years is short for a future years limit vs last
century in the face of 25 and 30 year loans.

EDI fencepost error exists.

Magic.pm POD publishes the values for enumerated ids.

Money epsilon is high.

ehuelsmann:
I was ready to merge this PR and locally checked the PR out to resolve the POD coverage failures. As it turns out, these are very hard to fix and I need inspiration: the combination of exported constants (or even exported functions?) and `namespace::autoclean` is the cause of these doc test failures. I found [this thread on perlmonks](http://www.perlmonks.org/?node_id=951831) which explains the problem.

ziuq:
Don't waste time on this problem.  Disable the test on all but the latest and future Perl versions.
The coverage is not changing from version to version.

Great advice, just doesn't sound  inspiring.

ehuelsmann:
@ziuq if you think it's better to run POD coverage testing on a different version, please submit a PR to do so. The only reason it's running on 5.18 is because that once *was* the latest and greatest...

Anyway, I changed it for you now. Will PR in a bit.

ehuelsmann:
Merged through #2872. Great work! Thanks!

PR #2837, opened 2017-05-24
Check syscalls

ziuq:
Not much to say, just threatened a bunch of syscalls with death if they don't meet expectations.


lib/LedgerSMB/Template/DB.pm
```diff
@@ -141,8 +141,11 @@ sub get_from_file {
     my $content = '';
     open my $fh, '<', $path
         or die "Failed to open template file $path : $!";
-    $content .= $_ while <$fh>;
-    close $fh;
+    {
+        local $/ = undef;
+        $content =  <$fh>;
+    }
+    close $fh or die "Cannot close file $path";
```
sbts: @ehuelsmann can you please critique this entire block of change.
It doesn't look right to me, feels like we may end up loosing bytes.
I am probably wrong though

ehuelsmann: @sbts the block changes a large(ish) number of (line) reads to a single line read by setting the EOL character variable to `undef`. It's roughly the UTF-8 compliant variant of slurping a file into memory.

ziuq:
I don't recall ever having a close fail, so I do not pretend to know how harmless it would be.  I tend to not check my closes, but I would admit to it being sloppy about handling a very small risk.

The way that critic policy writer is probably thinking about this is  along these lines:

> The code will change.  The filehandle parameter could be any expression, that expression could be wrong so throwing an  exception is the right thing to do.   So it is more about protecting the developers  from ourselves and each other.

Given that scenario, dieing is helpful because secure code does not do undefined things,
and dieing is the best chance of doing something defined.  And that should happen in dev.

Given  your scenario of the 'universe about to explode', why would we want a blind 
user poking keys trying to fix things.   The universe might get reset and we might
have saved some work for someone.

ehuelsmann:
@ziuq I took the liberty to resolve the merge conflicts with the base branch so we can merge this PR. The reason for the conflicts was a series of refactoring changes in the LedgerSMB::Template{,::*} modules.

ehuelsmann:
@sbts and I discussed that filtering the paths from the error messages is an orthogonal change which he'll issue a PR for.

ziuq:
Given that this is open source and that file locations can very often be derived from public information.  That apparently the attacker has some authorized access to the target Lsmb...  Well, I just don't understand.

What kind of attack are we intending to thwart?

PR #2850, opened 2017-05-30
Add comment to users table on username length

nick-prater:


nick-prater:
Closing to fix bad line terminator - colon instead of semicolon.

PR #2856, opened 2017-06-03
Silence deprecation warnings

ylavoie:
LedgerSMB isn't ready yet for PGObject 2. For the version to stay below 2 until then.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -29,7 +29,7 @@ requires 'Moose::Role';
 requires 'Moose::Util::TypeConstraints';
 requires 'MooseX::NonMoose';
 requires 'Number::Format';
-requires 'PGObject', '1.403.2';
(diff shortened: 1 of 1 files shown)


cpanfile
```diff
@@ -29,7 +29,7 @@ requires 'Moose::Role';
 requires 'Moose::Util::TypeConstraints';
 requires 'MooseX::NonMoose';
 requires 'Number::Format';
-requires 'PGObject', '1.403.2';
+requires 'PGObject', '< 2'; # '1.403.2';
```
ehuelsmann: If you need these to be numbers, then `1.403.2` is represented as `1.403002` so, we could have `>=1.403002 <2` which is a better representation of our requirements than simply smaller than 2, I'd say. Could you try that?

PR #2869, opened 2017-06-05
* No logging on Travis CI

ehuelsmann:
Logging pollutes the test logs and potentially obscures problems


ehuelsmann:
Does not do what it needs to.

PR #2878, opened 2017-06-06
Fix entity table primary key

nick-prater:
The existing entity table primary key serves no purpose. It is keyed on
(control_code, entity_class), but there is already another UNIQUE constraint
on control_code.

This patch drops the current primary key, creating a new primary key on "id".

Discussed here:
https://matrix.to/#/!qyoLumPqusaXqFJNyK:matrix.org/$14961375623951384kSlUP:matrix.org
https://matrix.to/#/!qyoLumPqusaXqFJNyK:matrix.org/$14961411943985499EYfHE:matrix.org


ehuelsmann:
Thanks for the thorough review of the schema!

PR #2905, opened 2017-06-11
Fix menu pane header. Add tests for new menu header

ylavoie:


UI/css/ledgersmb.css
```diff
@@ -73,11 +73,15 @@ table.warning {
     border: 0;
 }
 
-#company_info, #version_info, #login_info {
+#version_info {
    font-weight: bold;
    font-size: 9pt;
 }
 
+#company_info, #login_info {
```
ehuelsmann: If these elements are also styled in 'expanded.css', shouldn't this be moved to there as well?

UI/menu/expanding.css
```diff
@@ -8,3 +8,28 @@ li.menu_closed > ul {
...
 }
+
+#version_info {
+    text-align: center;
+}
+
+.header_table {
+    margin-left: auto;
+    margin-right: auto;
+    border=none; border-height=0; padding-top=0; padding-bottom=0; margin-top=0; margin-bottom=0;
```
ehuelsmann: This is invalid CSS syntax, as far as I know. CSS uses a colon, not an equation sign.

xt/66-cucumber/11-ar/step_definitions/pageobject_steps.pl
```diff
@@ -19,9 +19,6 @@
 use Test::BDD::Cucumber::StepFile;
 
 
-use Data::Dumper;
-
-
```
ehuelsmann: Even though Data::Dumper should probably not be there, it should be PR-ed in a separate PR. (11-ar has really nothing to do with this PR...)

xt/lib/PageObject/App/Menu.pm
```diff
@@ -76,14 +76,14 @@ my %menu_path_pageobject_map = (
...
     my ($self) = @_;
 
-    my @logged_in_found =
-        $self->find_all('*contains', text => "Logged in as");
-    my @logged_into_found =
-        $self->find_all('*contains', text => "Logged into");
+    my @logged_in_company =
+        $self->find_all("//*[\@id='company_info' and string-length(normalize-space(text())) > 0]");
+    my @logged_in_login =
+        $self->find_all("//*[\@id='login_info' and string-length(normalize-space(text())) > 0]");
```
ehuelsmann: Why do you stop checking the exact text? If there's an important reason to do that, it probably deserves mention in the log message. If there's not, I'd prefer we stick to checking that the screen has *exactly* the desired text.

xt/66-cucumber/11-ar/step_definitions/pageobject_steps.pl
ylavoie: Left over from other issue. Removing

UI/menu/expanding.css
ylavoie: Indeed.

xt/lib/PageObject/App/Menu.pm
ylavoie: Moving to trim in the server instead

UI/css/ledgersmb.css
ylavoie: Your call. They need to be in one place. I just fixed the layout but kept the actual distribution. This should be addressed in another PR.

ehuelsmann: Don't quite understand the "kept distribution" part: there's a #version_info in expanding too

ylavoie: It is duplicated in many places, see https://github.com/ledgersmb/LedgerSMB/search?utf8=%E2%9C%93&q=version_info&type=
This should be addressed in another PR

PR #2907, opened 2017-06-13
Check captures

ziuq:
Silent the P::C beast regarding regex captures not being checked for success, else $1 may be undef or a grandfather.

Minor rescoping.  Fairly straight-forward.


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -37,8 +37,12 @@ use LedgerSMB::Magic qw( EC_EMPLOYEE HTTP_454 PERL_TIME_EPOCH );
 use HTTP::Status qw( HTTP_OK HTTP_UNAUTHORIZED );
 
 my $logger = Log::Log4perl->get_logger('LedgerSMB::Scripts::setup');
-$LedgerSMB::VERSION =~ /(\d+\.\d+)./;
-my $MINOR_VERSION = $1;
+my $MINOR_VERSION;
+if ( $LedgerSMB::VERSION =~ /(\d+\.\d+)./ ) {
+    $MINOR_VERSION = $1;
+} else {
+        die "Error: Invalid data";
```
sbts: This one is a little trickier.
As long as we continue to use version numbers in the same way it should be fine.
If, for some reason, we change to not using the expected version number format, we shouldn't die.
A warning Message would make sense (after all the change could be a local one for testing purposes)

lib/LedgerSMB/Template/DB.pm
```diff
@@ -131,12 +131,15 @@ does not call this without carefully whitelisting values.
...
     if ($path =~ m|/.*:| ){
        die 'Cannot run on NTFS alternate data stream!';
     }
-    $path =~ m|(.*)/([^/]+)$|;
-    $fname = $2;
+    if ( $path =~ m|(.*)/([^/]+)$| ) {
+        $fname = $2;
+    }  else {
+        $fname = $path;
+    }
```
sbts: @ehuelsmann there is a change to the logic here. __*I THINK*__ it's a good change, but can you confirm there are no unwanted side affects please.

Original Logic
- fname = $path
- match then fname = $2
- the result is fname should end up as = $2, or an undefined value that may have already been in $2 from the previous match

New Logic
- If match fname = $2
- if not match fname = $path
- result is the raw path could end up in fname. Is this desirable?

What I infer from the original logic is we probably want the else clause to set $fname = '' to make the path relative as the most likely behaviour would be $2 = undef

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1115,8 +1115,11 @@ if (!$request->{exrate}) {
 # LETS GET THE DEPARTMENT INFO
 # WE HAVE TO SET $dbPayment->{department_id} in order to process
 if ($request->{department}) {
- $request->{department} =~ /^(\d+)--*/;
- $Payment->{department_id} = $1;
+    if ( $request->{department} =~ /^(\d+)--*/ ) {
+        $Payment->{department_id} = $1;
+    } else {
+        die "Error: Invalid data";
```
ziuq: It is a logic change.  Die as assertion.   The code implies the  author did not expect this match to ever fail, but this code does not control the input.

Assigning a usable default for the user seemed like bad business logic to me.

If ``department_id''s are constrained somewhere, please tell me where.

lib/LedgerSMB/Template/DB.pm
ziuq: I agree with your analysis.  And I think this change is worse that you do.  I'll rework it.

ziuq:
`die "some" "list" "that has no ending newline`  will append the filename, line number and input line if applicable.

That you suspect the `die`s will be called and that I suspect the code probably worked, is exactly why
I put in the `die`s.   We don't know. We don't know now.  We may know soon.  Then the  code changes somewhere else, and we don't know again.   `Die`ing forcefully shows us our false assumptions and demands the code be fixed.

lib/LedgerSMB/Scripts/payment.pm
ehuelsmann: Departments are defined in the `business_unit` table, with a specific `business_unit_class`. It's completely acceptable for a company to be set up without departments, in which case I'd expect the `$request->{department}` to be `undef` (not submitted).

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1216,10 +1219,18 @@ for (my $i=1 ; $i <= $request->{overpayment_qty}; $i++) {
...
      $request->{"overpayment_topay_$i"} = LedgerSMB::PGNumber->from_input($request->{"overpayment_topay_$i"});
-     $request->{"overpayment_account_$i"} =~ /^(\d+)--*/;
-     my $id = $1;
-     $request->{"overpayment_cash_account_$i"} =~ /^(\d+)--*/;
-     my $cashid = $1;
+     my $id;
+     if ( $request->{"overpayment_account_$i"} =~ /^(\d+)--*/) {
+        $id = $1;
+     } else {
+        die "Error: Invalid data";
```
ehuelsmann: Overpayment accounts aren't required to be set up on a company either. They are defined in the database in the `account` table, with referencing records in the `account_link` table where the `description` equals either `AR_overpayment` or `AP_overpayment`.

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1216,10 +1219,18 @@ for (my $i=1 ; $i <= $request->{overpayment_qty}; $i++) {
...
+     }
+     my $cashid;
+     if ( $request->{"overpayment_cash_account_$i"} =~ /^(\d+)--*/) {
+        $cashid = $1;
+     } else {
+        die "Error: Invalid data";
```
ehuelsmann: Actually in the current code, I'm expecting a cash account for overpayments to be selected. *However* I have a change planned where the "overpayments" table isn't shown when no overpayment accounts have been set up (because it can't be used anyway). When I implement that change, then this value too *can* be `undef`.,

lib/LedgerSMB/Template/DB.pm
```diff
@@ -131,13 +132,16 @@ does not call this without carefully whitelisting values.
...
-    my $fname = $path;
+
     if ($path =~ m|/.*:| ){
        die 'Cannot run on NTFS alternate data stream!';
     }
-    $path =~ m|(.*)/([^/]+)$|;
-    $fname = $2;
+
+    my ( undef, undef, $fname)  = File::Spec->splitpath( $path);
+    die "Bad template file name $fname, should have one dot."
```
ehuelsmann: Just one remark: the split being unequal to two could mean either fewer (1) or more (3+) values being returned. Elsewhere in the code I used `split(<regex>, <value>, 2)` to restrict to return at most 2. Is that sane here too?

ziuq: The old code line 140  was looking for templatename.formatname, where neither can have a dot.   Now that is error checked.   

I am assuming a name like templatename.fmtname.moreStuffWithOrWithoutDots is not a well formed name here.

Your `split <regex>, <value>, 2` will yield 1 or 2 fields ( or list elements).  One field being the entire string if there is no match.  If there are matches, two fields will result.  The first  match divides the string into two fields.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Shouldn't that be handled in a test? What good would it be for a user to receive such message?

ziuq: @ylavoie, what should be handled in a test?

lib/LedgerSMB/Scripts/payment.pm
ziuq: Will make it default to `undef`.  Thanks for the guidance.

lib/LedgerSMB/Scripts/payment.pm
ziuq: Will change.

If I don't set up one or more G/L overpayment accounts, where do I apply the excess of a payment?   A question for lsmb.user or .dev.

lib/LedgerSMB/Scripts/setup.pm
sbts: @ziuq and @ylavoie No, this shouldn't be in a test.
It's a valid regex for runtime, but it shouldn't die, rather it should simply set $MINOR_VERSION to undef or ''

lib/LedgerSMB/Template/DB.pm
ehuelsmann: Ok. Let's leave it as it is: requiring exactly 2 parts split by a single dot.

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: @sbts actually, the $MINOR_VERSION value is later interpolated into strings to find an upgrade script. I'd say that the only way to find that that no longer works is to have the setup script fail early. It won't hit any users, I assume, but it'll make Travis show we changed something that's not compatible with other modules, when we do.

ziuq: I can put the exception back, you guys decide.   If I'm touching this code again, renaming $MINOR_VERSION would be appealing, any recommendations?  I only come up with $CURRENT_VERSION as, at least, not misleading.

ehuelsmann: @ziuq I'd vote for `$CURRENT_MINOR_VERSION`

it's my call we leave the match in but store it in this renamed variable which then gets renamed elsewhere in the setup.pm script too.

PR #2912, opened 2017-06-14
* Make sure PGObject::Simple::Role is installed in the correct version

ehuelsmann:
before starting Travis CI tests...


ehuelsmann:
Taking a different approach.

PR #2914, opened 2017-06-16
No No-Warnings

ziuq:
The change to Report.pm is a logic change.  The existing sort is not
intuitive.  This is an example:

 -2.25xyz -1 <BLANK> +qqq -0 -0x -d 0 0x a d d1 1.4z 2 +2.72 5y 100

This can be duplicated, but probably is not wanted.

The rest were not as difficult as I expected.  If something seems
off, I can share my logic to expedite a critique.


lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1603,15 +1603,16 @@ if (($Payment->{"new_entity_id"} != $Payment->{"entity_credit_id"})&& !$Payment-
...
+    my $overpay = $overpayments[$ref];
+    push @ui_overpayments, {
+            id          =>  $overpay->{chart_id},
+            accno       =>  $overpay->{accno},
+            description =>  $overpay->{description},
+            amount      =>  $overpay->{movements},
+            available   =>  $overpay->{available},
+            touse       =>  $amount_to_be_used{$overpay->{accno}},
+    };
+    $ui_avble_subtotal += $overpay->{available};
```
ehuelsmann: I think we need to catch `$overpay->{available}` to be not-`undef` as well.

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1603,15 +1603,16 @@ if (($Payment->{"new_entity_id"} != $Payment->{"entity_credit_id"})&& !$Payment-
...
-    no warnings 'uninitialized';
-       push @ui_overpayments, {     id               =>  $overpayments[$ref]->{chart_id},
-                                    accno          =>  $overpayments[$ref]->{accno},
-                                    description    =>  $overpayments[$ref]->{description},
-                                    amount         =>  "$overpayments[$ref]->{movements}",
-                                    available      =>  "$overpayments[$ref]->{available}",
-                                    touse          =>  qq|$amount_to_be_used{"$overpayments[$ref]->{accno}"}|
-                            };
-       $ui_avble_subtotal += "$overpayments[$ref]->{available}";
+    my $overpay = $overpayments[$ref];
```
ehuelsmann: Shouldn't we skip the entire following segment when `$overpayments[$ref]` turns out to be `undef`?

lib/LedgerSMB/Report.pm
```diff
@@ -281,8 +281,8 @@ sub _render {
                        $srt_b = $srt_b->to_sort
                            if eval { $srt_b->can('to_sort') };
                    }
-                   no warnings 'numeric';
-                   $srt_a <=> $srt_b or $srt_a cmp $srt_b;
+
+                   $srt_a cmp $srt_b;
```
ehuelsmann: This one is indeed a bit tricky.

I'll study what it's supposed to handle and let you know what direction to go.

ziuq: Yes, let me know.   My thinking:
To make sense it needs `/[-\.\+]{0,1}\d+.*/` or /\D+/ values, but not both.  But a user could have a convention where positive ints are normal, and specials are non-numeric and so can be easily segregated.  But I did not get so far as seeing if such a convention would make any sense in with the sorted objects.  Or whether the "missorting" would be a significant aggravation to a user.
The list given above is an actual sorted sample.

lib/LedgerSMB/Scripts/payment.pm
ziuq: ... I don't believe that `undef` can be retrieved from acc_trans.amount into the view overpayments.

The added `$overpay` var is about shortening lines.  I hope people don't mind, I will generally want to shorten to 78 or less when editing.

lib/LedgerSMB/Scripts/payment.pm
ziuq: That would be a logic change.   The no-warning was only needed because of the concatenation or quoting of the values and for the addition.  But since overpayments.available is not null so must  `$overpayments[$ref]` be defined.  (Barring really weird abuse of the DBI).

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -417,8 +417,8 @@ sub run_backup {
         die $request->{_locale}->text('Error creating backup file');
 
     if ($request->{backup_type} eq 'email') {
-        # suppress warning of single usage of $LedgerSMB::Sysconfig::...
-        no warnings 'once';
```
ehuelsmann: This warning suppression was there to prevent a warning of `$LedgerSMB::Company_Config::settings` being used only once. Without this suppression, the warning will return. The solution is to create a function-based interface.

ziuq: You are wrong; there is no such warning.  That is why I am against the dynamic code in Sysconfig.pm.

I would give up Perl for the lack of this warning.

Witness that `Used::ok` is used okay:

use strict;
use warnings;

package Used;
use vars '$ok';

$ok = 1;
$Used::never = 2;

package main;

my $v = $Used::ok;
$v = $Used::misspelled;

Name "Used::misspelled" used only once: possible typo at ./once line 13.
Name "Used::never" used only once: possible typo at ./once line 8.

ehuelsmann: Well, all I know is that before I restructured Sysconfig to the current structure, we had lots of "used only once" errors in the code base. But I see your example does indeed not produce the warning. Nor do my attempts to prove I can generate one.

lib/LedgerSMB/Scripts/payment.pm
ehuelsmann: Yes 78 is the desired maximum length.

lib/LedgerSMB/Report.pm
ehuelsmann: Ok. I think the most often occurring sort criteria here should probably be on text fields or date fields (the latter being stringified first anyway).

lib/LedgerSMB/Scripts/setup.pm
ziuq: Basically it is a warning that should never be shut off.  It is always easy to silence.

einhverfr: How do you propose to silence the warning?

einhverfr: (might be helpful to have the proposed solution in the PR so the functionality is otherwise unchanged)

ziuq: @einhverfr,  if you refer to line 423, I am saying the line will not warn.  The solution is to just remove the
unnecessary `no warnings 'once';`

lib/LedgerSMB/Report.pm
einhverfr: The problem here is that there are cases where you want to sort on amounts.  I think this one ought to be looked at more deeply since it represents a difference in the output when you click on a header link.

lib/LedgerSMB/Scripts/setup.pm
einhverfr: Ok so the thing that worries me about this one is that if you look at the specific commit in question, it pretty much re-indents some code and adds this line.  It's a single commit which suggests it was there to silence warnings in logs.  Reviewing the code, I wonder if the fix was correct.  I don't see $csettings used again in the scope, so maybe that's the real cause of the warning that was silenced.  It's possible that something changed at one point and the warning came about because of another change and the warning was turned off.

So this is the sort of thing where manual testing might be a really good idea because maybe the wrong problem was fixed.

lib/LedgerSMB/Report.pm
einhverfr: (I think this one needs more discussion, and should probably be pulled out of this PR, with a bug filed instead, if I can give my vote.)

lib/LedgerSMB/Scripts/setup.pm
einhverfr: If manual testing is too onerous, I would be ok with accepting this line and filing a bug on the warning.  I am guessing that the problem is that $csettings does not appear to be used following.  And that should be a fairly easy one to isolate and fix.

einhverfr:
As a word of encouragement here, it is worth noting that this area is rather tricky in many of these parts of our code and that in some cases the silencing of the warnings may be more about design problems than code problems.  I think this PR is extremely valuable in starting a number of discussions.  And I want to thank you for adding it.

lib/LedgerSMB/Scripts/setup.pm
einhverfr: Actually due to use strict....  could you try to see if removing the assignment line itself throws a compile error?

ziuq: Good call.
Seduced by a comment's promised ease, the primrose path I tread; @einhverfr , I'll follow your rede.
Barring side effects, I'll just delete the line.

Please explain how to find the commit to which you refer; my git skills are still weak.  The best I've got is to `git log -L:run_backup:setup.pm`  which stops when the file was presumably moved.

lib/LedgerSMB/Report.pm
sbts: I agree with @einhverfr here.
Also, however we resolve it, some comments explaining what, why and how the code works/is-the-way-it-is could help prevent others from looking to simplify without deeper knowledge

lib/LedgerSMB/Scripts/setup.pm
sbts: @ziuq one way to dig into commit history and git blame info is to play around with the github interface.
Start out by browsing to [Scripts/setup.pm Line 421](https://github.com/ziuq/LedgerSMB/blob/master/lib/LedgerSMB/Scripts/setup.pm#L421)
Then at the top of the page you can click on "blame" or "history"

Once there you can click on either a commit (in History) or a blame entry (on the left in Blame) to get more info.

You can do the same from the commandline, but sometimes it's easier to learn what's possible from a Visual interface.

There are also some GUI tools you can install on your local desktop that may help.
These are just 2 of many
- `Git Cola`
- `Git DAG`

lib/LedgerSMB/Report.pm
ziuq: There is no good algorithm for a universal sort of user data.  There is a design issue of not bringing type info in with the rows.

I will put it back as it was with a no-critic.  If this code is going to remain, I will rewrite the current sort.

lib/LedgerSMB/Scripts/setup.pm
ziuq: Dave, thanks.

ziuq:
You are welcome and thanks for the word.   I have been down this road of warning-,  strict-, and/or p::critic-driven refactoring a few times.

ehuelsmann:
Thanks for even more patience!

PR #2915, opened 2017-06-17
Local punc vars

ziuq:
Localize puncutation variables.   This is trivial.

Live and learn--I would think of the %ENV as a punc var, but it is sensible that the policy would apply to all language defined :: variables.

### DIFF
--- a/lib/LedgerSMB/Sysconfig.pm
+++ b/lib/LedgerSMB/Sysconfig.pm
@@ -132,7 +132,7 @@ sub def {
         # If an environment variable is associated, set it  based on the
         # current value (taken from the config file, default, or pre-existing
         #  env var.
-        local $ENV{$envvar} = ${$name}                  
+        local $ENV{$envvar} = ${$name}
             if $envvar && defined ${$name};
 
         # create a functional interface
--- a/lib/LedgerSMB/Sysconfig.pm
+++ b/lib/LedgerSMB/Sysconfig.pm
@@ -495,7 +495,7 @@ sub check_permissions {
 
     my $tempdir = LedgerSMB::Sysconfig::tempdir();
     # commit 6978b88 added this line to resolve issues if HOME isn't set
-    $ENV{HOME} = $tempdir;
+    local $ENV{HOME} = $tempdir;
 
 
     if(!(-d "$tempdir")){
--- a/lib/LedgerSMB/Sysconfig.pm
+++ b/lib/LedgerSMB/Sysconfig.pm
@@ -128,7 +128,11 @@ sub def {
         }
 
         ${$name} = $ENV{$envvar} if ( $envvar && defined $ENV{$envvar} );  # If an environment variable is associated and currently defined, override the configfile and default with the ENV VAR
-        $ENV{$envvar} = ${$name}                                # If an environment variable is associated Set it based on the current value (taken from the config file, default, or pre-existing env var.
+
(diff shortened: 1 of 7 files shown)


ziuq:
I can delete the branch on ziuq and close this PR?

ehuelsmann:
@ziuq please do not close and delete: I think the bit with LSMB_ALWAYS_MONEY *is* valid. Can we retain that (and drop the other env vars for now)?

lib/LedgerSMB/Sysconfig.pm
```diff
@@ -128,7 +128,11 @@ sub def {
         }
 
         ${$name} = $ENV{$envvar} if ( $envvar && defined $ENV{$envvar} );  # If an environment variable is associated and currently defined, override the configfile and default with the ENV VAR
-        $ENV{$envvar} = ${$name}                                # If an environment variable is associated Set it based on the current value (taken from the config file, default, or pre-existing env var.
+
+        # If an environment variable is associated, set it  based on the
+        # current value (taken from the config file, default, or pre-existing
+        #  env var.
+        local $ENV{$envvar} = ${$name}
```
ehuelsmann: Basically, only this one and the one below are meant to modify global state, meaning that the others can be merged, but these two can't.

lib/LedgerSMB/Sysconfig.pm
```diff
@@ -495,7 +499,7 @@ sub check_permissions {
 
     my $tempdir = LedgerSMB::Sysconfig::tempdir();
     # commit 6978b88 added this line to resolve issues if HOME isn't set
-    $ENV{HOME} = $tempdir;
+    local $ENV{HOME} = $tempdir;
```
ehuelsmann: This is the other one meant to modify global state (which can't be merged).

einhverfr:
My 0.02 on this.  I am all for localizing all environment variables in the old code block.  That avoids the environment variables changing in relation to new code and helps cabin whatever shims we have to do for old code.

ehuelsmann:
@einhverfr, the old code runs forked per request thus isolates %ENV changes that way. Do we still need to localize?

lib/LedgerSMB/Sysconfig.pm
ziuq: That is in the queue.

We really should not be using `%ENV` as we are.   The idea of pushing our config variables into `%ENV` is just bad.

When you see a `$LSMB::Sysconfig::variable` you know it was processed--it is ours.  That leaves `%ENV`  just to be `local`ized and initialized in small call chains for "export" to subshells.
That means ignoring `%ENV` other than at startup.

For example LSMB_ALWAYS_MONEY is set many places and it is deleted in App_State, IIRC.
The flow is unknown to me, but what would happen if it were true at start up.  It is not vetted in Sysconfig and there are probably no tests.

`%ENV{HOME}` is set in Sysconfig.  Why? That is rhetorical, but it is hard to answer unless you were there.  `%ENV{HOME}` only appears in Sysconfig.  If it were brought into `$LSMB::Sysconfig::home`  it would self document when pushed out again.   

The enviros LSMB_ALWAYS_MONEY, LSMB_NOHEAD, REQUEST_URI are not handled by
Sysconfig.  That is not bad, but having a convention where enviros are generally set, makes the question require an answer.  When the answer is just that it was missed, it just reinforces my point that the work of forcing the environment is a lot of work for no gain.

The short version:  global variables are bad.

PR #2916, opened 2017-06-19
Document the need for gcc

sbts:


ehuelsmann:
@sbts could you document that gcc can be removed after the installation process with CPANM completes? I like that for completeness and to stress the install-time-only dependency.

sbts:
Yep. Shall do, and as I previously mentioned on #ledgersmb I'll also remove gcc from the system requirements section so it's just in the cpan section.

PR #2921, opened 2017-06-23
Remove prevention against non-existent warnings

ylavoie:
leftovers cleanup. LedgerSMB::_process_cookies used to complain about undefined entries. Those were left overs then _process_cookies was fixed.

### DIFF
--- a/t/02-number-handling.t
+++ b/t/02-number-handling.t
@@ -3,7 +3,6 @@
 use strict;
 use warnings;
 
-$ENV{TMPDIR} = 't/var';
 $ENV{REQUEST_METHOD} = 'GET';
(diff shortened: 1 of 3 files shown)


ehuelsmann:
@ylavoie from the description in the commit, it's not clear to me which problem this PR fixes. Please elaborate.

t/02-number-handling.t
```diff
@@ -3,7 +3,6 @@
 use strict;
 use warnings;
 
-$ENV{TMPDIR} = 't/var';
 $ENV{REQUEST_METHOD} = 'GET';
```
ehuelsmann: @ylavoie why do you remove REQUEST_METHOD below, but not here?

PR #2922, opened 2017-06-23
Move tempdir out of the source tree

ylavoie:


t/04-template-handling.t
```diff
@@ -30,10 +22,10 @@ Log::Log4perl::init(\$LedgerSMB::Sysconfig::log4perl_config);
 
 
 
-$LedgerSMB::Sysconfig::tempdir = 't/var';
+$LedgerSMB::Sysconfig::tempdir .= 't/var';
```
ehuelsmann: Why can't we just use the temp dir? I mean, why do we need to create `/tmp/t/var` ??

sbts: I don't really see a need to create a special tempdir, rather we should just use the tempdir that is provided by sysconfig.
For tests, that can be overridden (If necessary) by setting the environment variable `TEMP` before tests are run

ylavoie: Fixed

Issue #2930, opened 2017-06-25
Allow complete HTML validation


ylavoie:
HTML code is checked for validity to ensure quality, except when Template conditionals are used because of the difficulty to handle the conditions at test time.
We also skip menu validation because the code is generated dynamically. The upcoming switch to Dojo menus will fix that.

ehuelsmann:
Dojo and Vue work with html fragments too. In other words, it's unlikely we will refer validate the full html.

PR #2931, opened 2017-06-25
HTML checks for snippets

ylavoie:
Use HTML comments instead of Dojo comments to indicate HTML snippets and to allow skipping validation for expanding menus. This fixes the first part of #2930


ehuelsmann:
@ylavoie while I'm not against this change, I'm wondering what value it provides: aren't we just adding "doesn't need checking" criteria we're not using?

ylavoie:
We used to rely on file names to decide if a snippet could validate. This proposes a comment instead.
Use case is menu/expanding.html, where `li` and `ul` are dynamically inserted and defeat html check for balanced opening and closing tags.

ehuelsmann:
Thanks for that explanation. I think that belongs in the commit messages too. Putting it in the merge message instead.

PR #2935, opened 2017-06-27
Fixes inventory activity drill down

einhverfr:
This corrects the drill down.  Manually tested on a clone of demo.  Also fixes ar/ap transaction searches by part number.  This is a clear bug:  exposed in UI.  Supported in Perl modules.  Not supported on db backend. :-(


ehuelsmann:
Deserves mention in the Changelog for both 1.4.42 and 1.5.9

PR #2940, opened 2017-06-28
Fixes #2938

### LINKED ISSUE #2938: Goods and services search with ar AND ap selected broken
Returns 0 results.  In 1.2 returns the UNION of the results of ar and ap searches

einhverfr:
The filters for the goods search are totally wrong and were set to *exclude*
anything checked while the UI said it included anything checked.  This restores
old behavior of including and makes the behavior consistent with the UI

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -9,6 +9,7 @@ Changelog for 1.4.41
 * Refactor filter pages for income statement and balance sheet (Erik H)
 * Follow-up to earlier compensation for '.' missing from @INC (Erik H, #2275)
 * Fix reversal of payments against non-eca-default AR/AP account (Erik H, #2558)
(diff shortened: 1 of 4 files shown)


Changelog
```diff
@@ -9,6 +9,7 @@ Changelog for 1.4.41
 * Refactor filter pages for income statement and balance sheet (Erik H)
 * Follow-up to earlier compensation for '.' missing from @INC (Erik H, #2275)
 * Fix reversal of payments against non-eca-default AR/AP account (Erik H, #2558)
+* Fix goods/services search with AR/AP invoices (Chris T, #2938)
```
ehuelsmann: @einhverfr 1.4.41 has already been released. You want to start collecting into 1.4.42

PR #2946, opened 2017-07-01
Fixes #2783

### LINKED ISSUE #2783: permalink button mangles links the 2nd time around
Michael reported this in [this email](https://sourceforge.net/p/ledger-smb/mailman/message/35818601/)
Using a permalink results in the displayed permalink button presenting a link that invalidates the session.

I've just had a play on https://efficito.com/demo and reproduced this.

The steps are....

- General Jornal => Search and GL
- Select an Account  (1000--CURRENT ASSETS)
- Continue
- Click on [permalink]
You are now at the first link you shown below
- Click on [permalink] again
Your session is force expired as you have tried to visit the second link.

[LINK 1](https://demo.cloud.efficito.com/erp/1.5/login.pl?action=login&company=demo15#https://demo.cloud.efficito.com/erp/1.5/journal.pl?action=search&sort=transdate&reference=&accno=1000&source=&memo=&description=&notes=&business_unit_1=&business_unit_2=&business_unit_3=&business_unit_4=&business_unit_5=&bc_count=5&from_date=&to_date=&from_month=&from_year=&interval=none&from_amount=&to_amount=&category=X&is_approved=Y&col_transdate=Y&col_reference=Y&col_description=Y&col_debits=Y&col_credits=Y&col_source=Y&col_accno=Y&path=bin%2Fmozilla&login=demo_en&company=demo15#s2)
[LINK 2](https://demo.cloud.efficito.com/erp/1.5/login.pl?action=login&company=demo15#https://demo.cloud.efficito.com/erp/1.5/journal.pl?action=search&sort=transdate&reference=&accno=1000&source=&memo=&description=&notes=&business_unit_1=&business_unit_2=&business_unit_3=&business_unit_4=&business_unit_5=&bc_count=5&from_date=&to_date=&from_month=&from_year=&interval=none&from_amount=&to_amount=&category=X&is_approved=Y&col_transdate=Y&col_reference=Y&col_description=Y&col_debits=Y&col_credits=Y&col_source=Y&col_accno=Y&path=bin%2Fmozilla&login=demo_en&company=demo15&company=demo15#s1)
The only difference between the two links the end of the link
    &company=demo15#s2
vs
    &company=demo15&company=demo15#s1

einhverfr:
Fix tested.  From experimental to full status.  It is the right way to fix #2783  I think. 

The approach I took was to modify a copy of the query string based on whether it contains the relevant variable.  Simple, effective, readable.

### DIFF
--- a/UI/Reports/display_report.html
+++ b/UI/Reports/display_report.html
@@ -8,7 +8,13 @@ PROCESS "elements.html";
 PROCESS "dynatable.html";
 
 FORMATS = LIST_FORMATS();
-LINK = request.script _ '?' _ ENVARS.QUERY_STRING _ '&amp;company=' _ DBNAME;
+QSTRING = ENVARS.QUERY_STRING;
+
+IF NOT QSTRING.match('&amp;company=');
+   QSTRING = QSTRING _ '&amp;company=' _ DBNAME;
+END;
+
+LINK = request.script _ '?' _ QSTRING;
 LINK = LINK.replace('\&amp;\&amp;', '&amp;');
 
 ?>


sbts:
@einhverfr it would be preferable to include a brief description of what it fixes as well as the issue number.
That helps when reviewing changes and compiling changelists for a release etc.

In this case the issue title is
"permalink button mangles links the 2nd time around"

einhverfr:
@sbts it's good to keep these under 50 chars though because this becomes the merge commit title when merged in and Git assumes 50 char or less in the first line of the commit message.  Github doesn't do a great job with this, so would rather keep the titles easy to make valuable for history if given the choice so they print nicely on a git log --oneline --annotate --graph run.

einhverfr:
the last comment is assuming an 80-char wide terminal which is useful also.

einhverfr:
hmm looking at it, I guess it doesn't.  Sorry, will change approach here.  However branch name is important so we should probably insist on short branch names.

sbts:
No argument about keeping the first 50 chars a self contained description, but nothing to say we can't have multi line commit messages, and PR descriptions.
from bash, it's easy to do a multiline commit message on the commandline using something like....

`git commit SomeFile $'First Line\nSecond Line\nThird Line'`
the `$'` form of quoting in bash allows newlines and some other characters (eg single quote) to be inserted in what is otherwise a literal string

einhverfr:
sure.  but when Github merges a pr it creates a commit and we have no control over the content of that commit message other than the inputs that go into it.  So, my point is those inputs should be sufficiently small as to create a <50 char first line.  This means that the git username and branch cannot be longer than about 20 chars to be safe.

I was assuming the pr title would go into the commit title.  That would probably be saner than what Github actually does.  So I am suggesting we should insist on short branch names after proving myself wrong :-D

PR #2960, opened 2017-07-02
Fix ship/receive column alignments

einhverfr:
When we rewrite the order listing to use the new reporting engine in 1.4
we neglected to pass in all expected parameters.  This corrects this
oversight by using the existing order information to determine whether
we are shipping or receiving.


einhverfr:
Fixes #2960

PR #2965, opened 2017-07-02
1.5: always show plack logs on failure

sbts:


sbts:
NOTE: I cheated here and merged the master version of Is_LSMB_running.sh.
It had significant improvements, and nothing that shouldn't work on 1.5

PR #2966, opened 2017-07-03
Implement @one-db logic to use the same test db throughout a scenario

ylavoie:
Tests should run in a controlled environment and, by default, best practices dictate that the complete conditions be reproduced for each test.
Menu tests only require to make sure we can exercise every menu. Most of the test time is spent in creating a new test database and logging into the application for each menu. This PR removes the first delay by ensuring that we reuse the same database for each menu.

### DIFF
--- a/xt/lib/Pherkin/Extension/LedgerSMB.pm
+++ b/xt/lib/Pherkin/Extension/LedgerSMB.pm
@@ -104,6 +104,8 @@ sub pre_scenario {
     $stash->{"the admin password"} = $self->admin_user_password;
     $stash->{"the company"} = $self->last_feature_stash->{"the company"}
         if $self->last_feature_stash->{"the company"};
+
+    $stash->{one_db} = grep { $_ eq 'one-db'} @{$scenario->tags};
 }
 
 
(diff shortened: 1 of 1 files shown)


xt/lib/Pherkin/Extension/LedgerSMB.pm
```diff
@@ -162,10 +164,19 @@ sub create_user {
...
+sub check_database {
+    my ($self,$db) = @_;
+    my $returnstring = `psql -c "SELECT 1 FROM pg_database WHERE datname='$db'"`;
+    return grep { /^ +1$/ } split("\n", $returnstring);
+}
 
 sub create_template {
     my ($self) = @_;
 
+    if ($self->last_scenario_stash->{one_db} && $self->check_database($self->template_db_name)) {
```
ehuelsmann: This isn't the right place to change this: the template database must be created. It's the copying from the template to the test database which should be elongated

xt/lib/Pherkin/Extension/LedgerSMB.pm
```diff
@@ -202,25 +213,25 @@ sub create_template {
...
 
-
 sub create_from_template {
     my ($self, $company) = @_;
 
-    my $template = $self->template_db_name;
-    $self->ensure_nonexisting_company($company);
-    $self->super_dbh->do(qq(CREATE DATABASE "$company" TEMPLATE "$template"));
-
     $self->last_feature_stash->{"the company"} = $company;
```
ehuelsmann: Doesn't this break the callers contact? The result of this call is that the company does not exist after the call. I would say that there should be a modification to the callers to determine this call should not be made. What is the reason  you did it this way.

xt/lib/Pherkin/Extension/LedgerSMB.pm
ylavoie: The template must be created, once. If PSQL reports not found, let's create it. If we have one_db and the template is there, forget it.

xt/lib/Pherkin/Extension/LedgerSMB.pm
ylavoie: Same check as above, create once.

ehuelsmann:
As we established in the matrix/freenode channel, this branch isn't needed to prevent new databases from being created.

PR #2975, opened 2017-07-03
Use $tempdir to test BLACKLIST

ylavoie:
In order to have the source tree read-only, move the temporary file required to test the BLACKLIST into the temporary directory. Provide a switch to allow make to regenerate it in place.

### DIFF
--- a/t/06-blacklist.t
+++ b/t/06-blacklist.t
@@ -1,10 +1,13 @@
 #!/usr/bin/perl
 
+use LedgerSMB::Sysconfig;
+
 use Test::More tests => 3;
 use Digest::SHA 'sha512_base64'; #already a dependency
 use FindBin;
 
 my $sqldir = "$FindBin::Bin/../sql/modules";
+my $tempdir = $LedgerSMB::Sysconfig::tempdir;
 
 open my $blist, '<', "$sqldir/BLACKLIST";
 my $contents = join "", map { $a = $_; chomp $a; $a } <$blist>;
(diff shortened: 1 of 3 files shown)


t/06-blacklist.t
```diff
@@ -14,7 +17,7 @@ ok($contents, "Got contents from original blacklist");
 
 diag `perl $FindBin::Bin/../tools/makeblacklist.pl`;
 
-open $blist, '<', "$sqldir/BLACKLIST";
+open $blist, '<', "$tempdir/BLACKLIST";
```
ehuelsmann: are you sure you don't want to read the original blacklist?

t/06-blacklist.t
```diff
@@ -1,10 +1,13 @@
 #!/usr/bin/perl
 
+use LedgerSMB::Sysconfig;
```
ehuelsmann: I'm not sure I like this one. Can't you use File::Temp ? This test really doesn't have anything to do with the configuration of LedgerSMB...

sbts: @ehuelsmann I'm not so concerned about the use of Sysconfig.pm here.
We ave other tests that also use it so the precedent is set

t/06-blacklist.t
ylavoie: The original is read at line 12. This is the new copy and yes, I want to read it from the temporary copy I just generated.

t/06-blacklist.t
ylavoie: Using File::Temp would mean a new parameter for tools/makeblacklist.pl, which would mean parsing, the command line, validating...
The objective of the PR is to allow a read-only source tree. Proper usage of temp files and removal of sysconfig dependency should be the object of another, which address that more globally.

PR #2983, opened 2017-07-05
Check caps

ziuq:
This is a rework of 2907, to meet the last requests there.

In Template/DB.pm there is a small logic change.  If $path is just a basename, $fname will be undefined instead of reusing the value from the last match where $path contained slash(es).

In Scripts/setup.pm  $MINOR_VERSION got renamed $CURRENT_MINOR_VERSION as more descriptive.  Lowercase was too extreme for me.


ehuelsmann:
Thanks for your patience and your contribution!

PR #3001, opened 2017-07-12
Allow tests on a remote database

ylavoie:


xt/lib/Pherkin/Extension/LedgerSMB.pm
```diff
@@ -214,13 +214,13 @@ sub ensure_template {
...
-    $self->ensure_nonexisting_company($company);
-    $self->super_dbh->do(qq(CREATE DATABASE "$company" TEMPLATE "$template"));
+        my $template = $self->template_db_name;
+        $self->ensure_nonexisting_company($company);
+        $self->super_dbh->do(qq(CREATE DATABASE "$company" TEMPLATE "$template"));
 
     $self->last_feature_stash->{"the company"} = $company;
     $self->last_scenario_stash->{"the company"} = $company;
-    $self->_clear_admin_dbh;
+        $self->_clear_admin_dbh;
```
ehuelsmann: Any reason to indent some of the function (but not the rest)? Ideally, there would be no re-indenting here.

ylavoie: Leftover from a previous try. Fixed.

PR #3003, opened 2017-07-12
Use Widgets Tree and JSON as menu.

ylavoie:
Add menu store routines.
Add Cache to menus and cache module.
Add URL handling
Add CSS to remove duplicated icons
Add children to menu_item.
Whitelist menu_item.
Adjust BDD Tests for new menu structure

### DIFF
--- a/UI/js-src/lsmb/menus/Cache.js
+++ b/UI/js-src/lsmb/menus/Cache.js
@@ -1,0 +1,41 @@
+define(["dojo/_base/lang","dojo/when" /*=====, "dojo/_base/declare", "dojo/store/api/Store" =====*/],
+function(lang, when /*=====, declare, Store =====*/){
+
+// module:
+//        lsmb/menus/Cache
+// This is to query the masterstore and rely on it for get.
+// This should be refactored to use the standard cache and only have
+// a specialized getter
+
+var Cache = function(masterStore, cachingStore, options){
+    options = options || {};
+    return lang.delegate(masterStore, {
+        query: function(query, directives){
+            var results = masterStore.query(query, directives);
+            results.forEach(function(object){
+                if(!options.isLoaded || options.isLoaded(object)){
+                    cachingStore.put(object);
+                }
+            });
+            return results;
+        },
+        // look for a queryEngine in either store
+        queryEngine: masterStore.queryEngine || cachingStore.queryEngine,
+        get: function(id, directives){
+            return when(cachingStore.get(id), function(result){
+                return result || when(masterStore.query(directives), function(result){
+                    if(result){
+                        result.forEach(function(child) {
+                            cachingStore.put(child, {id: child.id});
+                        });
+                        result = cachingStore.get(id);
+                    }
+                    return result;
+                });
+            });
+        },
+    });
+};
+return Cache;
+});
+
--- a/sql/modules/Menu.sql
+++ b/sql/modules/Menu.sql
@@ -12,13 +12,16 @@ CREATE TYPE menu_item AS (
    label varchar,
    path varchar,
    parent int,
-   args text[]
+   args text[],
+   children int[]
 );
 
 
 
 CREATE OR REPLACE FUNCTION menu_generate() RETURNS SETOF menu_item AS
 $$
+        WITH t (position,id,level,label,path,parent,to_args)
+        AS (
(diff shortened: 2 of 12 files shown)


xt/lib/PageObject/App/Menu.pm
```diff
@@ -8,6 +8,8 @@ use PageObject;
 use MIME::Base64;
 use Test::More;
 
+use Try::Tiny;
```
ehuelsmann: Are you using this module?

sql/modules/Menu.sql
```diff
@@ -12,13 +12,16 @@ CREATE TYPE menu_item AS (
    label varchar,
    path varchar,
    parent int,
-   args text[]
+   args text[],
+   children int[]
```
ehuelsmann: I can't find where we use the new `children`. Is it unused?

sql/modules/Menu.sql
```diff
@@ -102,6 +112,18 @@ It is thus similar to menu_generate() but it only returns the menu items
 associated with nodes directly descendant from the parent.  It is used for
 menues for frameless browsers.$$;
 
+CREATE OR REPLACE FUNCTION menu_item(in_id int) RETURNS SETOF menu_item
```
ehuelsmann: Do we really need this? I mean, we've sent the menu to the client already, haven't we? Can't the client just use what we've sent it?

UI/js-src/lsmb/menus/Cache.js
```diff
@@ -0,0 +1,41 @@
+define(["dojo/_base/lang","dojo/when" /*=====, "dojo/_base/declare", "dojo/store/api/Store" =====*/],
```
ehuelsmann: Why do we need  a specific cache when we already implement a very specific REST store below?

sql/modules/Menu.sql
ylavoie: Nope, used directly below in menu_generate and in the Dojo MenuRestStore

sql/modules/Menu.sql
ylavoie: Same for LedgerSMB::DBObject::Menu::generate_(section|item)
I indended to decide about those after this and Preferred menu item PRs

UI/js-src/lsmb/menus/Cache.js
ylavoie: Because we wanted to cache each child without having to pull the whole tree each time and remove any possibility for add of remove, which are present in the standard Dojo store cache.

ehuelsmann: But then, why not retrieve the whole tree at once? For most sessions the tree won't change during the session.

ylavoie: We do with that.

sql/modules/Menu.sql
```diff
@@ -102,6 +112,18 @@ It is thus similar to menu_generate() but it only returns the menu items
...
+AS $$
+SELECT * FROM menu_generate() where id = $1;
+$$ language sql;
```
ehuelsmann: Here we already have the entire tree, just don't pa it to our callers. How much impact would it be to send it?

sql/modules/Menu.sql
```diff
@@ -80,7 +83,14 @@ $$
                                            like c.path::text || ',%')
             GROUP BY n.position, n.id, c.level, n.label, c.path, c.positions,
                      n.parent
-            ORDER BY string_to_array(c.positions, ',')::int[]
+            ORDER BY string_to_array(c.positions, ',')::int[])
+        SELECT t.*, ch.children from t
+        LEFT JOIN (
+            SELECT parent, level, array_agg(id) as children
+            FROM t
```
ehuelsmann: Why do we do it this way instead of querying the children by the parent's id when we retrieve them? That doesn't require the join, instead returning records from an index search.

ylavoie: Because I need the childs for the Dojo menu_item

sql/modules/Menu.sql
ylavoie: We do send it at once. As I said before, Same for LedgerSMB::DBObject::Menu::generate_(section|item)
I indended to decide about those after this and **Preferred menu item** PRs

PR #3033, opened 2017-07-21
Issue 3020 add return value checks to method do

sbts:
- (fixes #3030)
- Checks the return value of `do()`
- generates a simple error message on stdout which would normally be seen in the web browser
- generates a detailed error message as a warn() on the console 
- paramaterises the filename to `do` for convenience
- adds a couple of missing semicolons from other code as noticed during this PR
- reformats a line that was obscure due to a pointless linebreak

### DIFF
--- a/lib/LedgerSMB/Scripts/asset.pm
+++ b/lib/LedgerSMB/Scripts/asset.pm
@@ -1009,7 +1009,7 @@ sub run_import {
     return begin_import($request);
 }
 
-eval {
+{
     local ($!, $@);
     my $do_ = "scripts/custom/asset.pl";
     if ( -e $do_ ) {
--- a/lib/LedgerSMB/Scripts/asset.pm
+++ b/lib/LedgerSMB/Scripts/asset.pm
@@ -1010,7 +1010,18 @@ sub run_import {
 }
 
 ###TODO-LOCALIZE-DOLLAR-AT
-eval { do "scripts/custom/asset.pl"};
+eval {
+    local ($!, $@);
+    my $do_ = "scripts/custom/asset.pl"
+    if ( -e $do_ ) {
+        unless ( do $do_ ) {
+            if ($! or $@) {
+                print "Status: 500 Internal server error (asset.pm)\n\n";
+                warn "Failed to execute $do_ ($!): $@\n";
+            }
+        }
+    }
+};
 
 1;
 
--- a/lib/LedgerSMB/Scripts/asset.pm
+++ b/lib/LedgerSMB/Scripts/asset.pm
@@ -1012,7 +1012,7 @@ sub run_import {
 ###TODO-LOCALIZE-DOLLAR-AT
 eval {
     local ($!, $@);
-    my $do_ = "scripts/custom/asset.pl"
+    my $do_ = "scripts/custom/asset.pl";
     if ( -e $do_ ) {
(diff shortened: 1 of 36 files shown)


lib/LedgerSMB/Scripts/asset.pm
```diff
@@ -1010,7 +1010,18 @@ sub run_import {
 }
 
 ###TODO-LOCALIZE-DOLLAR-AT
```
ehuelsmann: Actually, 2 lines below, you *have* localized "dollar-at" (`$@`), so this remark can go.

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -528,6 +528,17 @@ any later version.  Please see the included LICENSE.txt for more details.
 =cut
 
 ###TODO-LOCALIZE-DOLLAR-AT
```
ehuelsmann: Same here.

lib/LedgerSMB/Scripts/journal.pm
```diff
@@ -135,5 +135,17 @@ files.
 =cut
 
 ###TODO-LOCALIZE-DOLLAR-AT
```
ehuelsmann: Same.

lib/LedgerSMB/Scripts/login.pm
```diff
@@ -207,7 +207,18 @@ sub logout_js {
 
 
 ###TODO-LOCALIZE-DOLLAR-AT
```
ehuelsmann: Yup. here too.

lib/LedgerSMB/Scripts/asset.pm
```diff
@@ -1010,7 +1010,18 @@ sub run_import {
...
-eval { do "scripts/custom/asset.pl"};
+eval {
```
ehuelsmann: these eval{}s are here to suppress errors such as files not existing. Probably *just* to suppress the file-does-not-exist error.

Since we prevent that error now, I'd say that we don't need to wrap an eval{} around the code block.

sbts: Ok, I'll strip them out. I suspected that was the case, but wasn't certain so left them behind.

sbts:
@ehuelsmann I've placed this on hold pending discussion of 2 things with you.
- switching to die() instead of print + warn in some locations
- bdd tests to cover the affected code in it's "normal" use

old/lib/LedgerSMB/Reconciliation/CSV.pm
```diff
@@ -23,13 +23,20 @@ use Try::Tiny;
...
+    no warnings;
+    opendir (DCSV, 'old/lib/LedgerSMB/Reconciliation/CSV/Formats');
+    for my $format (readdir(DCSV)){
+        if ($format !~ /^\./){
+            local ($!, $@);
+            my $do_ = "old/lib/LedgerSMB/Reconciliation/CSV/Formats/$format";
+            unless ( do $do_ ) {
+                if ($! or $@) {
+                    print "Status: 500 Internal server error (CSV.pm)\n\n";
+                    warn "Failed to execute $do_ ($!): $@\n";
```
ehuelsmann: This one needs to be a die() too

PR #3040, opened 2017-07-24
Less Numbing Numbers

ziuq:
Add 100 to non-magical list.
Build a list by name instead of id.
Locally name constants for the indices of view chart.
Shorten a line.
Use split and list assignment instead of if if else loop.

### DIFF
--- a/lib/LedgerSMB/Locale.pm
+++ b/lib/LedgerSMB/Locale.pm
@@ -155,9 +155,9 @@ sub date {
     }
 
     $dd *= 1;
-    $yy += 2000 if length $yy == 2;
-    $dd = substr( "0$dd", -2 );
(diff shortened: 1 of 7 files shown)


lib/LedgerSMB/Locale.pm
```diff
@@ -155,9 +155,9 @@ sub date {
     }
 
     $dd *= 1;
-    $yy += 2000 if length $yy == 2;
-    $dd = substr( "0$dd", -2 );
-    $mm = substr( "0$mm", -2 );
+    $yy = "20" . $yy if length $yy == 2;
```
ehuelsmann: Agreed, but if we go by the style guide, we should be using single quotes for non-interpolated strings:

```perl
  $yy = '20' . $yy if ....;
```

PR #3043, opened 2017-07-25
Migration fixes

ylavoie:
- Show the version found when refusing to migrate a foreign database.
- The actual count for the migration steps is 1 too many.

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -204,7 +204,8 @@ sub login {
     _init_db($request);
     sanity_checks($database);
     $request->{login_name} = $version_info->{username};
-    if ($version_info->{status} eq 'does not exist'){
+    # $version_info->{status} isn't always defined
+    if (defined $version_info->{status} && $version_info->{status} eq 'does not exist'){
         $request->{message} = $request->{_locale}->text(
              'Database does not exist.');
         $request->{operation} = $request->{_locale}->text('Create Database?');
@@ -226,7 +227,7 @@ sub login {
         if (! defined $request->{next_action}) {
             $request->{message} = $request->{_locale}->text(
                 'Unknown database found.'
-                );
+                ) . $version_info->{full_version};
             $request->{operation} = $request->{_locale}->text('Cancel?');
             $request->{next_action} = 'cancel';
         } elsif ($request->{next_action} eq 'rebuild_modules') {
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -759,13 +760,12 @@ sub _failed_check {
                    size => 15,
            }};
         push @$rows, $row;
-        $hiddens->{"id_$count"} = $row->{$check->id_column};
         ++$count;
+        $hiddens->{"id_$count"} = $row->{$check->id_column};
```
ehuelsmann: You need to move the count to before this line because you made the count 'zero-based' above. However, I don't think you need either change. In case the count needs to be zero-based, please add a comment as to why that is in the code.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -204,7 +204,8 @@ sub login {
     _init_db($request);
     sanity_checks($database);
     $request->{login_name} = $version_info->{username};
-    if ($version_info->{status} eq 'does not exist'){
+    # $version_info->{status} isn't always defined
```
ehuelsmann: The reason you're checking "$version_info->{status}" for defined-ness on this level is obvious. A better comment explains why it's not defined. Can you add a comment to that effect?

lib/LedgerSMB/Scripts/setup.pm
ylavoie: We absolutely need it, otherwise the count will always be one more than real.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Yes. The routine that sets it only do so on one branch of a if test. Logic remains ok, but lots of undefined messages are generated.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: For updates, that means updating the last one twice or trying to access an invalid id, for inserts, it would means duplicated entries.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -204,7 +204,9 @@ sub login {
     _init_db($request);
     sanity_checks($database);
     $request->{login_name} = $version_info->{username};
-    if ($version_info->{status} eq 'does not exist'){
+    # $version_info->{status} isn't always defined by get_info, so useless undefined messages
+    # are generated.
+    if (defined $version_info->{status} && $version_info->{status} eq 'does not exist'){
```
ehuelsmann: Your response on my comment here was quick (thanks) but not quite what I meant to ask for. I'll submit apr today to show you what I did mean.

PR #3047, opened 2017-07-26
T sl30 migration1

ylavoie:
This PR bring 2 new possibilities for migration:
1. Edition of multiple fields per record, either as selects or inputs
2. Allow inserting 1 simple record in an empty table. Business is a good example. SQL-Ledger has tables which are optional but mandatory in LedgerSMB. This eases the migration process.

It generalize the migration code by using the proper column id instead of the previous fixed value.
It also add back a default SQL extension to the upgrade file.

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -682,7 +682,6 @@ sub upgrade {
     my $database = _init_db($request);
     my $dbinfo = $database->get_info();
     my $upgrade_type = "$dbinfo->{appname}/$dbinfo->{version}";
-    my @selectable_values = ();
 
     $request->{dbh}->{AutoCommit} = 0;
     my $locale = $request->{_locale};
@@ -691,15 +690,20 @@ sub upgrade {
         next if ($check->min_version gt $dbinfo->{version})
             || ($check->max_version lt $dbinfo->{version})
             || ($check->appname ne $dbinfo->{appname});
-        if ( $check->selectable_values ) {
-            my $sth = $request->{dbh}->prepare($check->selectable_values);
-            $sth->execute()
-                or die "Failed to execute pre-migration check " . $check->name;
-            while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
-                push @selectable_values, { value => $row->{value},
-                                           text => $row->{id}
-                };
+        my @selectable_values = ();
+        for my $selectable_value (@{$check->selectable_values // []}) {
+            my @values = ();
+            if ( $selectable_value ) {
+                my $sth = $request->{dbh}->prepare($selectable_value);
+                $sth->execute()
+                    or die "Failed to execute pre-migration check " . $check->name;
+                while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
+                    push @values, { value => $row->{value},
+                                     text => $row->{id}
+                    };
+                }
             }
+            push @selectable_values,[@values];
         }
         my $sth = $request->{dbh}->prepare($check->test_query);
         $sth->execute()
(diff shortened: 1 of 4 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -738,33 +744,43 @@ sub _failed_check {
...
     my $hiddens = {table => $check->table,
-                    edit => $check->column,
-                           id_column => $check->{id_column},
-                            id_where => $check->{id_where},
+               id_column => $check->{id_column},
+                id_where => $check->{id_where},
+                  insert => $check->{insert},
                 database => $request->{database}};
+    my $i = 1;
+    # Move around Can't use string "ARRAY as an ARRAY ref while "strict refs" in use
```
ehuelsmann: This comment is a complete mystery to me... Are you sure it's both still relevant and the best way to document the code below it?
(Note: to me, the code below looks pretty straight forward...)

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -804,16 +820,37 @@ sub fix_tests{
...
+    my @edits;
+    my $i = 1;
+    # Move around Can't use string "ARRAY as an ARRAY ref while "strict refs" in use
+    while (defined $request->{"edit_$i"}) {
+      push @edits, $request->{"edit_$i"};
+      $i++;
+    }
+    my $sth = $request->{insert}
+      ? $request->{dbh}->prepare(
+            "INSERT INTO $table(" . join(',',@edits) . ") VALUES(" .
```
ehuelsmann: I'm pretty sure we want to quote identifiers from `@edits`, right?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -804,16 +820,37 @@ sub fix_tests{
...
+                join(',',map { '?' } @edits) . ")" )
+      : $request->{dbh}->prepare(
+            "UPDATE $table SET " .
+                join(',',map {$request->{dbh}->quote_identifier($_) . " = ?"} @edits) .
+                " where $where = ?");
```
ehuelsmann: Does `$where` come from the web? (It seems to have been retrieved from `$request` a little bit above...)

If that's the case, we *must* quote_identifier()

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -106,14 +106,23 @@ Repair query table to run once per result.
...
 Repair query column to get the values from
 
 =cut
 
-has selectable_values => (is => 'ro', isa => 'Str', required => 0);
+has selectable_values => (is => 'ro', isa => 'HashRef', required => 0);
+
+=item insert
+
+Allow insert instead of update. This to set defaults on a very limited subset
```
ehuelsmann: Allow, to me, means that both will be accepted if the boolean is true, which contrasts with "instead of".

Can we be more specific? Is it either/or or is it update or insert+update?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -137,7 +146,7 @@ Repair query column to run once per result
 
 =cut
 
-has column => (is => 'ro', isa => 'Str', required => 0);
+has column => (is => 'ro', isa => 'ArrayRef[Str]', required => 0);
```
ehuelsmann: should we make this a plural? `column` suggests one as it's a singular. I think the rest of the code uses plurals for arrays?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1104,7 +1141,7 @@ sub process_and_run_upgrade_script {
         template => $template,
         no_auto_output => 1,
         format_options => {extension => 'sql'},
-        output_file => 'upgrade',
+        output_file => 'upgrade.sql',
```
ehuelsmann: I thought that by setting the extension in the format options, the result would be a file *with* extension. Is this a bug elsewhere (which we may need to fix where it appears?)

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Good point. Fixed.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Yes they need to.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: We neved did, why now?

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Insert data instead of update.

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: > Does $where come from the web? If that's the case, we *must* quote_identifier()

That's why now.

ehuelsmann: A better solution would be to use the value from the Upgrade_Tests array element to which this data belongs

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Our Templater code doesn't work with objects in hiddens.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: But we are. There is no changes in $where handling.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: I though so too and would have prefered that but that ain't the case. I that the job of the templater?

einhverfr: If we are specifying an extension it should be recognised and managed by the template engine.  Taking a quick look

ylavoie: I thought so too and would welcome it. But it should be the object of another PR for it will bring unrelated files into this one and remove the focus.

einhverfr: Yeah, this is supposed to be supported by LedgerSMB::Template and LedgerSMB::Template::TXT

PR #3050, opened 2017-07-27
* Fix an error when trying to access the 'preferences' menu

ehuelsmann:
Found earlier by Yves, the PSGI authentication cleanup left one call
to the (now removed) LedgerSMB::Auth::get_credentials() routine in place.


ylavoie:
Menu is accessible but it then complains that Company does not exist when changing password.

ehuelsmann:
The complaint I will look at and solve using another issue

PR #3052, opened 2017-07-31
allow print and say. Apply critique.

ziuq:
This does nothing but require close and open to be checked.  The critique can be expanded to call regarding directories, etc.


ehuelsmann:
Agreed, as long as we're not debating *how* to check the result.

I'm all for warning, but against dying on the failure to `close()` a filehandle.


Nice next step would indeed be to start checking the `*dir()` family (opendir, readdir and closedir), taking the above remark into account, extending `close()` to `closedir()`.

PR #3053, opened 2017-07-31
Fix utf8, apply critique.

ziuq:
Fix utf8 and apply critique.


ehuelsmann:
Thanks for this critique. I'm wondering if the change is complete though. What about these occurrances of assignments to `$binmode` which are later used as arguments to `binmode`?

* https://github.com/ledgersmb/LedgerSMB/blob/8de8f5db1cbcbdf32597bbd596010166ffdaaf0f/lib/LedgerSMB/Template/CSV.pm#L19
* https://github.com/ledgersmb/LedgerSMB/blob/6846c671464201c60bcbe62e606d9d7599c6d3af/lib/LedgerSMB/Template/TXT.pm#L25

?

PR #3055, opened 2017-07-31
Not Grandmother's parlour

ziuq:
This no-critics code just to get critiques in force.  These are problematic spots of code where I lack the  wit, patience or both to provide better solutions.


einhverfr:
There are places where turning of strictures is the right thing to do.  For example, I am pretty sure Moo and Moose do so when setting up symbol table entries.  The question ought to be whether getting rid of the needs here would be an improvement or not.  But still, adding these no critic blocks allows better checks on other code so I am for that.

PR #3056, opened 2017-07-31
Master improved sl30 migration

ehuelsmann:


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -789,7 +789,7 @@ sub _failed_check {
     }
```
ylavoie: We could use scalar(@rows) here too.

PR #3059, opened 2017-07-31
* Make LedgerSMB::Database->get_info() adhere to its contract

ehuelsmann:
And by consequence, don't check for definedness in the caller to suppress
warnings...

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -204,9 +204,7 @@ sub login {
     _init_db($request);
     sanity_checks($database);
     $request->{login_name} = $version_info->{username};
-    # $version_info->{status} isn't always defined by get_info, so useless undefined messages
-    # are generated.
-    if (defined $version_info->{status} && $version_info->{status} eq 'does not exist'){
+    if ($version_info->{status} eq 'does not exist'){
         $request->{message} = $request->{_locale}->text(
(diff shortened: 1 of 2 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -204,9 +204,7 @@ sub login {
     _init_db($request);
     sanity_checks($database);
     $request->{login_name} = $version_info->{username};
-    # $version_info->{status} isn't always defined by get_info, so useless undefined messages
-    # are generated.
-    if (defined $version_info->{status} && $version_info->{status} eq 'does not exist'){
+    if ($version_info->{status} eq 'does not exist'){
```
ylavoie: I think that you opted for 'unknown' in lib/LedgerSMB/Database.pm#284

ehuelsmann: L284 sets the `appname` key to unknown. I removed the line right after which sets the `exists` key (because we don't advertize it in the API and I think it's a thinko). Then I added back for the entire block where the first connect succeeded, assignment of the `status` key. (The block where I removed the 'exists' is a sub-block of the block which sets the 'status' key.)

Then right here, we check the `status` key, so I think that the check can never use an `undef` value anymore.

ehuelsmann:
After discussion with Yves, merging.

PR #3060, opened 2017-08-01
* Document the mapping from 'dynatable-old' to 'dynatable'

ehuelsmann:
Note: 'dynatable-old' is deprecated and needs to be phased out. Doing
  so needs understanding of the mapping of old to new feature sets.
  Documenting the results of my investigation so far.


ehuelsmann:
@ylavoie with this commit, you should have a good starting point (together with the documentation of the *current* dynatable that I added yesterday) to migrate existing code over from form-dynatable (which uses dynatable-old) to something better.

PR #3062, opened 2017-08-01
Enhance migration functionalities by allowing tables to be extended (INSERT) when REFERENCES keys are missing

ylavoie:
Fix the business discounts for migration.
- Use null for unused entries instead of 0
- Ensure coherence between discount definition and usage in customer and vendor tables. Discounts usage have to refer to existing discounts.
- Provide means to either enter the missing discounts or remove their usage

### DIFF
--- a/lib/LedgerSMB/Locale.pm
+++ b/lib/LedgerSMB/Locale.pm
@@ -16,6 +16,12 @@ Locale support module for LedgerSMB.  Uses Locale::Maketext::Lexicon as a base.
 Returns a locale handle for accessing the other methods.  Inherited from
 Locale::Maketext.
 
+=item marktext
+
+Identity function. Allows text to me marked for translation so that it 
+will be picked by the PO scanner. The actual translation has to be done
+in the calling module when the page is prepared.
+
 =item text ($string, @params)
 
 Returns the translation for the given string.  Use this method with a litteral
@@ -85,8 +91,11 @@ package LedgerSMB::Locale;
 
 use strict;
 use warnings;
+use Exporter;
 
-use base 'Locale::Maketext';
+use base qw( Locale::Maketext Exporter );
+our @EXPORT_OK = qw(marktext);
+
 use LedgerSMB::Sysconfig;
 use Locale::Maketext::Lexicon;
 use Encode;
@@ -98,6 +107,10 @@ Locale::Maketext::Lexicon->import(
         _decode => 1,
     }
 );
+
+sub marktext {
+    return shift;
+}
 
 sub text {
     my ( $self, $text, @params ) = @_;
--- a/lib/LedgerSMB/Locale.pm
+++ b/lib/LedgerSMB/Locale.pm
@@ -16,11 +16,12 @@ Locale support module for LedgerSMB.  Uses Locale::Maketext::Lexicon as a base.
 Returns a locale handle for accessing the other methods.  Inherited from
 Locale::Maketext.
 
-=item marktext
+=item marktext ($string)
 
 Identity function. Allows text to me marked for translation so that it 
 will be picked by the PO scanner. The actual translation has to be done
 in the calling module when the page is prepared.
+Note: This isn't a method but a utility function.
 
 =item text ($string, @params)
 
@@ -91,7 +92,6 @@ package LedgerSMB::Locale;
 
 use strict;
 use warnings;
-use Exporter;
 
 use base qw( Locale::Maketext Exporter );
 our @EXPORT_OK = qw(marktext);
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -767,20 +767,15 @@ verify_check => md5_hex($check->test_query),
 
     my $heading = { map { $_ => $_ } @{$check->display_cols} };
     my %buttons = map { $_ => 1 } @{$check->buttons};
-    my %tooltips;
-    while ( my ($key, $value) = each $check->tooltips ) {
-      $tooltips{$key} = $request->{_locale}->text($value);
-    }
     my $buttons;
     push @$buttons,
            { type => 'submit',
              name => 'action',
             value => 'fix_tests',
           tooltip => { id => 'action-fix-tests',
-                       msg => $tooltips{'Save and Retry'},
-                       position => (qw/above below after before/)
+                       msg => $request->{_locale}->text($check->{tooltips}{'Save and Retry'}),
+                       position => 'above'
                      },
-             text => $request->{_locale}->text('Save and Retry'),
              text => $request->{_locale}->text('Save and Retry'),
             class => 'submit' },
         if $check->columns;
@@ -789,8 +784,8 @@ verify_check => md5_hex($check->test_query),
              name => 'action',
             value => 'cancel',
           tooltip => { id => 'action-cancel',
-                       msg => $tooltips{'Cancel'},
-                       position => (qw/above below after before/)
+                       msg => $request->{_locale}->text($check->{tooltips}{'Cancel'}),
+                       position => 'above'
                      },
              text => $request->{_locale}->text('Cancel'),
             class => 'submit' }
@@ -800,8 +795,8 @@ verify_check => md5_hex($check->test_query),
              name => 'action',
             value => 'force',
           tooltip => { id => 'action-force',
-                       msg => $tooltips{'Force'},
-                       position => (qw/above below after before/)
+                       msg => $request->{_locale}->text($check->{tooltips}{'Force'}),
+                       position => 'above'
                      },
              text => $request->{_locale}->text('Force'),
             class => 'submit' }
@@ -867,7 +862,7 @@ sub fix_tests{
 
     my $query;
     if ($check->{insert}) {
-        my @id = $id_displayed ? $request->{id_column} : ();
+        my @id = $id_displayed ? $check->{id_column} : ();
         push @id,@edits;
         my $columns = join(', ', map { $dbh->quote_identifier($_) } @id);
         my $values = join(', ', map { '?' } @id);
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -767,14 +767,19 @@ verify_check => md5_hex($check->test_query),
 
     my $heading = { map { $_ => $_ } @{$check->display_cols} };
     my %buttons = map { $_ => 1 } @{$check->buttons};
+    my %tooltips;
+    while ( my ($key, $value) = each $check->tooltips ) {
+      $tooltips{$key} = $request->{_locale}->text($value);
+    }
     my $buttons;
     push @$buttons,
            { type => 'submit',
              name => 'action',
             value => 'fix_tests',
           tooltip => { id => 'action-fix-tests',
-                       position => qw/above below after before/,
-                       msg => $check->{tooltips}{'Save and Retry'}},
+                       msg => $tooltips{'Save and Retry'},
+                       position => (qw/above below after before/)
+                     },
              text => $request->{_locale}->text('Save and Retry'),
              text => $request->{_locale}->text('Save and Retry'),
             class => 'submit' },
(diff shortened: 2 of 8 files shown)


ehuelsmann:
Wondering why you claim that LSMB requires one, while I see https://github.com/ledgersmb/LedgerSMB/blob/master/sql/Pg-database.sql#L882 which doesn't have a NOT NULL constraint. To that effect, the we don't require or need a business.

Can you explain a bit more?

ylavoie:
LedgerSMB requires null or defined id, SQL-Ledger required greater then 0. Fixed.

lib/LedgerSMB/Template/HTML.pm
```diff
@@ -34,6 +34,8 @@ sub escape {
     return undef unless defined $vars;
     #$vars = encode_entities($vars);
     $vars = escape_html($vars);
+    #Allow back only a few decorations.
+    $vars =~ s~&lt;(/?[biu]|br)&gt;~<$1>~g;
```
ehuelsmann: This is the wrong place to do that. What if someone added <b> to their invoice comments for some dark reason?

A better place to put the decoding is in the template itself, where it can be a specific contract between the string provider and the template, for specifically only those strings.

The HTML template sets up a function named `UNESCAPE` (https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Template/HTML.pm#L54) to allow the template to reverse escaping of specific strings.

lib/LedgerSMB/Locale.pm
```diff
@@ -16,6 +16,12 @@ Locale support module for LedgerSMB.  Uses Locale::Maketext::Lexicon as a base.
 Returns a locale handle for accessing the other methods.  Inherited from
 Locale::Maketext.
 
+=item marktext
```
ehuelsmann: All pre-existing functions document their parameters, meaning that this line should read:

`=item marktext ($string)`

lib/LedgerSMB/Locale.pm
```diff
@@ -85,8 +91,11 @@ package LedgerSMB::Locale;
 
 use strict;
 use warnings;
+use Exporter;
+
+use base qw( Locale::Maketext Exporter );
```
ehuelsmann: Since `Locale::Maketext` didn't need to be `use`-d, I don't think Exporter needs to (or, for consistency: should) either.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -775,27 +775,54 @@ verify_check => md5_hex($check->test_query),
...
-             text => $request->{_locale}->text($check->columns
-                                                ? 'Save and Retry'
-                                                : 'Cancel'),
+            value => 'fix_tests',
+          tooltip => { id => 'action-fix-tests',
+                       msg => $request->{_locale}->text($check->{tooltips}{'Save and Retry'}),
+                       position => 'above'
+                     },
+             text => $request->{_locale}->text('Save and Retry'),
             class => 'submit' },
```
ehuelsmann: There's a stray comma (value separator) here (left over from when this was an array).

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -775,27 +775,54 @@ verify_check => md5_hex($check->test_query),
...
+           { type => 'submit',
+             name => 'action',
+            value => 'cancel',
+          tooltip => { id => 'action-cancel',
+                       msg => $request->{_locale}->text($check->{tooltips}{'Cancel'}),
+                       position => 'above'
+                     },
+             text => $request->{_locale}->text('Cancel'),
+            class => 'submit' }
+    if $buttons{Cancel} or scalar($check->columns) == 0;
```
ehuelsmann: The indenting here is off (should have been the same as `if $check->columns;` above).

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -775,27 +775,54 @@ verify_check => md5_hex($check->test_query),
...
+    push @$buttons,
            { type => 'submit',
              name => 'action',
-            value => $check->columns ? 'fix_tests' : 'cancel',
-             text => $request->{_locale}->text($check->columns
-                                                ? 'Save and Retry'
-                                                : 'Cancel'),
+            value => 'fix_tests',
+          tooltip => { id => 'action-fix-tests',
+                       msg => $request->{_locale}->text($check->{tooltips}{'Save and Retry'}),
```
ehuelsmann: Please write `$check->{tooltips}->{'Save and Retry'}` instead; it's the style we use throughout the code base (note, there are other occurrances of the same below).

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -775,27 +775,54 @@ verify_check => md5_hex($check->test_query),
...
-        format => 'HTML',
+        format => 'HTML'
     );
 
     return $template->render_to_psgi({
            form               => $request,
            base_form          => 'dijit/form/Form',
            heading            => $heading,
-           headers            => [$check->display_name, $check->instructions],
+           headers            => [$request->{_locale}->text($check->display_name),
```
ehuelsmann: any occurrance of `text(<string>)` means the string needs to be translated. The scanner will find here that the string can't be translated due to the fact that it's not a string constant.

Instead of using `text(<variable>, <@parameters>)`, please use `maketext(<variable>, <@parameters>)`; the latter doesn't get marked for translation, but does do the translation.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -838,11 +865,17 @@ sub fix_tests{
     my $table = $check->table;
     my $where = $check->id_where;
     my @edits = @{$check->columns};
+    # If we are inserting and id is displayed, we want to insert
+    # at this exact location
```
ehuelsmann: I can see that  the line below asserts whether we're inserting and the id column is being displayed, but what's unclear to me is what `this exact location` means. Isn't the id displayed so we insert with that exact id-value?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -838,11 +865,17 @@ sub fix_tests{
...
+    my $id_displayed = $check->{insert}
+                and grep( /^$check->{id_column}$/, @{$check->display_cols} );
 
     my $query;
-    if ($request->{insert}) {
-        my $columns = join(', ', map { $dbh->quote_identifier($_) } @edits);
-        my $values = join(', ', map { '?' } @edits);
+    if ($check->{insert}) {
+        my @id = $id_displayed ? $check->{id_column} : ();
```
ehuelsmann: Is the `id_column` an array? Judging by the name of the column, that wouldn't make sense (so we seem to have a mismatch of types here).

But when we have a single value and we *have* the value, why do we need to store in an array?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -838,11 +865,17 @@ sub fix_tests{
...
+        push @id,@edits;
```
ehuelsmann: Do the `@edits` become `id`s?

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: I'm wondering why we put the id in right here. Aren't we receiving the id value from the web? If not, why not?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -855,11 +888,13 @@ sub fix_tests{
     for my $count (1 .. $request->{count}){
         my $id = $request->{"id_$count"};
         my @values;
+        push @values, $id
+            if $check->{insert} and $id_displayed;
```
ehuelsmann: the value of `$check->{insert}` determines the value of `$id_displayed`. No need to check it again.

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: Investigating more: there's no case (currently) where `id` isn't displayed. What would be a suitable `insert` scenario where `id` isn't displayed?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1383,6 +1418,27 @@ sub cancel{
...
+
+    my %test = map { $_->name => $_ } LedgerSMB::Upgrade_Tests->get_tests();
+    my $force_queries = $test{$request->{check}}->{force_queries};
+
+    for my $force_query ( @$force_queries ) {
+        my $dbh = $request->{dbh};
+        $dbh->do($force_query);
+        $dbh->commit;
+    }
+    return upgrade($request);
```
ehuelsmann: Hmm. But what if the strategy is "ok, thanks for notifying me, please proceed, I'm accepting the data-inconsistencies for now (Force)"? Doesn't re-running the upgrade without further measures then simply provide me the same result? Shouldn't we have an indication of upgrade checks that were forced and thus shouldn't be checked again?

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Inserting a new value for which we don’t have a previous key. If we insert _and_ display, we want to make sure that we will match the values of the keys use by referring tables. If we insert and do not display, it should be obvious that we do not have referrals at this moment.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: We ran $force_query, which handled the cases presented. Retrying the upgrade will find more or proceed if everything is ok.

lib/LedgerSMB/Template/HTML.pm
ylavoie: If we only aim for tooltips, I’d rather do that in https://github.com/ylavoie/LedgerSMB/blob/5e00f2c217c3288a6cd7a19cc32c3ddd92de8772/UI/lib/elements.html#L439, otherwise every template will have to look for all elements, searching for tooltips that may need unescaping.

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: I'm sorry, I don't understand the bit which says `If we insert and do not display, it should be obvious that we do not have referrals at this moment.`: why does displaying have anything to do with having references or not? I mean, I can choose not to display the exact reference number, yet to allow data entry for the description of it, right?

lib/LedgerSMB/Template/HTML.pm
ylavoie: And in both options, when will the unescaped string will be processed? The only processing happens before it reaches the Template, so we will only get the format string back, not the processed one.

einhverfr:
Just a note for the future....  It would be helpful to have separate branches for changes that might want to be considered individually.  For example the template changes vs the migration changes.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: We need to address 2 situations in inserting:

1. I need data in this table, please provide some. For this, there is no key to display

2. I have found referals to a missing entry in this table, here's what I got, please complete. For this, I show the referal key

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -173,19 +188,50 @@ has display_name => (is => 'ro', isa => 'Str', required => 1);
...
+    default => sub {
+        return {
+            'Save and Retry' => marktext('Save the fixes provided and attempt to continue migration'),
+            'Cancel' => marktext('Cancel the migration')
+    }},
+    required => 0);
+
+=back
+
+=head1 Methods
```
ehuelsmann: For next time: `head1`s are always in full capitals in our code (and when they're not, that's a bug...)

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -173,19 +188,50 @@ has display_name => (is => 'ro', isa => 'Str', required => 1);
...
+
+=cut
 
 sub _get_tests {
     my ($request) = @_;
 
     my @tests;
-    my $locale = LedgerSMB::App_State::Locale;
```
ehuelsmann: this was actually the only reason the tests array was collected *within* _get_tests; before it was all moved to `$locale`, it was actually outside of `_get_tests`.

PR #3065, opened 2017-08-02
SL30 migration3

ylavoie:
Cosmetic and workflow fixes.

1. Automatically select when only one entry is available in a list
2. Allow `Cancel` when migration fixing is impossible
3. Alternate row background colors, prevent dates to be wrapped and set a minimum width for selects

Protect the code when CSV option import isn't available, to reduce console warnings.

### DIFF
--- a/UI/setup/stylesheet.css
+++ b/UI/setup/stylesheet.css
@@ -60,3 +60,13 @@ img.logo {
 button {
     margin-top: 10px;
 }
+
+.listtop {
+    font-size: larger;
+}
+
+tr.listheading {
+    background-color: lightgrey;
+}
+tr.listrow:nth-child(even) {background: #FFF}
+tr.listrow:nth-child(odd) {background: #EEE}
--- a/UI/setup/stylesheet.css
+++ b/UI/setup/stylesheet.css
@@ -70,3 +70,7 @@ tr.listheading {
 }
 tr.listrow:nth-child(even) {background: #FFF}
 tr.listrow:nth-child(odd) {background: #EEE}
+
+td.dijitStretch {
+    min-width: 125px;
+}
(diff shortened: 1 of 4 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -763,7 +763,7 @@ sub _failed_check {
                    name => $column . "_$id",
                    id => $id,
                    options => $selectable_value,
-                   default_blank => 1,
+                   default_blank => ( 1 != @$selectable_value ) ? 1 : 0
```
ehuelsmann: I'm thinking `( 1 != @$selectable_value )` should be enough here (it's a boolean by itself)

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -784,7 +781,9 @@ sub _failed_check {
            { type => 'submit',
              name => 'action',
             value => 'fix_tests',
-             text => $request->{_locale}->text('Save and Retry'),
+             text => $request->{_locale}->text($check->columns
+                                                ? 'Save and Retry'
+                                                : 'Cancel'),
```
ehuelsmann: Uhm, but that will call `fix_tests`, which currently doesn't know about cancelling migrations... Shouldn't we do something about that?

UI/setup/stylesheet.css
```diff
@@ -70,3 +70,7 @@ tr.listheading {
 }
 tr.listrow:nth-child(even) {background: #FFF}
 tr.listrow:nth-child(odd) {background: #EEE}
+
+td.dijitStretch {
+    min-width: 125px;
```
ehuelsmann: Isn't this just as arbitrary as no width?

PR #3066, opened 2017-08-02
Allow reconciliations on asset, liability and equity accounts only

ylavoie:
- Validate proper categories
- Warn on non-contiguous series
- Add Force and Skip buttons to apply a suitable fix or skip a test completely

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -38,6 +38,7 @@ use LedgerSMB::DBObject::Admin;
 use LedgerSMB::DBObject::User;
 use LedgerSMB::Magic qw( EC_EMPLOYEE HTTP_454 PERL_TIME_EPOCH );
 use LedgerSMB::PSGI::Util;
+use LedgerSMB::Upgrade_Pre_Tests;
 use LedgerSMB::Upgrade_Tests;
 use LedgerSMB::Sysconfig;
 use LedgerSMB::Template::DB;
@@ -721,6 +722,13 @@ sub _upgrade_test_is_applicable {
             && ($test->appname eq $dbinfo->{appname}));
 }
 
+sub _applicable_upgrade_pre_tests {
+    my $dbinfo = shift;
+
+    return grep { _upgrade_test_is_applicable($dbinfo, $_) }
+                  LedgerSMB::Upgrade_Pre_Tests->get_pre_tests;
+}
+
 sub _applicable_upgrade_tests {
     my $dbinfo = shift;
 
@@ -736,6 +744,16 @@ sub upgrade {
 
     $request->{dbh}->{AutoCommit} = 0;
     my $locale = $request->{_locale};
+
+    for my $check (_applicable_upgrade_pre_tests($dbinfo)) {
+        next if defined $request->{"applied_$check->{name}"}
+             && $request->{"applied_$check->{name}"} eq 'On';
+        my $sth = $request->{dbh}->prepare($check->test_query);
+        my $status = $sth->execute()
+            or die 'Failed to execute pre-migration pre-check ' . $check->{name} . ', ' . $sth->errstr;
+        $request->{"applied_$check->{name}"} = 'On';
+        $sth->finish();
+    }
 
     for my $check (_applicable_upgrade_tests($dbinfo)) {
         next if $check->skipable
@@ -899,7 +917,6 @@ sub fix_tests{
         . 'found different test by the same name while fixing data'
         if $request->{verify_check} ne md5_hex($check->test_query);
 
-
     my $table = $check->table;
     my @edits = @{$check->columns};
     # If we are inserting and id is displayed, we want to insert
@@ -931,15 +948,14 @@ sub fix_tests{
         for my $edit (@edits) {
           push @values, $request->{"${edit}_$count"};
         }
-        push @values, map { MIME::Base64::decode($_)} split(/,/,$request->{"id_$count"})
+        push @values, map { $_ ne '' ? MIME::Base64::decode($_) : undef} split(/,/,$request->{"id_$count"})
            if ! $check->{insert};
 
         my $rv = $sth->execute(@values) ||
             $request->error($sth->errstr);
         return LedgerSMB::PSGI::Util::internal_server_error(
-            qq{Upgrade query affected $rv rows, while only a single row
-was expected})
-            if $rv > 1;
+            qq{Upgrade query affected $rv rows, while a single row was expected})
+                if $rv != 1;
     }
     $sth->finish();
     $dbh->commit;
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -38,7 +38,7 @@ use LedgerSMB::DBObject::Admin;
 use LedgerSMB::DBObject::User;
 use LedgerSMB::Magic qw( EC_EMPLOYEE HTTP_454 PERL_TIME_EPOCH );
 use LedgerSMB::PSGI::Util;
-use LedgerSMB::Upgrade_Pre_Tests;
+use LedgerSMB::Upgrade_Preparation;
 use LedgerSMB::Upgrade_Tests;
 use LedgerSMB::Sysconfig;
 use LedgerSMB::Template::DB;
@@ -722,11 +722,11 @@ sub _upgrade_test_is_applicable {
             && ($test->appname eq $dbinfo->{appname}));
 }
 
-sub _applicable_upgrade_pre_tests {
+sub _applicable_upgrade_preparations {
     my $dbinfo = shift;
 
     return grep { _upgrade_test_is_applicable($dbinfo, $_) }
-                  LedgerSMB::Upgrade_Pre_Tests->get_pre_tests;
+                  LedgerSMB::Upgrade_preparations->get_preparations;
 }
 
 sub _applicable_upgrade_tests {
@@ -745,13 +745,13 @@ sub upgrade {
     $request->{dbh}->{AutoCommit} = 0;
     my $locale = $request->{_locale};
 
-    for my $check (_applicable_upgrade_pre_tests($dbinfo)) {
-        next if defined $request->{"applied_$check->{name}"}
-             && $request->{"applied_$check->{name}"} eq 'On';
-        my $sth = $request->{dbh}->prepare($check->test_query);
+    for my $preparation (_applicable_upgrade_preparations($dbinfo)) {
+        next if defined $request->{"applied_$preparation->{name}"}
+             && $request->{"applied_$preparation->{name}"} eq 'On';
+        my $sth = $request->{dbh}->prepare($preparation->preparation);
         my $status = $sth->execute()
-            or die 'Failed to execute pre-migration pre-check ' . $check->{name} . ', ' . $sth->errstr;
-        $request->{"applied_$check->{name}"} = 'On';
+            or die 'Failed to execute migration preparation ' . $preparation->{name} . ', ' . $sth->errstr;
+        $request->{"applied_$preparation->{name}"} = 'On';
         $sth->finish();
     }
 
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -726,7 +726,7 @@ sub _applicable_upgrade_preparations {
     my $dbinfo = shift;
 
     return grep { _upgrade_test_is_applicable($dbinfo, $_) }
-                  LedgerSMB::Upgrade_preparations->get_preparations;
+                  LedgerSMB::Upgrade_Preparation->get_migration_preparations;
 }
 
 sub _applicable_upgrade_tests {
--- a/lib/LedgerSMB/Upgrade_Pre_Tests.pm
+++ b/lib/LedgerSMB/Upgrade_Pre_Tests.pm
@@ -1,0 +1,116 @@
+=head1 NAME
+
+LedgerSMB::Upgrade_Pre_Tests - Upgrade pre-tests for LedgerSMB
+
+=head1 SYNPOPSIS
+
+ TODO
+
+=head1 DESCRIPTION
+
+This module has a single function that returns upgrade pre-tests.
+
+=cut
+
+package LedgerSMB::Upgrade_Pre_Tests;
+use strict;
+use warnings;
+use Moose;
+use Moose::Util::TypeConstraints;
+use namespace::autoclean;
+
+=head1 FUNCTIONS
+
+=over
+
+=item get_pre_tests()
+
+Returns the pre-test array
+Pre-tests are run only once before any tests, to adjust some tables for
+data uniqueness to allow edit, for example.
+
+=cut
+
+sub get_pre_tests {
+    my ($self) = @_;
+    my @pre_tests = $self->_get_pre_tests;
+    return @pre_tests;
+}
+
+=back
+
+=head1 TEST DEFINITION
+
+Each test is a Moose object with the following properties (optional ones marked
+as such).
+
+=over
+
+=item name
+
+Name of the test
+
+=cut
+
+has name => (is => 'ro', isa => 'Str', required => 1);
+
+=item min_version
+
+The first version to run this against
+
+=cut
+
+has min_version => (is => 'ro', isa => 'Str', required => 1);
+
+=item max_version
+
+The maximum version to run this against
+
+=cut
+
+has max_version => (is => 'ro', isa => 'Str', required => 1);
+
+=item appname
+
+The appname of the application the test belongs to.
+Can currently be 'ledgersmb' or 'sql-leder'.
+
+=cut
+
+has appname => (is => 'ro', isa => 'Str', required => 1);
+
+=item test_query
+
+Text of the query to run
+
+=cut
+
+has test_query => (is => 'ro', isa => 'Str', required => 1);
+
+=back
+
+=head1 Methods
+
+=cut
+
+sub _get_pre_tests {
+    my ($request) = @_;
+
+    my @pre_tests;
+
+    push @pre_tests, __PACKAGE__->new(
+        # Add a unique key to allow editing
+        test_query => 'ALTER TABLE acc_trans DROP COLUMN IF EXISTS lsmb_entry_id;
+                       ALTER TABLE acc_trans add column lsmb_entry_id SERIAL UNIQUE;',
+        name => 'add_unique_acc_trans_key',
+           appname => 'sql-ledger',
+       min_version => '2.7',
+       max_version => '3.0'
+    );
+
+    return @pre_tests;
+}
+
+__PACKAGE__->meta->make_immutable;
+
+1;
--- a/lib/LedgerSMB/Upgrade_Tests.pm
+++ b/lib/LedgerSMB/Upgrade_Tests.pm
@@ -1,6 +1,6 @@
 =head1 NAME
 
-LedgerSMB::Upgrade_Tests - Upgrade tests for LedgerSMB
+LedgerSMB::Upgrade_Pre_Tests - Upgrade pre-tests for LedgerSMB
 
 =head1 SYNPOPSIS
 
@@ -19,7 +19,6 @@ use Moose;
 use Moose::Util::TypeConstraints;
 use namespace::autoclean;
 use List::Util qw( first );
-
 use LedgerSMB::Locale qw(marktext);
 
 =head1 FUNCTIONS
@@ -444,40 +443,6 @@ push @tests, __PACKAGE__->new(
   min_version => '1.3',
   max_version => '1.4'
 );
-
-#=pod
-
-# This must be done before the acc_trans test, for they have duplicates
-# and need a unique key
-    push @tests, __PACKAGE__->new(
-        # Add a unique key to allow editing
-        # Make the test serially reusable and protect against triggers
-        test_query => 'ALTER TABLE acc_trans DISABLE TRIGGER USER;
-                       ALTER TABLE acc_trans DROP COLUMN IF EXISTS i_key;
-                      CREATE TABLE acc_trans1 (LIKE acc_trans INCLUDING ALL);
-                       ALTER TABLE acc_trans1 DISABLE TRIGGER USER;
-                       ALTER TABLE acc_trans1 add column i_key SERIAL UNIQUE;
-                      INSERT INTO acc_trans1 SELECT * FROM acc_trans;
-                        DROP TABLE acc_trans;
-                       ALTER TABLE acc_trans1 RENAME TO acc_trans;
-                       ALTER TABLE acc_trans ENABLE TRIGGER USER;
-                      SELECT trans_id
-                        FROM acc_trans
-                       WHERE i_key IS NULL',
-      display_name => marktext('Transactions unique key'),
-              name => 'add_unique_acc_trans_key',
-      display_cols => ['trans_id'],
-           columns => ['trans_id'],
-        id_columns => ['i_key'],
-          id_where => 'i_key IS NULL AND ',
-      instructions => marktext(
-                       'This should never show. We added a unique key to acc_trans'),
-           buttons => ['Save and Retry', 'Cancel'],
-             table => 'acc_trans',
-           appname => 'sql-ledger',
-       min_version => '2.7',
-       max_version => '3.0'
-    );
 
 push @tests, __PACKAGE__->new(
    test_query => q{select distinct gifi_accno from chart
@@ -1223,9 +1188,7 @@ push @tests, __PACKAGE__->new(
 );
 
 push @tests, __PACKAGE__->new(
-    # Add a unique key to allow editing
-    # Make the test serially reusable and protect against triggers
-    test_query => q(SELECT ac.i_key, ac.trans_id, ac.id, ac.memo, ac.amount, xx.description,
+    test_query => q(SELECT ac.lsmb_entry_id, ac.trans_id, ac.id, ac.chart_id, ac.memo, ac.amount, xx.description,
                           ch.description as account, ch.accno, ch.link, ch.charttype, ch.category, ac.cleared, approved
                      FROM acc_trans ac
                      JOIN (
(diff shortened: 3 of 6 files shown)


ehuelsmann:
Do we reject the input if it's not reconciled on a bank account? If so, this belongs in the migrations, if not, it probably belongs elsewhere (like: "sanity checks"; a category we don't have yet)

ylavoie:
This belongs in sanity checks and we should create such a category. We do not reject the reconciliation currently, even if it has no functional value or of only some transactions a reconciled in an expense or cogs account.

einhverfr:
People do reconcile non-bank accounts, whether petty cash drawers or AR/AP accounts.

einhverfr:
Also, I can imagine that one might have reserve payment accounts that might not be flagged for the payment dropdown.

I don't think we should tie reconciliation to what dropdown it is currently a member of.  I think that's tying the wrong concerns together.

einhverfr:
It would make more sense to me if the check was (at least for now) on the category field, and we restricted reconciliations to asset, liability, and equity accounts.  

I am having trouble imagining why one would reconcile an income or expense account, though.

einhverfr:
And that seems to address the original concern (reconciliation of cogs accounts for example)

ylavoie:
- We should tie it to the proper categories, there is no value to reconcile expenses or cogs.
- With the new possibility for the user to force a migration even if the tests flagged cases, we should reports cases when we see non-continuous series of reconciliations or usage on non-conventional accounts and let the user decide.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -693,7 +693,7 @@ sub upgrade {
     for my $check (_applicable_upgrade_tests($dbinfo)) {
         my $sth = $request->{dbh}->prepare($check->test_query);
         $sth->execute()
-            or die "Failed to execute pre-migration check " . $check->name;
+            or die "Failed to execute pre-migration check " . $check->name . ", " . $sth->errstr;
```
ylavoie: This explains more in detail what is happening. I think it is justified for developers and migrators.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -740,67 +740,57 @@ verify_check => md5_hex($check->test_query),
 
     my $rows = [];
     while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
-      my $id = $row->{$check->{id_column}};
+      my $count = 1+scalar(@$rows);
+      #TODO: A '' will fail with invalid input syntax in fix_tests - YL
```
ylavoie: When dealing with multiple keys, we may find cases when not all are present in user record. We should deal with that.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -853,41 +843,46 @@ sub fix_tests{
...
+        $query = "UPDATE $table SET $setters WHERE $where "
+               . join(' AND ',map {"$_ = ? "} @{$check->id_columns});
     }
     my $sth = $dbh->prepare($query);
 
     for my $count (1 .. $request->{count}){
-        my $id = $request->{"id_$count"};
         my @values;
-        push @values, $id
+        push @values, split(',',@{$check->id_columns})
```
ylavoie: Perl split is prohibited. In view of the fact that those strings has to go through the hiddens, we need to find a better way.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -853,41 +843,46 @@ sub fix_tests{
...
         }
-        push @values, $request->{"id_$count"}
+        push @values, split(',',$request->{"id_$count"})
            if ! $check->{insert};
 
-        $sth->execute(@values) ||
+        my $rv = $sth->execute(@values) ||
             $request->error($sth->errstr);
+        #TODO: We should abort gracefully
+        die "$rv rows affected instead of only 1"
```
ylavoie: That ain't nice.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Fixed, ProhibitStringySplit prohibits strings as first argument on split and mandates using a regex

einhverfr:
If we want to tie to proper categories, then it seems to me the correct category is to allow reconciliation on asset, liability, and equity accounts but deny them on income and expense accounts.  The problem with cogs is the checkbox really only affects dropdown membership and isn't a permanent category.  Moreover you *can* select cogs and ar_paid if you like.....

UI/lib/elements.html
```diff
@@ -432,7 +432,7 @@
 
 <?lsmb BLOCK tooltip;
   IF element_data && element_data.id.defined() && element_data.msg.defined();
-     props="connectId:'" _ element_data.id _ "'";
+     props="connectId:'" _ element_data.id.replace('_','-') _ "'";
```
ehuelsmann: this doesn't make sense. Why would an id have an underscore that needs replacement with a hyphen?

It feels very much like this is papering over another problem: If the 'id' is explicitly set, we definitely shouldn't be changing it.

ehuelsmann:
9cdd1e4de8a7b395d725770c10b7c855d6356323 can and should be a PR of its own: It has nothing to do with reconciliations.

ehuelsmann:
2857e09ddb63ce1fb82b73c7bca6a93b14648457 fixes a problem coming from a PR not related to reconciliations. Please PR separately.

ehuelsmann:
c0de833a77cfacffe14738055cfa1befe7ee593d too is the follow-up to another PR  unrelated to reconciliations. Please PR separately too.

ehuelsmann:
733f6bb554f27ee7ac716fb3718cde25911f11fb and 29c64aa8dfea3534c311487d4b54c4b5dfbc89f1 seem like they make a net-zero effect. Could you drop both from the PR? it would seem they make the history unnecessarily lengthy.

ehuelsmann:
I think 036374adcd0de2805a54958a5549e681a6799267 makes sense, but what's the "multi-key" bit there?

UI/lib/elements.html
ylavoie: I'm not changing the id of the main item, Dojo take cares of normalizing underscores to hyphens, so we need to point the tooltip to the normalized widgetid.

ehuelsmann: It does? I know LedgerSMB "normalizes" items with a "name=" with an underscore to have an "id=" with a hyphen. Isn't your problem caused by the underscore-to-hyphen-in-LedgerSMB-html-generator code instead of by Dojo?

einhverfr:
As a general note, I really think we should enforce one PR per topic.  And that whitespace changes (blank line deletions etc) should be their own PR.  In this case I see at least 4 major PR-worthy topics (blank line deletions, changing behaviour on upgrade tests, SQL-Ledger upgrades, and the check on past reconciliations).  This makes review and discussion a whole lot harder.  PRs should be atomic.

I would be for letting it slide on this case only but let's keep things simple going forward.

UI/lib/elements.html
ylavoie: The problem wasn't Dojo but us, trying to get nicer CSS in our Template elements.

ehuelsmann: Can you be more verbose? This remark is too coarse. I have no idea what you're referring to.

ylavoie: I added a comment in elements.html where we "normalize" id but affect tooltips, for which we compensate in tooltip generation.

ehuelsmann: Actually, no, we don't. The code you commented doesn't apply to all ids so we should not compensate all cases: inputs which themselves provide an explicit "id" value don't get this transcoding (only those ids which are derived from "name" attributes)

ylavoie: Fixed, we now normalize the button and the related tooltip at the same point. I removed the Tooltip patch.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -417,6 +461,98 @@ selectable_values => { chart_id => q{SELECT concat(accno,' -- ',description) AS
...
+);
+
+#=pod
+
+# This must be done before the acc_trans test, for they have duplicates
+# and need a unique key
+    push @tests, __PACKAGE__->new(
+        # Add a unique key to allow editing
+        # Make the test serially reusable and protect against triggers
+        test_query => "ALTER TABLE acc_trans DISABLE TRIGGER USER;
```
ehuelsmann: You are aware that checks like these may be run multiple times during migration? I'm wondering if this is the best way to deal with this situation. Isn't the key ['trans_id', 'entry_id'] unique?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -1133,6 +1269,50 @@ push @tests, __PACKAGE__->new(
...
+                             UNION SELECT a.id, n.name        FROM ar a JOIN customer n ON n.id = a.customer_id
+                             UNION SELECT a.id, n.name        FROM ap a JOIN vendor n   ON n.id = a.vendor_id
+                         ) xx ON xx.id = ac.trans_id
+                         JOIN chart ch ON (ac.chart_id = ch.id)
+                        WHERE ( ch.category NOT IN ( 'A', 'L', 'Q' ) 
+                             OR ch.link NOT LIKE '%paid' ) 
+                          AND ac.cleared IS NOT NULL 
+                          AND ac.approved
+                     ORDER BY accno, transdate, ac.id",
+      display_name => marktext('Reconciliations on non-bank accounts'),
```
ehuelsmann: While I agree that this could be a sensible data integrity check, I'm wondering: does this situation prevent running an acceptable migration?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -233,6 +262,21 @@ sub _get_tests {
...
 
+push @tests, __PACKAGE__->new(
+    test_query => "SELECT * FROM pg_settings
+                    WHERE name = 'client_encoding'
+                      AND setting <> 'UTF8'",
+    display_name => marktext('Client encoding must be UTF8'),
+    name => 'client_encoding',
+    display_cols => ['name','setting','unit','category','short_desc','extra_desc','context','vartype','source','min_val','max_val','enumvals','boot_val','reset_val','sourcefile','sourceline'],
+ instructions => marktext(
+                   'Please fix the client encoding'),
```
ehuelsmann: Doesn't the client negotiate its own encoding? I mean, DBD::Pg >3.3.0 will always want to open a UTF8 encoded connection. If we require that, then why does the client encoding get set incorrectly?

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Entry_id won't exists until migration is started. Tests are done before that.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Nope, the user can force the migration even with those still present.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: What does this have to do with the current PR?

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Or skip this test completely

ehuelsmann:
This PR includes the "skip" button from #3351. Merging should wait until that's merged; it'll allow better review of the actual change here.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -725,6 +725,8 @@ sub _applicable_upgrade_tests {
                   LedgerSMB::Upgrade_Tests->get_tests;
 }
 
+my @checks;
```
ehuelsmann: This line makes me think this PR does not yet include the insights from #3351.

lib/LedgerSMB/Upgrade_Tests.pm
ehuelsmann: Indeed: what does this check have to do with the current PR? I think it does not relate to reconciliation aft all. Please remove this extra check from this PR. We might want to move it to the sanity checks later.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -415,6 +472,94 @@ database, or delete the offending rows through PgAdmin III or psql'),
...
+);
+
+#=pod
+
+# This must be done before the acc_trans test, for they have duplicates
+# and need a unique key
+    push @tests, __PACKAGE__->new(
+        # Add a unique key to allow editing
+        # Make the test serially reusable and protect against triggers
+        test_query => 'ALTER TABLE acc_trans DISABLE TRIGGER USER;
```
ehuelsmann: While I understand the problem, I'm not too hot about this solution: it abuses the validation system to execute migration code. 

Can you indicate what happens when this code is run twice?

old/lib/LedgerSMB/DBObject/Reconciliation.pm
```diff
@@ -386,7 +386,10 @@ sub get {
     $self->{account_info} = $ref;
     ($ref) = $self->call_dbmethod(funcname=>'reconciliation__get_cleared_balance',
                                 args => { chart_id => $ref->{id},
-                                          report_date => $self->{end_date}->clone->add_interval('month',-1) });
+                                          report_date => $self->{end_date}->clone
```
ehuelsmann: This has been merged yesterday (don't render the PR# of hand)

sql/modules/Reconciliation.sql
```diff
@@ -30,11 +30,17 @@ BEGIN;
...
 $$
+DECLARE
+        result NUMERIC;
 BEGIN
-        UPDATE cr_report set submitted = true where id = in_report_id;
-        PERFORM reconciliation__save_set(in_report_id, in_line_ids);
-
-        RETURN FOUND;
+        result = reconciliation__check_balanced(in_report_id);
+        IF result = 0 THEN
```
ehuelsmann: Why are you checking balancedness *before* adding the new line ids to the report?

sql/modules/Reconciliation.sql
```diff
@@ -93,6 +115,30 @@ $$Sets which lines of the report are cleared.$$;
 CREATE OR REPLACE FUNCTION reconciliation__delete_my_report(in_report_id int)
 RETURNS BOOL AS
 $$
+DECLARE
+    failed BOOL;
+BEGIN
+    SELECT approved INTO failed FROM cr_report
```
ehuelsmann: No need for the `failed` variable here (which is unused in the result):

`PERFORM * FROM cr_report ...`

sql/modules/Reconciliation.sql
```diff
@@ -93,6 +115,30 @@ $$Sets which lines of the report are cleared.$$;
...
+    WHERE id = in_report_id AND approved;
+    IF FOUND THEN
+        RAISE EXCEPTION 'Cannot delete approved recon report';
+        RETURN FALSE;
+    END IF;
+
+    -- Make sure that transactions present on this report do not have a
+    -- cleared_on date or they won't ever be accessible anymore.
```
ehuelsmann: If "posting" the report sets the cleared date, and posted reports can't be deleted, how come we may fail this test?

sql/modules/Reconciliation.sql
```diff
@@ -93,6 +115,30 @@ $$Sets which lines of the report are cleared.$$;
...
+    SELECT COUNT(*)>0 INTO failed
+    FROM cr_report_line rl
+    JOIN acc_trans ac ON rl.ledger_id = ac.entry_id
+    WHERE report_id = in_report_id
+      AND cleared_on IS NOT NULL;
+
+    IF failed THEN
+        RAISE EXCEPTION 'Transactions already cleared on this report';
```
ehuelsmann: But the report itself isn't POSTED?

sql/modules/Reconciliation.sql
```diff
@@ -101,9 +147,10 @@ $$
                                     and approved IS NOT TRUE);
     DELETE FROM cr_report
      WHERE id = in_report_id AND entered_username = SESSION_USER
-           AND submitted IS NOT TRUE AND approved IS NOT TRUE
-    RETURNING TRUE;
-$$ LANGUAGE SQL SECURITY DEFINER;
+           AND submitted IS NOT TRUE AND approved IS NOT TRUE;
+    RETURN FOUND;
+END;
+$$ LANGUAGE PLPGSQL SECURITY DEFINER;
```
ehuelsmann: Why do you want to change the procedure to PLPGSQL? The two conditions I see can't occur, or can be integrated.

sql/modules/Reconciliation.sql
```diff
@@ -166,20 +213,28 @@ disable the block_change_when_approved trigger on cr_report.$$;
...
+BEGIN
+    SELECT COALESCE(
+        (SELECT SUM(ac.amount) * CASE WHEN c.category in('A', 'E') THEN -1 ELSE 1 END
+            INTO result
+            FROM account c
+            JOIN acc_trans ac ON (ac.chart_id = c.id)
+        JOIN (      SELECT id FROM ar WHERE approved
+              UNION SELECT id FROM ap WHERE approved
+              UNION SELECT id FROM gl WHERE approved
+              ) g ON g.id = ac.trans_id
```
ehuelsmann: Can be replaced by a single SELECT on the `transaction` table.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -426,6 +426,60 @@ database, or delete the offending rows through PgAdmin III or psql'),
...
+ instructions => marktext('Make sure all meta numbers are unique.'),
+      appname => 'ledgersmb',
+  min_version => '1.3',
+  max_version => '1.4'
+);
+
+#=pod
+
+# This must be done before the acc_trans test, for they have duplicates
+# and need a unique key
```
ehuelsmann: actually, this will *always* run before the acc_trans test, even if the acc_trans test's UI has already been shown and "corrective information" has been provided.

There's one problem here though: because SQL doesn't guarantee set-ordering, the code below doesn't necessarily generate the same unique values for the same rows as the ones already in the "uniquified" set (and thus in the corrective data).

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -1188,6 +1242,51 @@ push @tests, __PACKAGE__->new(
   max_version => '3.0'
 );
 
+    push @tests, __PACKAGE__->new(
+        # Add a unique key to allow editing
+        # Make the test serially reusable and protect against triggers
```
ehuelsmann: This comment doesn't apply?

sql/modules/Reconciliation.sql
ylavoie: Fixed

sql/modules/Reconciliation.sql
ylavoie: Agreed, so we should never have cleared transactions in it.

sql/modules/Reconciliation.sql
ylavoie: FOUND requires PLPGSQL

sql/modules/Reconciliation.sql
ylavoie: Fixed

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Fixed

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Fixed, moved into pre-tests.

sql/modules/Reconciliation.sql
ylavoie: The entries are already added and a submit is asked for an unbalanced report.

sql/modules/Reconciliation.sql
ylavoie: Submit 2 reports, Refuse the first, Fix it by adding a transaction that was in the 2nd and you have an approved transaction in 2 reports and you can't remove it from the 2nd.
Batch submits in migration make ithat situation easy to arise.

sql/modules/Reconciliation.sql
```diff
@@ -30,11 +30,17 @@ BEGIN;
...
+        result NUMERIC;
 BEGIN
-        UPDATE cr_report set submitted = true where id = in_report_id;
-        PERFORM reconciliation__save_set(in_report_id, in_line_ids);
-
-        RETURN FOUND;
+        result = reconciliation__check_balanced(in_report_id);
+        IF result = 0 THEN
+            UPDATE cr_report set submitted = true where id = in_report_id;
+            PERFORM reconciliation__save_set(in_report_id, in_line_ids);
```
ehuelsmann: If they have already been added, then what does the line above do?

lib/LedgerSMB/Upgrade_Pre_Tests.pm
```diff
@@ -0,0 +1,116 @@
...
+=cut
+
+sub _get_pre_tests {
+    my ($request) = @_;
+
+    my @pre_tests;
+
+    push @pre_tests, __PACKAGE__->new(
+        # Add a unique key to allow editing
+        test_query => 'ALTER TABLE acc_trans DROP COLUMN IF EXISTS lsmb_entry_id;
```
ehuelsmann: Why is the module called "pre tests" ? I don't see a test in this row nor in the one below.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -1,6 +1,6 @@
 =head1 NAME
 
-LedgerSMB::Upgrade_Tests - Upgrade tests for LedgerSMB
+LedgerSMB::Upgrade_Pre_Tests - Upgrade pre-tests for LedgerSMB
```
ehuelsmann: You renamed this module, but its file is still called Upgrade_Tests.

sql/modules/Reconciliation.sql
ehuelsmann: Let me rephrase: what guarantees are there that the `in_line_ids` is exactly the same set of records as there is currently already in the report?

ehuelsmann: In other words: Is there any guarantee we're marking as reconciled, the same lines as the ones you used to check that the report is balanced?

sql/modules/Reconciliation.sql
ehuelsmann: So does `RAISE EXCEPTION`; my point is: do we really need to raise an exception? Or simply add conditions to the DELETE statements which make them "update nothing when the preconditions are not met"?

The function used to be defined to return FALSE on failure. You've made it raise an exception on failure. That's a major change in paradigm. Is that required, or can we do with a smaller change?

sql/modules/Reconciliation.sql
```diff
@@ -166,20 +211,26 @@ disable the block_change_when_approved trigger on cr_report.$$;
...
+            FROM account c
+            JOIN acc_trans ac ON (ac.chart_id = c.id)
+            JOIN transactions t on t.id = ac.trans_id
+        WHERE c.id = $1 AND cleared
+          AND t.approved IS TRUE
+          AND ac.approved IS true
+          AND ac.transdate <= in_report_date
+        GROUP BY c.id, c.category)
+    ,0);
+    RETURN result;
```
ehuelsmann: why do we need to convert this query/stored procedure to plpgsql? We're not doing anything specific that requires collecting the result into `result`.

sql/modules/Reconciliation.sql
```diff
@@ -166,20 +211,26 @@ disable the block_change_when_approved trigger on cr_report.$$;
...
-          ) g ON g.id = ac.trans_id
-    WHERE c.id = $1 AND cleared
-      AND ac.approved IS true
-      AND ac.transdate <= in_report_date
-    GROUP BY c.id, c.category;
-$$ LANGUAGE sql;
+DECLARE
+        result NUMERIC;
+BEGIN
+    SELECT COALESCE(
```
ehuelsmann: This outer query is ineffective. The function result isn't being coalesced as it isn't stored in the `result` variable.

sql/modules/Reconciliation.sql
ylavoie: Fixed

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -1188,6 +1187,48 @@ push @tests, __PACKAGE__->new(
...
+                 ORDER BY trans_id, ac.id, accno, transdate),
+  display_name => marktext('Unneeded Reconciliations'),
+          name => 'reconciliation_on_unrelated_accounts',
+  display_cols => ['trans_id', 'id', 'memo', 'amount', 'description',
+                   'accno', 'account', 'link', 'category', 'cleared', 'approved'],
+       columns => ['cleared'],
+    id_columns => ['lsmb_entry_id'],
+  instructions => marktext(
+                   'Reconciliations should be on asset, liability or equity accounts only.<br>
+Please review the dates and categories in the original application if needed'),
```
ehuelsmann: why "if needed"? Is this a request to do so (and thus a validation failure) or not?

ylavoie: They may be able to do all in LSMB or go back to SL to fix it.

lib/LedgerSMB/Upgrade_Tests.pm
ehuelsmann: No, I mean: You're changing the `=head1 NAME` section on  the wrong file.

lib/LedgerSMB/Upgrade_Tests.pm
ehuelsmann: maybe instead of "if needed" -> "or click 'Skip' to continue" ?

sql/modules/Reconciliation.sql
```diff
@@ -44,6 +50,22 @@ $$Submits a reconciliation report for approval.
...
+BEGIN
+      SELECT SUM(crl.our_balance) - cr.their_total INTO result
+        FROM cr_report_line crl
+        JOIN cr_report cr ON cr.id = crl.report_id
+       WHERE cr.id = in_report_id
+    GROUP BY cr.their_total;
+
+      RETURN COALESCE(result,0);
+END;
+$$ LANGUAGE PLPGSQL;
```
ehuelsmann: Same here. This can't be a `LANGUAGE SQL` function?

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: I don't want to promote Skip. It is fixable and should be.

PR #3068, opened 2017-08-02
Bring Development config in line with standard

ylavoie:
tools/starman-development.psgi should stay close to tools/starman.psgi and benefit from the same improvements


ehuelsmann:
It would be better if we could leverage the already existing infrastructure in the standard config, *adding to* that in the development psgi file.

PR #3075, opened 2017-08-07
Reduce Moose error verbosity

ylavoie:


ehuelsmann:
I'm sorry, this is a local solution to a (mostly) one-off problem. If nothing else, it'll result in confusing behaviour (because not consistent). It *is* something to keep in the list of snippets which developers might use to do their debugging, though. Maybe something to add to our community/developers guide on ledgersmb.org?

ylavoie:
This was presented as solution for Upgrade_Tests, it would be easy to make it global. It still feel that dumping a stacktrace on user screen isn’t the way to go.

PR #3084, opened 2017-08-11
Add translation support for decorated text

### LINKED ISSUE #3080: Add decorations in instructions
In order to allow emphasis on migration instructions, bold, italic, underline, line breaks... should be allowed.
Currently, all text is escaped, so any decoration will show up as text on the screen instead of being processed.
The easiest method seems to define a new BLOCK, similar to the one for tooltips, and allow unescaping. Although being not elegant, it causes no harm and fits the need for now.

ylavoie:
Use minimal unescaping and add translation for tooltips and migration instructions. Fixes #3080

### DIFF
--- a/UI/lib/elements.html
+++ b/UI/lib/elements.html
@@ -424,6 +424,12 @@
 
 <?lsmb BLOCK CLOSE_STATUS_DIV ?></div><?lsmb END ?>
 
+<?lsmb BLOCK decorated_text;
+  IF element_data && element_data.msg.defined();
+    UNESCAPE(element_data.msg);
+  END;
+END ?>
+
 <?lsmb BLOCK tooltip;
   IF element_data && element_data.id.defined() && element_data.msg.defined();
      props="connectId:'" _ element_data.id _ "'";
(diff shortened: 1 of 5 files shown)


lib/LedgerSMB/Template.pm
```diff
@@ -568,6 +568,7 @@ sub _render {
     $cleanvars->{UNESCAPE} = sub { return $unescape->(@_); }
         if ($unescape && !$self->{no_escape});
     $cleanvars->{text} = sub { return $self->_maketext($escape, @_); };
+    $cleanvars->{decorated_text} = sub { return $self->{locale}->maketext(@_); };
```
ehuelsmann: You're introducing a function `decorated_text()` which is a misnomer (it doesn't do anything about decoration: it translates), but it's also not necessary (see my comment above).

UI/lib/elements.html
```diff
@@ -424,6 +424,12 @@
 
 <?lsmb BLOCK CLOSE_STATUS_DIV ?></div><?lsmb END ?>
 
+<?lsmb BLOCK decorated_text;
+  IF element_data && element_data.msg.defined();
+    decorated_text(element_data.msg.replace('&lt;((/?[ibup]|br))&gt;','<$1>'));
```
ehuelsmann: What helps the reader to understand that "decorated_text()" means that the text is being translated?

Also, you're translating here, but you already did that here: https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Scripts/setup.pm#L776 ; so this function (decorated_text(), that is) is doing double work.

Given that all variable/dynamic values already have been translated when going into the template, I'd prefer the approach in https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Scripts/setup.pm#L776 rather than the one introduced here.

PR #3085, opened 2017-08-11
Master fix migration target (closes #2658)

### LINKED ISSUE #2658: 1.3->1.5 migrations: Need a fixed "end-point" schema to run the migration "to"
With a single data migration script *from* 1.3, we do not only need a single point of departure (latest 1.3 schema), we also need a single point of arrival (point to migrate *to*). From that point, the standard `sql/changes/` scripts will handle the migration as they do for "everybody else".

ehuelsmann:


ehuelsmann:
Closing for now.

ehuelsmann:
Reopening to force re-alignment with current master.

ehuelsmann:
Merging, but note that I have BDD tests in progress which are going to follow as soon as I get my local setup to include PhantomJS (or some other selenium server).

PR #3088, opened 2017-08-13
System information added to setup.pl

ehuelsmann:

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -37,6 +37,7 @@ requires 'PGObject::Type::DateTime', '1.0.4';
 requires 'PGObject::Type::ByteString', '1.1.1';
 requires 'PGObject::Util::DBMethod';
 requires 'PGObject::Util::DBAdmin', '0.09';
+requires 'Plack';
 requires 'Plack::App::File';
 requires 'Plack::Builder';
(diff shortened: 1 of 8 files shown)


cpanfile
```diff
@@ -37,6 +37,7 @@ requires 'PGObject::Type::DateTime', '1.0.4';
 requires 'PGObject::Type::ByteString', '1.1.1';
 requires 'PGObject::Util::DBMethod';
 requires 'PGObject::Util::DBAdmin', '0.09';
+requires 'Plack';
```
ylavoie: Pulling Plack as a whole brings many things, Request & Response, for example, that we felt unneccessary before. Should we be more precise?

ehuelsmann: You're correct, if the requires line would apply to the _dist_ Plack; however, the line declares a dependency on the _module_ Plack. Looking at the module sources, there's nothing more than a version declaration, meaning that just `use`-ing the module won't pull more modules into memory. The _dist_ Plack contains (next to the module Plack) indeed Plack::Request and Plack::Response, but we already depended on the dist Plack, because we use `Plack::Builder`, which is part of the same dist.

As such, the list of dependencies hasn't changed with the above line.

PR #3093, opened 2017-08-18
Cleanup UI header use and HTML tests

ehuelsmann:


UI/Reports/filters/inventory_adj.html
```diff
@@ -1,5 +1,4 @@
 <?lsmb
```
ylavoie: Every html file but this one has the process aligned on the same line as `lsmb`. Should we stay coherent on style?

PR #3101, opened 2017-08-23
Use Perl Pattern Recognizer (PPR) to extract strings for localization

ylavoie:
With the advent of multilines in tooltips and migration instructions, there was a need to support all forms of Perl strings. This proved difficult with the current state machine and developing the regexes to properly support `q`, `qq` and `heredocs` were redundant in regard of PPR ability to handle those.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -115,6 +115,7 @@ on 'develop' => sub {
     requires 'Perl::Critic::Moose';
     requires 'Pherkin::Extension::Weasel', '0.02';
     requires 'Plack::Middleware::Pod'; # YLA - Generate browseable documentation
+    requires 'PPR';
     requires 'Selenium::Remote::Driver';
     requires 'TAP::Parser::SourceHandler::pgTAP';
     requires 'Test::BDD::Cucumber', '0.52';
--- a/cpanfile
+++ b/cpanfile
@@ -104,12 +104,15 @@ feature 'debug', "Debug pane" =>
 on 'develop' => sub {
     requires 'App::Prove', '3.36';
     requires 'File::Util';
+    requires 'Getopt::Long';
     requires 'HTML::Lint';
     requires 'HTML::Lint::Parser', '2.26';
     requires 'HTML::Lint::Pluggable';
     requires 'HTML::Lint::Pluggable::HTML5';
     requires 'HTML::Lint::Pluggable::WhiteList';
     recommends 'Linux::Inotify2';
+    requires 'List::Util';
+    requires 'Locale::Maketext::Simple';
     requires 'Module::CPANfile'; # for 01.2-deps.t
     requires 'Perl::Critic';
     requires 'Perl::Critic::Moose';
--- a/cpanfile
+++ b/cpanfile
@@ -104,14 +104,12 @@ feature 'debug', "Debug pane" =>
 on 'develop' => sub {
     requires 'App::Prove', '3.36';
     requires 'File::Util';
-    requires 'Getopt::Long';
     requires 'HTML::Lint';
     requires 'HTML::Lint::Parser', '2.26';
     requires 'HTML::Lint::Pluggable';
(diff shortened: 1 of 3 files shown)


cpanfile
```diff
@@ -104,17 +104,21 @@ feature 'debug', "Debug pane" =>
...
     requires 'App::Prove', '3.36';
     requires 'File::Util';
+    requires 'Getopt::Long';
     requires 'HTML::Lint';
     requires 'HTML::Lint::Parser', '2.26';
     requires 'HTML::Lint::Pluggable';
     requires 'HTML::Lint::Pluggable::HTML5';
     requires 'HTML::Lint::Pluggable::WhiteList';
     recommends 'Linux::Inotify2';
+    requires 'List::Util';
```
ehuelsmann: This is a core module. Should not be included in cpanfile

cpanfile
```diff
@@ -104,17 +104,21 @@ feature 'debug', "Debug pane" =>
 on 'develop' => sub {
     requires 'App::Prove', '3.36';
     requires 'File::Util';
+    requires 'Getopt::Long';
```
ehuelsmann: This is a core module. Should not be included in cpanfile

t/07.1-extract-perl.t
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+
+use LedgerSMB::Sysconfig;
```
ehuelsmann: From what I can see, this is here solely for the temp dir. Please don't use Sysconfig for that, instead, use the core module File::Spec's `tmpdir` method.

utils/devel/extract-perl
```diff
@@ -1,132 +1,123 @@
...
 extract a sub-set of strings from our SQL files.
 
 =cut
 
-
 use strict;
 use warnings;
 
+use PPR;
+use utf8;
```
ehuelsmann: This is more a pragma, like `use strict`. It would best be in that block.

utils/devel/extract-perl
```diff
@@ -1,132 +1,123 @@
...
-use constant {
-    NUL => 1,
-    POD => 10,
-    TXT => 20,
-    TXT_FIN => 21,
-};
-
+#TODO: What relevant options to add?
+my $pack;
+  GetOptions ("pack"  => \$pack)     # pack references
```
ehuelsmann: --pack isn't a supported format by poEdit and other po editors. No need to make things more complex by including it (right?).

utils/devel/extract-perl
```diff
@@ -1,132 +1,123 @@
...
+            croak "$file:$line: Direct variable interpolation not supported; use bracketed ([_1]) syntax to replace $1";
+            next;
+        }
+
+        # Remove beginning and end delimitors
+        # The following doesn't work yet
+        #$string = $2
+        #    if $string =~ m( .* ((?&PerlString)) .* $PPR::GRAMMAR);
+        #warn "string = '$string', 1 = <$1>, 2 = <$2>, 3 = <$3>, 4 = <$4>";
+        #TODO: Replace by the above once PPR has added support
```
ehuelsmann: Where is the end of the block being replaced by the above?

cpanfile
ylavoie: Removed

cpanfile
ylavoie: Removed

t/07.1-extract-perl.t
ylavoie: Done

utils/devel/extract-perl
ylavoie: Pack shrinks LedgerSMB.pot to 62% of actual size.

ylavoie:
Perl 5.20 has memory leaks on modules using regex and locale together. See Bug #123198. Holding for now

PR #3102, opened 2017-08-24
Validate localization calls in Perl & LaTEX sources

ylavoie:
Validate localization calls in all Perl & LaTEX files

### DIFF
--- a/xt/06.1-Perl-PO.t
+++ b/xt/06.1-Perl-PO.t
@@ -23,9 +23,10 @@ use File::Find;
 
 my @on_disk;
 sub collect {
-    return if $File::Find::name !~ m/\.(pm|t|tex)$/;
+    return if $File::Find::name !~ m/\.(pl|pm|t|tex)$/;
 
     my $module = $File::Find::name;
+    return if $module =~ m/07.1-extract-perl.t/;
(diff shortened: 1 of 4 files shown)


xt/06.1-Perl-PO.t
```diff
@@ -0,0 +1,95 @@
...
+               ((?&PerlEndOfLine) (??{ ++$line }))*?     # Count the lines
+               (?: (?:[Mm]ark)?[Tt]ext\b                 # Introducer
+                   (?&PerlOWS) [\(\s] (?&PerlOWS))       # '(' or ' '
+               ((?&PerlString)) (?{ push @lines,$line }) # string and line
+               $PPR::GRAMMAR                             # Preload our grammar
+             }gmx;
+
+        # Deduplicate, standardize and pack line references
+        my %text;
+        for (@text) {
```
ehuelsmann: The content of the for block suggests that the extract-perl functionality isn't sufficiently abstracted into a library as this looks like a partial reimplementation.

ylavoie:
Perl 5.20 has memory leaks on modules using regex and locale together. See Bug #123198. Holding for now. Failure on Travis is only on that branch.

PR #3105, opened 2017-08-26
Make PGObject version is at least 2.000002

ylavoie:
Make sure we force upgrade instead of removing constraints.


ehuelsmann:
I disagree with this change. What does it bring us to require this?

ehuelsmann:
Discussion brought up on the development list: http://archive.ledgersmb.org/ledger-smb-devel/msg06689.html

PR #3114, opened 2017-08-31
Validate localization calls in all Perl files

ylavoie:
- Validate localization calls
- Fix commented validation calls

### DIFF
--- a/xt/06.1-Perl-PO.t
+++ b/xt/06.1-Perl-PO.t
@@ -1,0 +1,55 @@
+#!/usr/bin/perl
+
+=pod
+
+This script scans various Perl files for translatable strings.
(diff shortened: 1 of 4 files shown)


xt/06.1-Perl-PO.t
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/perl
+
+=pod
+
+This script scans various Perl files for translatable strings.
+
+The scanner uses the Perl Pattern Recognizer to find text() or marktext()
```
ehuelsmann: Old comments?

ylavoie: Yes, fixing

PR #3115, opened 2017-08-31
Remove base_form diversion in setup

ylavoie:
That diversion isn't needed anymore.

### DIFF
--- a/UI/form-dynatable.html
+++ b/UI/form-dynatable.html
@@ -2,9 +2,7 @@
 <?lsmb PROCESS 'dynatable-old.html' ?>
 
 <body class="lsmb <?lsmb dojo_theme ?>">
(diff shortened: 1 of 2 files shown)


UI/form-dynatable.html
```diff
@@ -2,9 +2,7 @@
 <?lsmb PROCESS 'dynatable-old.html' ?>
 
 <body class="lsmb <?lsmb dojo_theme ?>">
-<form data-dojo-type="<?lsmb IF base_form; base_form;
-                             ELSE;        "lsmb/Form";
-                             END ?>"
+<form data-dojo-type=lsmb/Form"
```
ehuelsmann: missing a quote here: `data-dojo-type=lsmb/Form"` (before `lsmb/`)

ylavoie: Fixed

PR #3119, opened 2017-09-01
Record possible debugger

ylavoie:
For documentation purposes, record a possible Perl debugger for the application.


ehuelsmann:
For next time, doc fixes/additions can be done with a [skip ci] tag at the end of the commit message (because we probably don't need to run tests in those cases)

PR #3121, opened 2017-09-02
Settings database handle access tests

einhverfr:
This currently just creates a test that we can access the database as we currently expect.  The test mocks the database.


einhverfr:
It's worth noting here that the mocks are manual right now.  I would like to create a PGObject::Test::Fixture module eventually after I get some basic experience here with the sorts of mocking needed.

einhverfr:
After 5 days, I haven't seen any negative feedback or a reason these tests are a problem.  Merging.  If there is disagreement we can delete the test file later.

PR #3138, opened 2017-09-07
Remove No unassigned amounts in Transactions test

ylavoie:
Unassigned amounts are SL delayed payments, this test was worthless


lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -481,7 +436,7 @@ push @tests, __PACKAGE__->new(
                                        from gifi
                                       where gifi.accno = account.gifi_accno)
                          and gifi_accno is not null
-                         and gifi_accno !~ '^\\s*\$'},
+                         and gifi_accno !~ '^\s*$'},
```
ehuelsmann: These belong in a separate PR: they are follow-ups to the `q{}` critic changes Rob made and have nothing to do with the removal of the "No unassigned amounts" change.

ylavoie: Fixed

PR #3161, opened 2017-09-13
1.5 fixed assets fixes

ehuelsmann:


sql/modules/Assets.sql
```diff
@@ -665,7 +665,10 @@ $$
      FROM asset_item ai
      JOIN asset_class ac ON (ai.asset_class_id = ac.id)
      JOIN asset_dep_method adm ON (adm.id = ac.method)
-LEFT JOIN asset_report_line rl ON (ai.id = rl.asset_id)
+LEFT JOIN (select arl.*
+             from asset_report_line arl
+             join asset_report ar on arl.report_id = ar.id
+            where approved_at is not null) rl ON (ai.id = rl.asset_id)
 LEFT JOIN asset_report r on (rl.report_id = r.id)
```
einhverfr: I am thinking there has to be a better way to write that.  Not complaining, just noting for later.  The easy ways notably don't work.

ehuelsmann: Thanks for the review. Yes, I figured I could repeat the WHERE and / or HAVING clauses in CASE statements in the aggregators, but that's pretty ugly too...

ehuelsmann: For now (as we have users who're trying to use the module), I'll proceed; we'll need to come up with something "aesthetically correct" later.

PR #3163, opened 2017-09-14
Real Abstract Tables

ziuq:
Just enforcement of the abstract tables.


einhverfr:
I think this change should be made.

ziuq:
@einhverfr:  Agreed.  Just going to put it into `sql/changes/`.  And if that is necessary, then thought should be given to the possibility of data pollution in the  tables.   Will the migration process scrub the tables or will we use 'not valid' to grandfather in any existing records?   I believe the migration aborting or scrubbing is the correct choice.

PR #3166, opened 2017-09-15
Real abstract

ziuq:
Enforce abstraction of tables that are supposed to be abstract.

I think this is a good thing to do but I have reservations about procedure.  Mistakenly, I put these changes in `sql/Pg-database.sql`.  When moved to changes it seems to become a half-done thing.   Someone else will need to resolve the migration issues this creates.   If I thought that task was going to be delayed; I would put this up as an issue instead.


ehuelsmann:
Thanks for this PR. In order to have the change script applied to the schema though, we need to add it to `sql/changes/LOADORDER`.

Such a change would need to be made after this line: https://github.com/ledgersmb/LedgerSMB/blob/master/sql/changes/LOADORDER#L51 
As you have created the change in `sql/changes/1.5`, I take it we'll want to backport the change to 1.5.

ehuelsmann:
It looks like this change was never backported to 1.5. Since we're in the process of working toward a 1.7 release, I think that's no longer required.

PR #3192, opened 2017-09-17
Q17-18 or something

ziuq:
More protection and celebratory elimination of some excessive quoting that was in my path.
We should be installing `ProhibitInterpolationOfLiteral` as this may be all the violations.

Some of my duplicated work, I would have resolved in reverse if I considered that it would be easier to review.  Sorry, I'll try to keep it in mind in the future.

### DIFF
--- a/lib/LedgerSMB.pm
+++ b/lib/LedgerSMB.pm
@@ -495,7 +495,7 @@ sub merge {
         }
         elsif ( !defined $dst_arg && !defined $src->{$arg} )
         {
-            $logger->trace('LedgerSMB.pm: merge setting \$dst_arg is undefined \$src->{\$arg} is undefined');
+            $logger->trace('LedgerSMB.pm: merge setting $dst_arg is undefined $src->{\$arg} is undefined');
         }
         $self->{$dst_arg} = $src->{$arg};
     }
--- a/lib/LedgerSMB/Scripts/payment.pm
+++ b/lib/LedgerSMB/Scripts/payment.pm
@@ -1569,18 +1569,18 @@ if (($Payment->{new_entity_id} != $Payment->{entity_credit_id})&& !$Payment->{ne
       $ovp_repeated_invoices{$avble_invoices[$ref]->{invoice_id}}->{$Selected_entity->{new_accno}} = $Selected_entity->{new_accno};
 
       #the "repeated invoice" flag will check if this invoice has already been printed, if it does, do not print the apply discount checkbox in the UI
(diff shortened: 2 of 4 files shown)


lib/LedgerSMB.pm
```diff
@@ -495,7 +495,7 @@ sub merge {
         }
         elsif ( !defined $dst_arg && !defined $src->{$arg} )
         {
-            $logger->trace("LedgerSMB.pm: merge setting \$dst_arg is undefined \$src->{\$arg} is undefined");
+            $logger->trace('LedgerSMB.pm: merge setting \$dst_arg is undefined \$src->{\$arg} is undefined');
```
ehuelsmann: Now that interpolation doesn't take place anymore, shouldn't the backslashes be removed?

lib/LedgerSMB/DBH.pm
```diff
@@ -38,11 +38,11 @@ sub connect {
     return undef unless $username && $password;
     return undef if $username eq 'logout';
 
-    my $dbh = DBI->connect(qq|dbi:Pg:dbname="$company"|, $username, $password,
+    my $dbh = DBI->connect("dbi:Pg:dbname=$company", $username, $password,
            { PrintError => 0, AutoCommit => 0,
              pg_enable_utf8 => 1, pg_server_prepare => 0 })
         or return undef;
-    $dbh->do("set client_min_messages = 'warning'");
+    $dbh->do('set client_min_messages = \'warning\'');
```
ehuelsmann: Would be nice to use `q{}` here due to the escaping backslashes in the current solution.

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1406,7 +1406,7 @@ my %amount_to_be_used;
 my %ovp_repeated_invoices;
 my %invoice_id_amount_to_pay;
 my $count;
-my $warning = $Payment->{"warning"};
+my $warning = $Payment->{'warning'};
```
ehuelsmann: below, you decided to leave `warning` unquoted. I agree with that choice, but then this one should be unquoted as well -- for consistency.

ehuelsmann:
I love how this PR makes payment.pm much more idiomatic.

lib/LedgerSMB.pm
ziuq: Yes.  Funny, I remember seeing this line and pausing and moving on.  I'm fairly confident that it is the only line with escaped varnames I touched in all this de-interpolation.

Correction pushed.

PR #3198, opened 2017-09-18
Remove spurious warning about main::SOURCE

ylavoie:
Used only once warnings were failing the tests.

### DIFF
--- a/utils/devel/extract-perl
+++ b/utils/devel/extract-perl
@@ -44,6 +44,7 @@ sub add_entry {
 }
 
 sub slurp{
+    no warnings 'once';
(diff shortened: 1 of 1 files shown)


utils/devel/extract-perl
```diff
@@ -44,6 +44,7 @@ sub add_entry {
 }
 
 sub slurp{
+    no warnings 'once';
```
ehuelsmann: Isn't there anything we can do about the file in fact being used once, instead of suppressing?

ylavoie: Rewritten to use modern Perl and remove the no warning.

PR #3199, opened 2017-09-18
Require newer to circumvent _DBH initialization problems

ylavoie:
PGObject::Simple::Role sometimes uses _DBH before initialization, which fails in our template tests.
Force upgrade.


ylavoie:
Retired. Mixup of versions in Docker-dev between distro and cpanm because of the use of our PPA

PR #3205, opened 2017-09-21
Remove Dojo leftovers from previous optimization

ylavoie:
Make sure that missing GIFI tests do not include chart headers.

### DIFF
--- a/UI/lib/ui-header.html
+++ b/UI/lib/ui-header.html
@@ -24,14 +24,8 @@
             # *persistent*
             # http://www.w3.org/TR/html401/present/styles.html#h-14.3.1
         ?>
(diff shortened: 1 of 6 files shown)


ehuelsmann:
I'm sorry, but DOJO_LOCATION *is* being used in the source of this page.

ylavoie:
Isn't it already included by ui-header?

sbts:
As per discussions above [here](https://matrix.to/#/!qyoLumPqusaXqFJNyK:matrix.org/$15062191962769163QFXJZ:matrix.org) changing the general architecture to propagate both dojo_built and dojo_location makes sense.
Then location is determined at the perl level rather than html/js/templating level
Also it's only calculated once in Sysconfig.pm

PR #3207, opened 2017-09-23
no critic

ziuq:
no critic ProhibitMagicNumbers  to allow enforcement


lib/LedgerSMB/Entity/Location.pm
```diff
@@ -111,7 +111,7 @@ $self->set_class_name is called.
 
 our %classes = ( 1 => $locale->text('Billing'),
                  2 => $locale->text('Sales'),
-                 3 => $locale->text('Shipping'),
+                 3 => $locale->text('Shipping'),  ## no critic (ProhibitMagicNumbers) sniff
```
ehuelsmann: Can't we create file-local constants for these? Or does that make no sense at all?

lib/LedgerSMB/Scripts/budgets.pm
```diff
@@ -50,8 +50,8 @@ sub new_budget {
 
 sub _render_screen {
     my ($budget) = @_;
-    my $additional_rows = 5;
-    $additional_rows +=20 unless $budget->lines;
+    my $additional_rows = 5;  ## no critic (ProhibitMagicNumbers) sniff
+    $additional_rows +=20 unless $budget->lines;  ## no critic (ProhibitMagicNumbers) sniff
```
ehuelsmann: 5 could be the constant ADDITIONAL_ROWS whereas 20 could be BASE_ROW_COUNT?

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -114,7 +114,7 @@ sub _aa_multi {
         $acst->execute($acct);
         ($pass) = $acst->fetchrow_array;
         $request->error("Account $acct not found") if !$pass;
-        ($acct) = split /--/, $ref->[3];
+        ($acct) = split /--/, $ref->[3];  ## no critic (ProhibitMagicNumbers) sniff
         $acst->execute($acct);
```
ehuelsmann: The columns in the CSV input aren't accessed by heading, but by number. Maybe we should create constants which map the numbered columns to mnemonics?

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -416,7 +416,7 @@ sub print {
             my $amt = $check->{amount}->copy;
             $amt->bfloor();
             $check->{text_amount} = $payment->text_amount($amt);
-            $check->{decimal} = ($check->{amount} - $amt) * 100;
+            $check->{decimal} = ($check->{amount} - $amt) * 100;  ## no critic (ProhibitMagicNumbers) sniff
```
ehuelsmann: I thought you introduced a "fractional currency to cents" conversion factor? that could be applied here.

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -482,14 +482,14 @@ sub display_payments {
                 $payment->{"paid_$_->{contact_id}"} = ''
                     unless defined $payment->{"paid_$_->{contact_id}"};
             }
-            $invoice->[6] = $invoice->[3] - $invoice->[4] - $invoice->[5];
-            $contact_to_pay += $invoice->[6];
-            $invoice->[7] = $invoice->[6]->to_db;
+            $invoice->[6] = $invoice->[3] - $invoice->[4] - $invoice->[5];  ## no critic (ProhibitMagicNumbers) sniff
+            $contact_to_pay += $invoice->[6];  ## no critic (ProhibitMagicNumbers) sniff
+            $invoice->[7] = $invoice->[6]->to_db;  ## no critic (ProhibitMagicNumbers) sniff
```
ehuelsmann: These all would benefit from a series of constants "naming" the column indexes.

lib/LedgerSMB/X12.pm
```diff
@@ -35,7 +35,7 @@ use LedgerSMB::Magic qw( EDI_PATHNAME_MAX );
 use LedgerSMB::Sysconfig;
 use DateTime;
 
-my $counter = 1000; #for 997 generation
+my $counter = 1000; #for 997 generation  ## no critic (ProhibitMagicNumbers) sniff
```
ehuelsmann: This counter isn't used in this file. Maybe we don't need to put a "no critic" on it, but instead just remove it?

lib/LedgerSMB/X12.pm
```diff
@@ -112,7 +112,7 @@ sub _ISA {
 
     my @keys;
 
-    push @keys, sprintf('ISA%02d', $_) for (1 .. 16);
+    push @keys, sprintf('ISA%02d', $_) for (1 .. 16);  ## no critic (ProhibitMagicNumbers) sniff
```
ehuelsmann: Referring to http://edicrossroad.blogspot.nl/2008/12/isa-and-gs-segment-elements-enumeration.html , 1 is the minimum ISA header number and 16 the max. So, it seems we can safely create constants for these.

lib/LedgerSMB/Entity/Location.pm
ziuq: I am confused.  I do not grasp the idea behind using ints as hash keys; it just seems like useless indirection to me.    'Physical' and 'Marketing" already seem to be missing.

lib/LedgerSMB/Scripts/budgets.pm
ziuq: I recognized the similarity to the check writing code, iirc.   This made me wonder about the correctness of that change.  Why would a budget and  check stub have the same  limits?

lib/LedgerSMB/Scripts/import_csv.pm
ziuq: I suspect that when I see a batch being run this may make sense, but static analysis is losing in a cost/benefit way.

lib/LedgerSMB/Scripts/payment.pm
ziuq: I thought I got all of these.  I'll get to this.

lib/LedgerSMB/X12.pm
ziuq: I considered that.   But my thinking went:  What's '997 generation'?  Aspiring or dying code?  Don't know.

I dislike no-criticing, but not finding the time to fix makes it compelling to avoid more
infractions.

lib/LedgerSMB/X12.pm
ziuq: I remember not finding any authoritative naming for the EDI fields.  I cannot read this page now.  Do you accept it as authoritative?  Does it have more than "ISA07" for names?

I suppose the loop could be unrolled. 

What do you think?

ehuelsmann: Yes, the page has a mapping between codes and description:

```
    ISA01 Authorization Information Qualifier : min/max – 2/2
    ISA02 Authorization Information : min/max – 10/10
    ISA03 Security Information Qualifier : min/max – 2/2
    ISA04 Security Information : min/max – 10/10
    ISA05 Interchange ID Qualifier : min/max – 2/2
    ISA06 Interchange Sender ID : min/max – 15/15
    ISA07 Interchange ID Qualifier : min/max – 2/2
    ISA08 Interchange Receiver ID : min/max – 15/15
    ISA09 Interchange Date : min/max – 6/6
    ISA10 Interchange Time : min/max – 4/4
    ISA11 Interchange Control Standards ID : min/max – 1/1
    ISA12 Interchange Control Version Number : min/max – 5/5
    ISA13 Interchange Control Number : min/max – 9/9
    ISA14 Acknowledgment Requested : min/max – 1/1
    ISA15 Test Indicator : min/max – 1/1
    ISA16 Subelement Separator : min/max – 1/1
```

With this information, unrolling the loop is probably the best option we have.

lib/LedgerSMB/X12.pm
ziuq: We missed the use of `${counter}`

lib/LedgerSMB/Entity/Location.pm
ziuq: Where is `%classes` used?

ziuq: Where is `%classes` used?

lib/LedgerSMB/Scripts/budgets.pm
ziuq: How about EDIT_BUDGET_ROWS and NEW_BUDGET_ROWS?   It looks like the 25 lines are for an empty budget to be entered and 5 lines are a trailer for expanding a budget.

lib/LedgerSMB/Scripts/payment.pm
ziuq: Actually I PRed and you accepted 100 as a non-magical number.

lib/LedgerSMB/Scripts/payment.pm
ziuq: I will pass on this task as I have no idea of the namings.

lib/LedgerSMB/Entity/Location.pm
ehuelsmann: Nowhere, from what I can see. I've run the test-suite with it commented out and the tests still run fine...

lib/LedgerSMB/Scripts/budgets.pm
ehuelsmann: Sounds good. Go for it.

lib/LedgerSMB/Scripts/payment.pm
ehuelsmann: Ok. But if that's the case, then we don't need to mark 100 here as a magical number for which the critic needs to be suppressed, right?

lib/LedgerSMB/Scripts/payment.pm
ehuelsmann: ok. then I'll do that as soon as this gets merged.

PR #3208, opened 2017-09-23
no critic ProhibitExplicitReturnUndef

ziuq:
to allow enforcement.

### DIFF
--- a/lib/LedgerSMB.pm
+++ b/lib/LedgerSMB.pm
@@ -249,7 +249,7 @@ sub clear_session {
 
     $self->{cookie} = '';
 
-    return undef;
+    return undef;  ## no critic (ProhibitExplicitReturnUndef) sniff
 }
 
 sub verify_session {
--- a/lib/LedgerSMB/DBH.pm
+++ b/lib/LedgerSMB/DBH.pm
@@ -35,13 +35,13 @@ Returns undef on connection failure or lack of credentials.
 sub connect {
     my ($package, $company, $username, $password) = @_;
 
-    return undef unless $username && $password;
-    return undef if $username eq 'logout';
+    return undef unless $username && $password;  ## no critic (ProhibitExplicitReturnUndef) sniff
+    return undef if $username eq 'logout';  ## no critic (ProhibitExplicitReturnUndef) sniff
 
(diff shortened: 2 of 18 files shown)


ehuelsmann:
I really doubt we want to enforce "ProhibitExplicitReturnUndef", because `return undef;` and `return;` are semantically different. I know we have discussed this before and that we don't have the same view on returning undef from a function due to the fact that it may be used in array context.

However, advice from outside our community was to explicitly return undef everywhere where we currently `return;`. Adding @einhverfr and @sbts for more review and discussion here.

ziuq:
Fixing "bad" code is not what perlcritic is about.  It seems that way when we critique and fix all the trivial errors.  Generating "good" new code is the purpose and that only starts when policies are in place.

Yes, I do not think a _RequireExplicitReturnUndef_ policy is good.  Looking at CPAN and tallying violations and no-critics on all the not-my Perl code installed on my personal machine, I find I am with the majority opinion.   The main argument given for _RequireExplicitReturnUndef_ is a version of "if you are incapable of understanding context you can do this".   I do not believe this group has that incapacity.

I do understand the devilish appeal of saving time by fewer keystrokes and by learning the least necessary to handle a task.   Sometimes those even work.

cloc on system Perl dirs:
Language                             files          blank        comment           code
Perl                                  4251         278864         543658        1170138

Violations:
perlcritic -s with   ProhibitExplicitReturnUndef EmptyReturn
/usr/*                                     2107               2673
lsmb-master/old                         5                     7
lsmb-master/lib                        37                   16

No-Critics in place:

Total                                       613
With a name                          467
"*Return*"                                19
"*EmptyReturn*"                       0
"*ProhibitExplicitReturnUndef*" 7

ehuelsmann:
> The main argument given for RequireExplicitReturnUndef is a version of "if you are incapable of understanding context you can do this". I do not believe this group has that incapacity.

Right. I think that's correct. However, we want to leave the project in a state which makes it easy to contribute for others as well (not limited to those who are already way into Perl).

ehuelsmann:
The above comment meaning: I'm still very much in doubt which way to go. (Either way, I think we should introduce a critic to enforce it that way, because otherwise we'll end up with a mixed style code base...)

PR #3212, opened 2017-09-25
Disable migration when AR or AP missing

ylavoie:
- Adjust dialogs for any LSMB and SL versions
- Reformat (add &w=1 to remove space changes)
- Allow cancel migration if default AR or AP cannot be find.

### DIFF
--- a/UI/setup/upgrade_info.html
+++ b/UI/setup/upgrade_info.html
@@ -9,7 +9,7 @@ include_stylesheet=["setup/stylesheet.css"] ?>
 <form data-dojo-type="lsmb/SimpleForm"
       action="setup.pl" method="POST"
       name="upgrade_info">
-            <?lsmb upgrade_allowed = 1 ?>
+<?lsmb upgrade_allowed = 1 ?>
 <?lsmb INCLUDE input element_data = {
     name = 'database'
     type = 'hidden'
@@ -22,11 +22,10 @@ include_stylesheet=["setup/stylesheet.css"] ?>
 </p>
 <p>
   In addition to these new defaults LedgerSMB adds stricter
-  checks on data validity in the database. Because of these
-  stricter checks it's no longer valid to leave companies
-  without a country or customers without accounts receivable
-  reference. The defaults you choose below will be used to add
-  values where these are currently missing but required.
+  checks on data validity in the database. Because of these stricter checks
+  it's no longer valid to leave companies without a country or customers
+  without accounts receivable reference. The defaults you choose below will
+  be used to add values where these are currently missing but required.
 </p>
 <div class="input_row">
 <?lsmb INCLUDE select element_data = {
@@ -36,28 +35,27 @@ include_stylesheet=["setup/stylesheet.css"] ?>
                             name  = 'default_country'
                            label  = text('Default Country') #'
                             class = 'country'
+                         required = 1
 } ?>
 </div>
 <div class="input_row">
 <p>
-  LedgerSMB supports multiple <em>Accounts receivable (AR)
-  </em> accounts per company.
-  <?lsmb IF ar_accounts.size > 1 ?>
-    <span>One of those must be the system default.
-   Please select your default below.</span>
-  <?lsmb ELSE; ?>
+  LedgerSMB supports multiple <em>Accounts receivable (AR) </em> accounts
+per company.
+<?lsmb IF ar_accounts.size > 1 ?>
+    <span>One of those must be the system default. Please select
+your default below.</span>
+<?lsmb ELSE; ?>
     <span>Your AR account has been selected to be the
     system default. Please review.</span>
-  <?lsmb END ?>
-  <?lsmb AR_DISABLED = "";
-    IF ar_accounts.size <= 1 && !ar_accounts.0.id; ?>
+<?lsmb END ?>
+<?lsmb AR_DISABLED = "";
+IF ar_accounts.size <= 1 && !ar_accounts.0.id && lsmb_info.ar > 0; ?>
     <span style='font-weight:bold'>
           No <em>AR</em> account available, your database is
           in an inconsistent state and needs to be fixed first.
     </span>
-  <?lsmb AR_DISABLED = "disabled";
-         upgrade_allowed = 0;
-  END ?>
+  <?lsmb AR_DISABLED = "disabled"; END ?>
 </p>
 <?lsmb INCLUDE select element_data = {
                             name  = 'default_ar'
@@ -66,29 +64,28 @@ include_stylesheet=["setup/stylesheet.css"] ?>
                             options = ar_accounts
                             value_attr = 'accno'
                             text_attr = 'desc'
+                            required = 1
                             disabled = AR_DISABLED
 } ?>
 </div>
 <div class="input_row">
 <p>
-LedgerSMB supports multiple <em>Accounts payable (AP)
-</em> accounts per company.
-<?lsmb IF ap_accounts.size > 1 ?>
-<span>One of those must be the system default.
-Please select your default below.</span>
-<?lsmb ELSE; ?>
-<span>Your AP account has been selected to be the
-system default. Please review.</span>
-<?lsmb END ?>
-<?lsmb AP_DISABLED = "";
-IF ap_accounts.size <= 1 && !ap_accounts.0.id; ?>
-<span style='font-weight:bold'>
-      No <em>AP</em> account available, your database is
-      in an inconsistent state and needs to be fixed first.
-</span>
-<?lsmb AP_DISABLED = "disabled";
-     upgrade_allowed = 0;
-END ?>
+  LedgerSMB supports multiple <em>Accounts payable (AP) </em> accounts
+per company.
+  <?lsmb IF ap_accounts.size > 1 ?>
+    <span>One of those must be the system default.
+   Please select your default below.</span>
+  <?lsmb ELSE; ?>
+    <span>Your AP account has been selected to be the
+    system default. Please review.</span>
+  <?lsmb END ?>
+  <?lsmb AP_DISABLED = "";
+    IF ap_accounts.size <= 1 && !ap_accounts.0.id && lsmb_info.ap > 0; ?>
+    <span style='font-weight:bold'>
+          No <em>AP</em> account available, your database is
+          in an inconsistent state and needs to be fixed first.
+    </span>
+  <?lsmb AP_DISABLED = "disabled"; END ?>
 </p>
 <?lsmb INCLUDE select element_data = {
                             name  = 'default_ap'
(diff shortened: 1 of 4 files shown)


UI/setup/upgrade_info.html
```diff
@@ -50,14 +65,29 @@
...
+   Please select your default below.</span>
+  <?lsmb ELSE; ?>
+    <span>Your AP account has been selected to be the
+    system default. Please review.</span>
+  <?lsmb END ?>
+  <?lsmb AP_DISABLED = "";
+    IF ap_accounts.size <= 1 && !ap_accounts.0.id; ?>
+    <span style='font-weight:bold'>
+          No <em>AP</em> account available, your database is
+          in an inconsistent state and needs to be fixed first.
```
ehuelsmann: Are we 100% sure that it's impossible to run books without an AR or AP account? I was under the impression we can (and thus should allow migration of data which doesn't).

einhverfr: Hmmm...  I can imagine some financial services might be able to work without an AR account because clients might actually be vendors that one accepts money on behalf of, and then distribute on a discounted basis.

At the same time, you *could* do everything over GL and not worry bout the map fields at all.

To improve this check maybe one should check to see if an AR account exists and, if not, if there are records in the ar table?  Ditto for AP?

ylavoie: In order to accept no AR account, there must not be any customer defined, and no vendor for AP. Migration will fail otherwise because we explicitly check for existing entity credit/debit account.

ehuelsmann: Ok. Then the direction proposed by Chris sounds like the right way to go: no customer ecas *and* no AR? Then no problem. Same for AP.

ylavoie: Done. Added Dojo form validation and submit button enabling

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -621,28 +621,29 @@ sub upgrade_info {
     my $dbinfo = $database->get_info();
     my $upgrade_type = "$dbinfo->{appname}/$dbinfo->{version}";
 
+    $request->{lsmb_info} = $database->stats();
 
     if (applicable_for_upgrade('default_ar', $upgrade_type)) {
-    @{$request->{ar_accounts}} = _get_linked_accounts($request, 'AR');
-    unshift @{$request->{ar_accounts}}, {}
-            unless scalar(@{$request->{ar_accounts}}) == 1;
+        @{$request->{ar_accounts}} = _get_linked_accounts($request, 'AR');
```
ehuelsmann: Chris (and others) expressed that we should separate indentation changes from functional changes. From what I can tell, I'm looking at a single commit with both functional and whitespace changes?

UI/setup/upgrade_info.html
```diff
@@ -6,10 +6,9 @@
 <div class="setupconsole">
 <h2><?lsmb text('Database Management Console') ?></h2>
 <div class="listtop"><?lsmb text('Upgrade Info') ?></div>
-<form data-dojo-type="dijit/form/Form"
+<form data-dojo-type="lsmb/setup/upgrade_info"
```
ehuelsmann: We have "data migration pre-checks" in the form of our Upgrade_Tests. Why are we suddenly taking a radically different approach? Is it impossible to handle the situation in an Upgrade_Test?

lib/LedgerSMB/Scripts/setup.pm
ylavoie: On this only 1 line, yes. May I remind everyone that adding `?w=1` on the url removes whitespace changes and allows to focus on the issue being fixed

ehuelsmann: well, in https://github.com/ledgersmb/LedgerSMB/pull/3212/commits/b61a26300445dd386e502f0ad454c01d0adf2c06 (a view on 1 commit), I see 1 line added to setup.pm and many lines changing indent.

The problem is that "the future" has to go and try to see why the whitespace changes get attributed to a change which says "AR/AP detection" in 'blame' output.

UI/setup/upgrade_info.html
ylavoie: This only thing this does is reflect the form input validity onto the submit button. This isn't yet at Upgrade_Test level, and should really have been handled by Dojo. When user inputs are coherent with fields restrictions, we should validate the Upgrade_Test.

ehuelsmann: But ... we don't do that anywhere else in the application... starting to do so here right now will result in an inconsistent UI experience. Is that what we want?

ylavoie: On the contrary, each field where we added constraints do prevent using the submit button. They just do not reflect it on the button state,  but refuse the submit when clicked with invalid fields on the form.

ehuelsmann: Agreed: clicking the button will show which fields are blocking submission. It's like that in the entire application. Why should we use this change as an opportunity to create inconsistencies?

ehuelsmann:
In general, I'm wondering if we really need AR and/or AP accounts to be in the ledger: what if people have been using GL postings only and have not had the need to classify accounts with the summary classifications and checkmarks; do we not want those people to migrate to 1.5 or later?

UI/setup/upgrade_info.html
ylavoie: Why do we impose waiting for the user to click the submit button instead of handling the form validity. We know that the form is invalid but do not tell and force the user to ask. I say that this should be reflected elsewhere instead of mainting the current behaviour.

ehuelsmann: I'm not saying we want to maintain the current behaviour. I'm saying if we want to change it, we need to do so everywhere. And a change like that needs its own PR. All I'm saying is that is not good to be inconsistent and that changing the behaviour is not part of the topic for this PR.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -683,6 +684,12 @@ sub upgrade {
     $request->{dbh}->{AutoCommit} = 0;
     my $locale = $request->{_locale};
 
+    for my $preset (LedgerSMB::Upgrade_Tests->get_presets) {
+        my $sth = $request->{dbh}->prepare($preset);
+        $sth->execute()
+            or die 'Failed to execute pre-migration preset ';
```
ehuelsmann: What is a pre-migration preset?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -546,6 +619,104 @@ number of "AR_*", "AP_*" and/or "IC_*" links concatenated by colons (:).'),
...
+                       where ((select array_agg(x in (
+                            SELECT value FROM pg_temp.account_link
+                           )) from unnest(string_to_array(c.link,':')) x) @> ARRAY[0]::bool[]
+                           or c.link is null
+                           or c.link = '')
+                         and charttype = 'A'
+                    order by accno;
+                    },
+    display_name => marktext('No unlinked account'),
+    name => 'no_unlinked_account',
```
ehuelsmann: I think this check is incorrect: accounts are allowed to be unlinked. In that case they only have "manual journal" value, but I have many accounts like that myself... Nothing to be worried about.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -546,6 +619,104 @@ number of "AR_*", "AP_*" and/or "IC_*" links concatenated by colons (:).'),
...
+                         group by link
+                         having link = 'AR'
+                     )
+                     order by accno
+                     },
+    display_name => marktext('No AR account'),
+    name => 'no_ar_account',
+    display_cols => ['accno', 'description', 'link'],
+ instructions => marktext(
+                   'Please make sure one AR account exists'),
```
ehuelsmann: Maybe we need to eliminate the question about the default AR account instead; asking this question *only* when there are multiple AR accounts?
My point: if I don't need an AR account for my use of LedgerSMB, then ... Why would the migration force me to create one?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -1069,22 +1240,37 @@ push @tests, __PACKAGE__->new(
...
+                  order by c.accno, ap.transdate, id, ac.cleared, ap.datepaid},
   display_name => marktext('Invalid or suspect cleared delays'),
           name => 'invalid_cleared_dates',
-  display_cols => ['name', 'id', 'datepaid', 'transdate', 'cleared', 'delay', 'amount'],
+  display_cols => ['name', 'accno', 'id', 'transdate', 'datepaid', 'cleared', 'delay', 'amount'],
+     alignment => {accno => 'center', transdate => 'center', datepaid => 'center',
+                    cleared => 'center', delay => 'right', amount => 'right'},
  instructions => marktext(
-                   'Suspect or invalid cleared delays have been detected. Please review the dates in the original application'),
+                   'Cleared dates much later than the transaction dates are suspect. Cleared dates before the transaction dates are invalid. Please review the dates in the original application'),
```
ehuelsmann: I think this change falls outside the scope of "disable migration when there's no AR or AP account", right?

If/When we decide to merge both functionalities, I think we want them in separate PRs.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -184,6 +199,18 @@ Human readable, localized display name
 
 has display_name => (is => 'ro', isa => 'Str', required => 1);
 
+=item alignment
+
+Column alignment
```
ehuelsmann: When writing documentation, please make it elaborate enough for another programmer to understand what to expect in the field or even what to put in it (and what the effects of that will/are supposed to be).

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -24,6 +24,18 @@ use LedgerSMB::Locale qw(marktext);
 
 =over
 
+=item get_presets()
+
+Returns the test array
```
ehuelsmann: Which test array? (Note, this is a request for more documentation, not for an answer to the question in a GitHub response)

UI/setup/upgrade_info.html
```diff
@@ -50,14 +63,28 @@ <h2><?lsmb text('Database Management Console') ?></h2>
...
+per company.
+  <?lsmb IF ap_accounts.size > 1 ?>
+    <span>One of those must be the system default.
+   Please select your default below.</span>
+  <?lsmb ELSE; ?>
+    <span>Your AP account has been selected to be the
+    system default. Please review.</span>
+  <?lsmb END ?>
+  <?lsmb AP_DISABLED = "";
+    IF ap_accounts.size <= 1 && !ap_accounts.0.id && lsmb_info.ap > 0; ?>
```
ehuelsmann: how can `lsmb_info.ap` be bigger than `0` when there's no account to hold these transactions?

ehuelsmann: If this would happen in actual practice, I'd suggest "serious database integrity damage (please consult a LedgerSMB professional or start over)"

lib/LedgerSMB/Scripts/setup.pm
ylavoie: I simplified some tests by pre-setting utility tables and routines in pg_temp before doing upgrade tests.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: A very good idea if there is an easy way to check when they are required. SL Customers and vendors are mapped in LSMB entities, so we would been database specific checks.
Currently, we were advocating the need for a default AR and AP acount, so this code only enforces that statement and default automatically when only 1 is present in a category.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Moved in another PR

lib/LedgerSMB/Upgrade_Tests.pm
ehuelsmann: We advocate having one, but what happens if there isn't one? Or when I don't select one? Do things fail?

Issue #3214, opened 2017-09-26
Duplicated initial user not managed anymore


ylavoie:
After migration or creation of a new database, the creation of the initial user will fail if it already exists. A pending aborted transaction will fail the next tries.

ehuelsmann:
Fixed by #3215

PR #3215, opened 2017-09-26
Fix duplicated initial user handling

ylavoie:
We need to manage the whole transaction after having disabled autocommit.
Initial user creation checks for duplicates and tie employee and user creation into a transaction to make sure they are done together. Transactions require non-transparent handling of commits and rollbacks. The latter was forgotten in the initial code.
Fixes issue #3214.

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -1074,7 +1074,7 @@ sub save_user {
 
     _init_db($request);
     #We enable transactions, we will have to manage commits and rollbacks ourselves.
-    $request->{dbh}->{AutoCommit} = 0; 
+    $request->{dbh}->{AutoCommit} = 0;
 
     $request->{control_code} = $request->{employeenumber};
     $request->{dob} = LedgerSMB::PGDate->from_input($request->{dob});
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -1073,7 +1073,7 @@ sub save_user {
     use LedgerSMB::PGDate;
 
     _init_db($request);
-    $request->{dbh}->{AutoCommit} = 0;
+    $request->{dbh}->{AutoCommit} = 0; #Enable transactions, must manage.
 
     $request->{control_code} = $request->{employeenumber};
     $request->{dob} = LedgerSMB::PGDate->from_input($request->{dob});
@@ -1085,6 +1085,7 @@ sub save_user {
     try { $user->create($request->{password}); }
     catch {
         if ($_ =~ /duplicate user/i){
+           $request->{dbh}->rollback;
            $request->{notice} = $request->{_locale}->text(
                        'User already exists. Import?'
             );
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -1073,7 +1073,6 @@ sub save_user {
     use LedgerSMB::PGDate;
 
     _init_db($request);
-    #We enable transactions, we will have to manage commits and rollbacks ourselves.
     $request->{dbh}->{AutoCommit} = 0;
 
     $request->{control_code} = $request->{employeenumber};
@@ -1086,7 +1085,7 @@ sub save_user {
     try { $user->create($request->{password}); }
     catch {
         if ($_ =~ /duplicate user/i){
-           $request->{dbh}->rollback; # Rolling back because of the error
+           $request->{dbh}->rollback;
            $request->{notice} = $request->{_locale}->text(
                        'User already exists. Import?'
             );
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1073,7 +1073,7 @@ sub save_user {
     use LedgerSMB::PGDate;
 
     _init_db($request);
-    $request->{dbh}->{AutoCommit} = 0;
+    $request->{dbh}->{AutoCommit} = 0; #Enable transactions, must manage.
```
ehuelsmann: Can you explain what you want to tell the programmer reading this comment? Especially the "must manage" part? Doesn't the caller roll back on error and commit on success? Then isn't the transaction already managed?

ylavoie: We decided to remove autocommit. We thus have to manage ourselves the commit and rollbacks.

ehuelsmann: Ok, but is this a special case? As in: apparently, the programmer nweds to be made aware of this, because this is an exceptional case. Can you provide some insight into what is exceptional here? As per the "inline comments" section of our coding guide lines https://ledgersmb.org/content/coding-guidelines

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1073,7 +1073,8 @@ sub save_user {
     use LedgerSMB::PGDate;
 
     _init_db($request);
-    $request->{dbh}->{AutoCommit} = 0; #Enable transactions, must manage.
+    #We enable transactions, we will have to manage commits and rollbacks ourselves.
```
ehuelsmann: From https://ledgersmb.org/content/coding-guidelines (emphasis mine):
> comments should describe things which are __*outside the code*__ -- why's and wherefore's, promises to developers, reasoning, etc.

Can you explain what's *outside the code* about your comment? I think 'we enable transactions' is an exact description *of* the code (so, it's *in* the code). The immediate consequence of it is then that we need to - of course - finish the transactions.

A comment which explains what's outside this direct code is:

```
# The request dispatcher will commit the transaction for us (or rollback in case of error)
```

However, it seems we don't depend on that behaviour here(?) given the rollbacks below, so, it would be a misleading comment in this exact case.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1085,7 +1086,7 @@ sub save_user {
     try { $user->create($request->{password}); }
     catch {
         if ($_ =~ /duplicate user/i){
-           $request->{dbh}->rollback;
+           $request->{dbh}->rollback; # Rolling back because of the error
```
ehuelsmann: This too is an exact description of what the statement does. Is there anything you want to add here that's not visible on the programmer's screen when (s)he looks at the code? Then that would be a helpful comment, as per the quote from Chris's description above.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1085,6 +1085,7 @@ sub save_user {
     try { $user->create($request->{password}); }
     catch {
         if ($_ =~ /duplicate user/i){
+           $request->{dbh}->rollback;
            $request->{notice} = $request->{_locale}->text(
                        'User already exists. Import?'
```
ehuelsmann: If we don't roll back here, we run into problems generating/querying elements for the page to be returned.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1111,7 +1112,10 @@ sub save_user {
            die $_;
        }
     };
-    return $duplicate if $duplicate;
+    if ( $duplicate ) {
+        $request->{dbh}->rollback;
```
ehuelsmann: However, that's not the case here. What happens if we don't roll back here?

ylavoie: 1- A transaction is started and we return with it still opened.
2- The remainder of the code will assume no transaction and no commits will be done, for we rely on autocommitting.
3- The first error will invalidate everything from this point on

ehuelsmann: Actually, I don't agree with (2). The request dispatcher will commit upon finalizing the request. So, commits *will* happen.

ylavoie: And this will commit the first part of an incomplete transaction, much later in the process and at a completely unrelated moment.

ehuelsmann: But you said the rest of the code won't commit. My point is that the code *will* commit.
In my view the moment of commit isn't unrelated as it is the commit to go with the returned document/user feedback.

My question was a different one though: is there code following this code which needs to query the database and needs to succeed at that?
If so, there are 2 follow up questions:
1. Why do we need a transaction here?
2. Why not just let the code run with auto-commit turned back on after the error?

ylavoie: 1- The code won't commit there. Somewhen, much after, an unrelated commit will impose the partial pending transaction.
2- Most likely we need a transaction because the original developer felt that employee and user creation were related. I agree with that assumption.
3- We create the employee, then the user. Without transaction, we would have to somehow undo the employee alterations.

PR #3217, opened 2017-09-29
Add DBIProfile, LazyLoad, Trace ENV and W3CValidate development tools

ylavoie:
- Add dev tools
- Remove duplicate in cpanfile

### DIFF
--- a/tools/starman-development.psgi
+++ b/tools/starman-development.psgi
@@ -50,7 +50,7 @@ sub check_config_option {
     my ($name,$module) = @_;
     return 0 if !eval "\$LedgerSMB::Sysconfig::$name";
     return 1 if !defined $module;
-    unless (eval "use $module") {
+    unless (eval "require $module") {
         warn "$name requires $module";
         return 0;
     }
--- a/tools/starman-development.psgi
+++ b/tools/starman-development.psgi
@@ -84,8 +84,8 @@ builder {
             if check_config_option('DBITrace','Plack::Middleware::Debug::DBITrace');
         enable 'Debug::TraceENV', method => $LedgerSMB::Sysconfig::TraceENV_method
             if check_config_option('TraceENV','Plack::Middleware::Debug::TraceENV');
-        enable 'Profiler::NYTProf', exclude => $LedgerSMB::Sysconfig::NYTProf_exclude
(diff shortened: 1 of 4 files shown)


tools/starman-development.psgi
```diff
@@ -55,14 +55,18 @@ Log::Log4perl::init(\$LedgerSMB::Sysconfig::log4perl_config);
 
 builder {
 
-    enable 'InteractiveDebugger';
+#    enable 'InteractiveDebugger';
```
ehuelsmann: Maybe we need to check for its availability and/or it being configured in 'ledgersmb.conf'? That feels much better than enabling/disabling code by putting it in comments.

tools/starman-development.psgi
```diff
@@ -55,14 +55,18 @@ Log::Log4perl::init(\$LedgerSMB::Sysconfig::log4perl_config);
...
     enable 'Debug',  panels => [
             qw(Parameters Environment Response Log4perl Session Timer Memory ModuleVersions PerlConfig),
-              [ 'DBITrace', level => 2 ],
-              [ 'Profiler::NYTProf', exclude => [qw(.*\.css .*\.png .*\.ico .*\.js .*\.gif .*\.html)], minimal => 1 ],
+              [ 'DBIProfile', profile => 2 ],
+#             [ 'DBITrace', level => 2 ],
+              'LazyLoadModules',
+              [ 'TraceENV', method => [qw/fetch store exists delete clear scalar firstkey nextkey/]],
+              'W3CValidate',
+#              [ 'Profiler::NYTProf', exclude => [qw(.*\.css .*\.png .*\.ico .*\.js .*\.gif .*\.html)], minimal => 1 ],
```
ehuelsmann: Same here.

tools/starman-development.psgi
ylavoie: Agreed, although at this point, those are mostly to help others develop and are not as standardize as the application.

ehuelsmann: Agreed. So these could be settings in Sysconfig's Debug section. I still feel like commenting out code "because it is for debugging" is a workaround for want of a real solution. Using config feels much more solid. 

Note that I'm not arguing that this code be integrated with the regular starman.psgi .

PR #3218, opened 2017-09-30
Add password field to dynatable

ylavoie:
Allow masked passwords in dynatable


einhverfr:
Probably a good idea.

PR #3225, opened 2017-10-06
Add MultiSelect capability

ylavoie:
There was an uncompleted attempt in elements.html to have MultiSelect capability. This fixes that.

### DIFF
--- a/UI/lib/elements.html
+++ b/UI/lib/elements.html
@@ -230,11 +230,10 @@
           <?lsmb PROCESS attributes  # Process regular attributes.
             attribute_data=element_data
             attribute_defaults =
-                (element_data.type.defined() && element_data.type == "multiple")
+                (element_data.type.defined() && element_data.type == "multiselect")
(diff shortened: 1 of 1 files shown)


ehuelsmann:
What's the value here?

https://www.w3schools.com/TAGS/att_input_multiple.asp

says it's only applicable to e-mail and file INPUT tags.

ehuelsmann:
Ah. Now I see that it's my mistake (or w3school's), because multiple applies to SELECT too. However, they also note:

> Because of the different ways of doing this, and because you have to inform the user that multiple selection is available, it is more user-friendly to use checkboxes instead.

which brings me back to the question: what's the vale here?

sbts:
Personally I like multi select, and find it annoys me when it's not available in places where it would be convenient.
Checkboxes are nice from a visual perspective, but SUCK from a UX perspective if there is a large list and you want to select one or more contiguous blocks

PR #3227, opened 2017-10-06
Remove SQL notices on DROP IF EXISTS

ylavoie:
Lower the noise level, those notices don't serve any usefull purpose.


ehuelsmann:
While I agree, with your comment, we'll need to find another way to achieve the same result. Those change files are not to be changed anymore: their sha256 sums are in databases all over the world to prevent double application of the same change.
If we change the content of those files, the changes *will* be applied twice... (and possibly fail and thereby locking the user out of their upgrade path)

PR #3240, opened 2017-10-10
Removes transaction errors in database creation

ylavoie:
Fixes transaction handling in database creation.

### DIFF
--- a/lib/LedgerSMB/Database/Change.pm
+++ b/lib/LedgerSMB/Database/Change.pm
@@ -52,7 +52,7 @@ sub new {
     #This should have been done with '!' in LOADORDER
     if ( $self->content(1) =~ m/^BEGIN;\s*(.+)\s*COMMIT;$/gis ) {
         $self->{_content} = $1;
-        $self->{properties}->{no_transactions} = "";
+        $self->{properties}->{no_transactions} = '';
     }
     return $self;
 }
--- a/lib/LedgerSMB/Database/Change.pm
+++ b/lib/LedgerSMB/Database/Change.pm
@@ -178,7 +178,6 @@ Applies the current file to the db in the current dbh.
 
 sub apply {
     my ($self, $dbh) = @_;
-    my $need_commit = _need_commit($dbh);
     my $before = '';
     my $after;
     my $sha = $dbh->quote($self->sha);
@@ -199,26 +198,22 @@ sub apply {
     if ($no_transactions){
         $dbh->do($after);
         $after = '';
-        $dbh->commit if $need_commit;
+        $dbh->commit;
     }
     my $success = eval {
          $dbh->prepare($self->content_wrapped($before, $after))->execute();
     };
     die "$DBI::state: $DBI::errstr while applying $path"
         unless $success or $no_transactions;
-    $dbh->commit if $need_commit;
+    $dbh->commit;
     $dbh->prepare("
             INSERT INTO db_patch_log(when_applied, path, sha, sqlstate, error)
             VALUES(now(), $path, $sha, ?, ?)
     ")->execute($dbh->state, $dbh->errstr);
-    $dbh->commit if $need_commit;
+    $dbh->commit;
     return;
 }
 
-sub _need_commit{
-    my ($dbh) = @_;
-    return 1; # todo, detect existing transactions and autocommit status
-}
 =head1 Package Functions
 
 =head2 init($dbh)
@@ -245,6 +240,7 @@ sub init {
     );
     ')->execute();
     die "$DBI::state: $DBI::errstr" unless $success;
+    $dbh->commit;
 
     return 1;
 }
@@ -256,14 +252,13 @@ Returns true if the tracking system needs to be initialized
 =cut
 
 sub needs_init {
-    my ($dbh) = @_;
-    local $@ = undef;
-    my $rows = eval { $dbh->prepare(
-       'select 1 from db_patches'
-    )->execute(); };
-    $dbh->rollback;
-    return 0 if $rows;
-    return 1;
+    my $dbh = pop @_;
+    my $count = $dbh->prepare(q{
+        select relname from pg_class
+         where relname = 'db_patches'
+               and pg_table_is_visible(oid)
+    })->execute();
+    return !int($count);
 }
 
 =head1 TODO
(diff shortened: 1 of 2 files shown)


lib/LedgerSMB/Database/Change.pm
```diff
@@ -49,6 +49,11 @@ sub new {
     my $self = bless { _path => $path }, $package;
     my @prop_names = qw(no_transactions reload_subsequent);
     $self->{properties} = { map { $_ => $init_properties->{$_} } @prop_names };
+    #This should have been done with '!' in LOADORDER
```
ehuelsmann: Can you be more verbose as to why this can't be solved by adding `'!'` in the LOADORDER file? (or why it's a bad idea to do that)

lib/LedgerSMB/Database/Change.pm
```diff
@@ -149,6 +154,7 @@ Returns true if the current sha matches one that has been applied.
 
 sub is_applied {
     my ($self, $dbh) = @_;
+    $dbh->clone({ AutoCommit => 1}); # Clone to isolate ourselves
```
ehuelsmann: Why aren't you catching the cloned value? (and using that to operate the commits further?)

lib/LedgerSMB/Database/Change.pm
```diff
@@ -178,47 +184,38 @@ Applies the current file to the db in the current dbh.
...
+                   WHERE sha = $sha;
+              "
+              : "
+                   INSERT INTO db_patches (sha, path, last_updated)
+                   VALUES ($sha, $path, now());
+                ";
+    my $success = $dbh->prepare($self->content_wrapped($before, $after))->execute();
+    die "$DBI::state: $DBI::errstr while applying $path"
+        unless $success or $no_transactions;
+    $dbh->commit unless !$success || !$no_transactions;
```
ehuelsmann: why are you `die`-ing when the $no_transactions condition is true? or does the above mean `(die ... unless) or $no_transactions`? If the latter, that's too obfuscated for me. If not, apparently, we always need to die when `$no_transactions` is true?

lib/LedgerSMB/Database/Change.pm
```diff
@@ -256,14 +254,13 @@ Returns true if the tracking system needs to be initialized
...
 sub needs_init {
-    my ($dbh) = @_;
-    local $@ = undef;
-    my $rows = eval { $dbh->prepare(
-       'select 1 from db_patches'
-    )->execute(); };
-    $dbh->rollback;
-    return 0 if $rows;
-    return 1;
+    my $dbh = pop @_;
```
ehuelsmann: This is highly unusual idiom. Why not use `my ($self, $dbh) = @_;` ? or leave out `$self` if that's not part of the deal.

lib/LedgerSMB/Database/Loadorder.pm
```diff
@@ -69,7 +69,7 @@ sub scripts {
     my @scripts =
         map { $self->_process_script($_)}
         grep { $_ =~ /\S/ }
-        map { my $string = $_; $string =~ s/#.*$//; $string }
+        map { my $string = $_; $string =~ s/\s*#.*$//; $string }
```
ehuelsmann: Is this solving a problem related to this PR? it allows comments to be prepended with spaces, so it changes the semantics of the LOADORDER file. Is that required for this pr?

lib/LedgerSMB/Database/Change.pm
ylavoie: If we applied the patch or if we are in a transaction, don't die. The former is we succeeded, the latter is it will be rolled back later.

lib/LedgerSMB/Database/Loadorder.pm
ehuelsmann: Ok. But what makes it required to make this change?

ylavoie: Some errors are generated because trailing spaces aren't currently stripped and lines do not match.

lib/LedgerSMB/Database/Change.pm
```diff
@@ -155,8 +155,7 @@ Returns true if the current sha matches one that has been applied.
 sub is_applied {
     my ($self, $dbh) = @_;
     my $sha = $self->sha;
-    # Clone to isolate ourselves
-    my $sth = $dbh->clone({ AutoCommit => 1})->prepare(
+    my $sth = $dbh->prepare(
```
ehuelsmann: Now the clone() is gone, but we're not cloning the dbh anyhere now... I think we should do the entire loading of the change under a cloned db handle, because we want to change the handle for the duration of the change application to `AutoCommit => 1` (or `0`, whatever the change warrants)

ylavoie: No we don't need autocommit anymore, so I'm with your original questionning, why would a select need autocommit? It doesn't.

ehuelsmann:
From #ledgersmb:

> You misunderstood my original question/remark: my point was that if you have a bit of code which expects to run in `AutoCommit=>0` and another bits that explicitly start transactions, then you have built-in assumptions about the state of the `$dbh`.
> my point about cloning the `$dbh` is that that is a means to make sure it can be set to the correct state (without influencing the state the caller set).
> so, my point right now is that if you remove the `clone()` call (and don't have one to isolate the module's use from its callers), do we know for sure if that code works for both handles in `AutoCommit=>1` as well as `AutoCommit=>0` state?

ehuelsmann:
I have further researched the topic. My conclusion is that all commands sent in a single string are interpreted by PostgreSQL to be in a single transaction.

This has multiple implications:
1. Our current approach to wrap the script content in BEGIN and END is futile (a transaction is already active)
2. The code which expects to run transactionless (because prefixed in the loadorder by '!') isn't running transactionless.

I'm thinking this PR isn't the right direction to address these issues, is it?

ehuelsmann:
Oh. BTW, I got my knowledge from carefully reading both the DBD::Pg docs *as well as* the sources.

ylavoie:
We may had to rethink the transactionless steps but this definetely removes all errors in database creation.

ylavoie:
Closing as this is implemented better with #3291

PR #3244, opened 2017-10-18
Keep displayed information if user must be imported

ylavoie:
This is an interim PR to stop voiding user entered information after we discovered that the desired user must be imported.
A complete solution would be also to check if username already exists immediately after user entry, even though if this requires a server trip, and guide the user on import instead of creation.

### DIFF
--- a/UI/setup/new_user.html
+++ b/UI/setup/new_user.html
@@ -13,15 +13,13 @@ include_stylesheet=["setup/stylesheet.css"] ?>
    value = database
 } ?>
 <div class="form">
-<?lsmb- IF notice -?>
-<div class="notice"><?lsmb notice ?></div>
-<?lsmb- END # IF notice -?>
 <div class="input_row">
 <?lsmb INCLUDE input element_data = {
                             name  = 'username'
                             type  = 'text'
                            label  = text('Username')
                             class = 'username'
+                            value = username
                          required = 'true'
 } ?>
 </div>
@@ -49,7 +47,7 @@ text('Password') ?></label>
          label_pos = 1 ;
          ?>
     </div>
-    <div>
+    <div style="float:left">
       <?lsmb
          PROCESS input element_data = {
          label   = text("Import existing user")
@@ -60,6 +58,9 @@ text('Password') ?></label>
          }
          label_pos = 1 ; ?>
     </div>
+    <?lsmb- IF notice -?>
+    <div class="notice" style="float:left"><?lsmb notice ?></div>
+    <?lsmb- END # IF notice -?>
   </div>
   <div data-dojo-type="dijit/Tooltip"
        data-dojo-props="connectId:'pls-import-1',position:['below']">
@@ -89,6 +90,7 @@ text('Password') ?></label>
         text_attr = 'salutation'
        value_attr = 'id'
     default_blank = 1
+   default_values = [salutation_id]
              name = 'salutation_id'
             label = 'Salutation'
 } ?>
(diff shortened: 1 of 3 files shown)


UI/setup/new_user.html
```diff
@@ -49,7 +47,7 @@
          label_pos = 1 ;
          ?>
     </div>
-    <div>
+    <div style="float:left">
```
ehuelsmann: What problem does this change solve? Is it part of the problem that this PR solves? In what way?

ylavoie: Whenever we discover that the user already exists, we void everything and ask to toggle import and enter all over again. That specific line points the message to the import selection to make it obvious.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1106,6 +1106,16 @@ sub save_user {
...
+               hiddens  => {
+                                employeenumber => $request->{employeenumber},
+                                 salutation_id => $request->{salutation_id},
+                                         perms => $request->{perms},
+                                           ssn => $request->{ssn},
+                                    first_name => $request->{first_name},
+                                     last_name => $request->{last_name},
+                                           dob => $request->{dob},
+                                      username => $request->{username},
+                           },
```
ehuelsmann: What problem does this solve? Note that this isn't feeding the default values "back" into the editboxes.

ylavoie: It certainly is

ehuelsmann: I'm sorry, that's too terse for me. It is what? (or isn't what?)

UI/setup/new_user.html
ehuelsmann: The PR as a whole addresses your point

> Whenever we discover that the user already exists, we void everything and ask to toggle import and enter all over again. 

Agreed with that. I'm not sure what that has to do with explicitly this `style="float:left"` change you added here. I don't understand what you mean by

> That specific line points the message to the import selection to make it obvious.

Is there a visual effect that it achieves which is related to the PR?

ylavoie: Float the message right to the `Import user` selection.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: This does feed the default values in the dialog, it sets the hiddens and I use them in the form.

ehuelsmann: Ok. If that's the case, then where do you do that? I only see you use values of the variables `ssn`, `dob`, `last_name`, etc. (which definitely don't come from `hiddens`).

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1106,6 +1106,16 @@ sub save_user {
            my $template = LedgerSMB::Template->new_UI(
                $request,
                template => 'setup/new_user',
+               hiddens  => {
```
ehuelsmann: Looking at https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Template.pm#L13 and https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Template.pm#L115, `new_UI` doesn't take a `hiddens` argument.
This needs changing before it can be merged.

ylavoie: Unused hiddens removed.

PR #3245, opened 2017-10-22
* Fix mapping of constant numeric values

ehuelsmann:
Note: the reason this needs fixing is because the numeric
 constants (e.g. FC_TRANSACTION) aren't being recognized
 as such; instead, they are interpreted as inline bare
 litteral strings. By consequence, { FC_TRANSACTION => 1 }
 is being parsed into { 'FC_TRANSACTION' => 1 } instead of
 the intended { '1' => 1 }.

This commit fixes the above by mapping numbers to the
target values by using array-positions.


ehuelsmann:
Not accepted; replaced by PR #3247.

PR #3261, opened 2017-11-09
HTML fixes through better validation

ylavoie:
This PR was produced by using W3C for full validation of our HTML code, both static and produced from templates. We focus on the structured, so default template variables values based on their variable names or random integers were used.
The PR includes the required fixes to the HTML files. Although numerous, most HTML fixes are one liners.

### DIFF
--- a/UI/Reports/balance_sheet.html
+++ b/UI/Reports/balance_sheet.html
@@ -69,7 +69,7 @@ END ;
    END;
  ?>
 <tr <?lsmb class; IF report.rheads.ids.$row.props.account_number;
-                     'data-lsmb-account="';
+                     ' data-lsmb-account="';
                      report.rheads.ids.$row.props.account_number;
                      '"';
                   END ?>>
--- a/UI/Reports/co/filter_bm.html
+++ b/UI/Reports/co/filter_bm.html
@@ -12,7 +12,7 @@
   <tr>
     <th class="listtop">Balance y Mayor</th>
   </tr>
-  <tr style="height:5"></tr>
+  <tr style="height:5px"></tr>
   <tr>
     <td>
       <table>
--- a/UI/Reports/co/filter_bm.html
+++ b/UI/Reports/co/filter_bm.html
@@ -12,7 +12,7 @@
   <tr>
     <th class="listtop">Balance y Mayor</th>
   </tr>
-  <tr height="5"></tr>
+  <tr style="height:5"></tr>
   <tr>
     <td>
       <table>
--- a/UI/accounts/edit.html
+++ b/UI/accounts/edit.html
@@ -91,7 +91,7 @@
 <form data-dojo-type="lsmb/Form" method="post" action="account.pl">
 <?lsmb PROCESS input element_data = {
         value = 'A'
-        name = 'charttype2'
+        name = 'charttype'
         type = 'hidden'
 } ?>
 <div class="inputline" id="accno-line2">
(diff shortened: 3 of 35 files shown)


UI/Reports/balance_sheet.html
```diff
@@ -69,7 +69,7 @@
    END;
  ?>
 <tr <?lsmb class; IF report.rheads.ids.$row.props.account_number;
-                     'data-lsmb-account="';
+                     ' data-lsmb-account="';
```
ehuelsmann: Why is there an extra space here?

UI/Reports/co/filter_bm.html
```diff
@@ -12,7 +12,7 @@
   <tr>
     <th class="listtop">Balance y Mayor</th>
   </tr>
-  <tr height="5"></tr>
+  <tr style="height:5"></tr>
```
ehuelsmann: This isn't valid content for the style attribute as "5" isn't a valid height (change to "5px" to make it valid); same for all solar occurrences.

UI/accounts/edit.html
```diff
@@ -91,10 +91,10 @@
 <form data-dojo-type="lsmb/Form" method="post" action="account.pl">
 <?lsmb PROCESS input element_data = {
         value = 'A'
-        name = 'charttype'
+        name = 'charttype2'
```
ehuelsmann: Did you test this before PR-ing?

UI/timecards/timecard.html
```diff
@@ -47,7 +47,6 @@
   <td>
     <div data-dojo-type="lsmb/parts/PartDescription"
          data-dojo-props="channel:'/timecard/part-select/day'"
-         name="description"
```
ehuelsmann: Doesn't this break the invoice?

UI/Reports/balance_sheet.html
ylavoie: Because otherwise this tag get immediately concatenated to the previous and W3C complains about a missing separation between tags.

ehuelsmann: Ah. I'm completely missing the "class" output in my review. Thanks for the explanation.

UI/lib/ui-header.html
```diff
@@ -13,7 +13,7 @@
         <meta http-equiv="Pragma" content="no-cache" />
         <meta http-equiv="Expires" content="-1" />
         <?lsmb IF refresh -?>
-            <meta http-equiv="refresh" content="<?lsmb refresh.delay ?>;<?lsmb refresh.url ?>">
+            <meta http-equiv="refresh" content="<?lsmb refresh.delay ?>; <?lsmb refresh.url ?>">
```
ehuelsmann: Didn't we have to remove the meta http-equiv headers for html5?

xt/01-html-checks.t
```diff
@@ -5,8 +5,22 @@ use warnings;
...
+use Template;
+use Template::Config;
+use Template::Stash;
+use Template::VMethods;
+
+use WebService::Validator::HTML::W3C;
+
+$Template::Stash::PRIVATE = undef;  # Make sure we can access variables with a leading underscore
+
+my $validator_uri = $ENV{'VALIDATOR_URI'} // 'http://localhost:45000/w3c-validator/check';
```
ehuelsmann: I disagree here: this means the tests will fail when the validator isn't set up. Instead of using a default, we should skip.

PR #3284, opened 2017-11-23
* Fix #3266: invoice entry item and description numbers should match …

### LINKED ISSUE #3266: Invoice entry "item number" search is foobar'd
On the (sales) invoice entry screen the partnumber field (dropdown suffers from search issues.
- is case sensitive.
- always returns the first entry of a multiple match if a partial partnumber is entered followed by "enter key"
- Auto dropdown with filtered data is noticeably slow to display so the user may not even notice it before they press enter

To Test on demo 1.5
- new sales invoice
- pick a customer (I used Cubex)
- type `assy` hit enter
- note that there were no matches
- replace `assy` with `ASSY` hit enter
- note that `ASSY-00001` matches
- update
- for new item type `ASSY-00002` and hit enter
- note that there is such an item, but you temporarily get "no such item" before it correctly updates

- even worse , if there is ANY match, it always returns the FIRST match

However, if you type `ASSY` then wait, a filtered list will appear below the text entry box. It can take a noticeable amount of time to do so though.

Expected behavior would be to display a list of all `ASSY*` matches in a case insensitive result before allowing enter to move on to the next field.
At a minimum it should not select a default if there are multiple matches.

If it's a single match, then auto selecting that would be fine.

This causes a poor UX, and potentially makes the system unusable
It's also a regression

ehuelsmann:
…case insensitive

Note that next to matching case insensitive, we now support matching
substrings as well as doing full text search (i.e. match words, even if
in the wrong order).

### DIFF
--- a/sql/modules/Parts.sql
+++ b/sql/modules/Parts.sql
@@ -10,7 +10,7 @@ RETURNS SETOF parts AS
 $$
 SELECT *
   FROM parts
- WHERE ($1 IS NULL OR (partnumber ilike $1 || '%'))
+ WHERE ($1 IS NULL OR (partnumber ilike '%' || $1 || '%'))
        AND ($2 IS NULL
             OR description ilike '%' || $2 || '%'
             OR plainto_tsquery(get_default_lang()::regconfig, $2)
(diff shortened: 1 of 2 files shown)


ehuelsmann:
@sbts signed off on this PR in our online meetup to discuss exactly this subject earlier today.

PR #3320, opened 2017-12-12
Ignore asset report lines with 0 amounts

einhverfr:
This prevents unsuccessful depreciation runs from skewing depreciation.

### DIFF
--- a/sql/modules/Assets.sql
+++ b/sql/modules/Assets.sql
@@ -92,7 +92,7 @@ $$
                   coalesce(sum(l.amount), 0)),
             ai.department_id, ai.location_id
        FROM asset_item ai
-  LEFT JOIN asset_report_line l ON (l.asset_id = ai.id)
+  LEFT JOIN asset_report_line l ON (l.asset_id = ai.id and l.amount > 0)
   LEFT JOIN asset_report r ON (l.report_id = r.id)
       WHERE ai.id = ANY($1)
    GROUP BY ai.id, ai.start_depreciation, ai.purchase_date, ai.purchase_value,
@@ -127,7 +127,7 @@ $$
                   coalesce(sum(l.amount), 0)),
             ai.department_id, ai.location_id
        FROM asset_item ai
-  LEFT JOIN asset_report_line l ON (l.asset_id = ai.id)
+  LEFT JOIN asset_report_line l ON (l.asset_id = ai.id and l.amount > 0)
   LEFT JOIN asset_report r ON (l.report_id = r.id)
       WHERE ai.id = ANY($1)
    GROUP BY ai.id, ai.start_depreciation, ai.purchase_date, ai.purchase_value,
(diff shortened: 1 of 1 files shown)


ehuelsmann:
The question this raises with me is whether we use the best approach here: if we would calculate the difference between the "cumulative should-be" depreciation and the"cumulative thus-far" balances, anything that happens in-between is of secondary importance.
Wouldn't that be a better approach?

ehuelsmann:
Second question: I see that the depreciation is taken since the newest date of the group of assets as a whole until the target date. However, is possible to exclude specific assets from the depreciation report. That way one period can be skipped by excluding it on one run, but including it in the next, where the rest of the group does get included in both, right?
Is that behaviour intentional?

einhverfr:
On the second my reading of the code was that it was calculated per-asset, and that was why this was reported by an old client as having assets in a class which were wrongly excluded from depreciation in 1.3.

On the first, I am kind of concerned about corner cases in both sides and given the choices this seemed like the least intrusive fix.  I think a good argument can be made that you are right about the approach you suggest being better.  There are a number of corner cases on either side and I think there is more that needs to be done before your approach can work.

So let's go over what's wrong with the approach in this patch and then what's wrong with the other approach and why this is the only sort of fix I could see backporting.  Then I will mention why back porting is necessary.

My patch has a problem:  If you have straight-line annual monthly depreciation and you depreciate unevenly with dates, you get bad results.  I don't consider this to be a show-stopper because this would require breaking just about any ok practice in monthly adjustments.  But it is worth noting that if you start depreciation in one month then you depreciate in 6 week increments, you will get one month of depreciation for each 6 weeks and that is wrong.  But that happens without the patch too so it is not a regression.  However it would be fixed by the approach you suggest.

Now let's talk about what happens if you edit the amounts on an asset (we can agree this is not supported).  With the current approach if this happens near the beginning of the lifecycle, the effect is negligible, but you might get a small deviation in what seems to be depreciated when.  With your solution any adjustments for depreciation would take full effect on the next depreciation run.  If something was undervalued, the next depreciation run adds a sudden amount to compensate for all back amounts.  Neither one is the end of the world.  Both can be corrected with adjustments.  And these are expected to be approximate anyway (since you might dispose of an asset for more or less than you expected).  Now, my patch does not change that behaviour.  It only fixes an error that comes with a certain form of data entry error.

However I think to make a expected - to-date model work, we'd probably need to think through how and when to support updating amounts on these a little better.  In an ideal world we just don't support updating amounts at all.  But that is a far more intrusive change.

The change was made because a large-scale user of LedgerSMB for whom I had originally built the fixed asset system had an item they could not depreciate and it was not clear why (they are on 1.3 RC2 still).  The item had been erroneously entered under the wrong asset class, then attempted to have depreciation entered on the wrong date.    Then the asset class was fixed but the asset depreciation entry was not.  The result was that every attempt to depreciate the item was left at 0 and as troubleshooting showed, pushed out the depreciation start date.  Deleting all 0-amount depreciation entries solved the problem.

If more people use depreciation they will run into this and currently it is very hard to see what is going on.  So I think a reasonable, non-invasive fix needs to be made to stable branches.  And I think that we can discuss whether we need a more radical solution in master.

ehuelsmann:
Yup. You're completely correct to my second concern: The max() is calculated on a per-asset basis. So, I'm going to agree to this fix as being the best available under the current approach.

As to my point regarding cumulative approach: I think we should indeed not have people "meddle" with an asset that's under an active depreciation approach. We probably want to document how people are supposed to handle changes in depreciation after depreciation has actually started. (Unwinding the asset and setting it up all fresh would probably be the answer, just need instructions on the 'how'...)

PR #3321, opened 2017-12-13
Protect against unset csv import

ylavoie:
Do not generate errors if a csv file is not available for import or if none is selected.

### DIFF
--- a/lib/LedgerSMB/Scripts/recon.pm
+++ b/lib/LedgerSMB/Scripts/recon.pm
@@ -204,7 +204,7 @@ sub _display_report {
     $recon->unapproved_checks;
 
     my $contents = '';
-    {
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Scripts/recon.pm
```diff
@@ -204,15 +204,15 @@ sub _display_report {
     $recon->unapproved_checks;
 
     my $contents = '';
-    {
+    if ($request->{csv_file}) {
```
ehuelsmann: Are you sure this field is even ever defined? Did you try uploading a csv before submitting this PR?

I'm asking because I think it won't be defined, even when you submit a csv.

ylavoie:
Fixed by #3327.

PR #3323, opened 2017-12-13
Correct padding inheritance

ylavoie:
Make sure padding adjustments on login fields do not propagate to child classes.


ehuelsmann:
Which problem (presumably in the layout?) Does this PR fix? I mean, so far the login screen doesn't stand out as problematic?

PR #3325, opened 2017-12-13
Allow custom base url for Is_LSMB_running.sh test

nick-prater:
Previously Is_LSMB_running.sh used a hard-coded url for ledgersmb,
meaning the Is_LSMB_running.sh test failed when a different port or url
was used, such as in the Docker ledgersmb-dev-lsmb container environment.

This change keeps the default localhost:5001 location, but allows a
different base url to be specified via the LSMB_BASE_URL environment
variable, as used in the standard lsmb Docker containers.

### DIFF
--- a/utils/test/Is_LSMB_running.sh
+++ b/utils/test/Is_LSMB_running.sh
@@ -13,6 +13,20 @@ if [[ -z $EX_NOHOST ]]; then
     exit 99;
 fi
 
+
+# base url for lsmb defaults to http://localhost:5001
+# this can be overriden by setting environment variable $LSMB_BASE_URL
+lsmb_base_url=${LSMB_BASE_URL:-http://localhost:5001}
+
+# We check Starman is running on a specific port
+url_without_port=${lsmb_base_url%:[[:digit:]]*}
+lsmb_port=${lsmb_base_url#$url_without_port}
+lsmb_port=${lsmb_port#:}  # strip colon
+
+# Default to port 5001
+lsmb_port=${lsmb_port:-5001}
+
+
 # You can add to either of these two variables to skip this test during travis setup.
 # If it's skipped during this early run, it should also be run later in xt/60 which will never be skipped.
 Repo_early_skip_list+=('https://github.com/ylavoie/LedgerSMB.git');
@@ -55,8 +69,10 @@ HELP() {
 	                     and 't/data/Is_LSMB_running.html' may differ by
 	                   This should always be a number greater than 1 to allow for
 	                     variations in the list of available db admin users
-	
-	
+
+	ENVIRONMENT VARIABLES
+	    LSMB_BASE_URL : Sets the base url used to check lsmb
+	                    default: http://localhost:5001
 	EOF
     exit $E;
 }
@@ -116,7 +132,7 @@ WaitForPlackup() {
     local httpdrunning=false;
     while (( i++ < 100 )); do # wait up to 10 seconds for plack or starman process to start
         pgrep -f 'plackup' >/dev/null && { processrunning=true; echo "plackup started after $i * 0.1 seconds"; break; }
-        pgrep -f 'starman.*8080' >/dev/null && { processrunning=true; echo "starman started after $i * 0.1 seconds"; break; }
+        pgrep -f "starman.*$lsmb_port" >/dev/null && { processrunning=true; echo "starman started after $i * 0.1 seconds"; break; }
         sleep 0.1;
     done
     if ! $processrunning; then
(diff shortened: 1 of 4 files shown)


utils/test/Is_LSMB_running.sh
```diff
@@ -13,6 +13,20 @@ if [[ -z $EX_NOHOST ]]; then
...
 
+
+# base url for lsmb defaults to http://localhost:5001
+# this can be overriden by setting environment variable $LSMB_BASE_URL
+lsmb_base_url=${LSMB_BASE_URL:-http://localhost:5001}
+
+# We check Starman is running on a specific port
+url_without_port=${lsmb_base_url%:[[:digit:]]*}
+lsmb_port=${lsmb_base_url#$url_without_port}
+lsmb_port=${lsmb_port#:}  # strip colon
```
sbts: If you insert the colon before the $ on line 23 it will work better and line 24 can be dropped

utils/test/Is_LSMB_running.sh
```diff
@@ -13,6 +13,20 @@ if [[ -z $EX_NOHOST ]]; then
...
+
+# Default to port 5001
+lsmb_port=${lsmb_port:-5001}
```
sbts: We should also handle the Travis port variable here. Unless thats been replaced replacer since I last checked

utils/test/Is_LSMB_running.sh
```diff
@@ -116,7 +132,7 @@ WaitForPlackup() {
     local httpdrunning=false;
     while (( i++ < 100 )); do # wait up to 10 seconds for plack or starman process to start
         pgrep -f 'plackup' >/dev/null && { processrunning=true; echo "plackup started after $i * 0.1 seconds"; break; }
-        pgrep -f 'starman.*8080' >/dev/null && { processrunning=true; echo "starman started after $i * 0.1 seconds"; break; }
+        pgrep -f "starman.*$lsmb_port" >/dev/null && { processrunning=true; echo "starman started after $i * 0.1 seconds"; break; }
```
sbts: Nice catch. That hard coded 8080 should have been long gone

utils/test/Is_LSMB_running.sh
sbts: Sorry, that's not clear, and puts the : at the wrong end of the variable.
An Example is probably better and this one is tested to be correct
lsmb_port=${lsmb_base_url#${url_without_port}:}

nick-prater: I'm grateful for the feedback on this, as bash string matching is a foreign language for me - I've dusted off the phrase book!

But, the change you suggest breaks if LSMB_BASE_URL omits the port, for example:
LSMB_BASE_URL=http://localhost

The version I coded, while uglier, uses the default port in that situation.

utils/test/Is_LSMB_running.sh
```diff
@@ -13,6 +13,20 @@ if [[ -z $EX_NOHOST ]]; then
     exit 99;
 fi
 
+
+# base url for lsmb defaults to http://localhost:5001
+# this can be overriden by setting environment variable $LSMB_BASE_URL
+lsmb_base_url=${LSMB_BASE_URL:-http://localhost:5001}
```
sbts: Unfortunately `$LSMB_BASE_URL` is actually NOT the correct variable to use.
That points at port 5001 which is the HTTPD port (and URL) for travis.yml
For the docker container it is correct though.

This test is specifically testing that STARMAN/PLACK is running and that is on port 5000 by default on travis.
For docker it's on 5762.

Currently there is no variable declared in travis.yml that defines the starman port.

I believe we need to declare a new `$STARMAN_BASE_URL` in both `travis.yml` and the docker container `start,sh` that points to the starman process in both cases, leaving `$LSMB_BASE_URL` to point at the httpd

nick-prater: I propose to implement a new environment variable $PSGI_BASE_URL tomorrow (will sleep on it tonight).

When testing using the ledgersmb-dev-lsmb docker container, $PSGI_BASE_URL will be the same as $LSMB_BASE_URL - both will point to Starman.

When testing using Travis, $PSGI_BASE_URL will point to plack, while $LSMB_BASE_URL will point to the nginx proxy.

Positive feedback received in chat, but happy to hear any other thoughts:
https://matrix.to/#/!qyoLumPqusaXqFJNyK:matrix.org/$1513196388405391qbWsS:matrix.org

ehuelsmann: Agreed with the proposal

PR #3326, opened 2017-12-14
Remove redundant $POD_TESTING option from Travis config

nick-prater:
Commit 7de4dbaa44b867254cc5669ea1c2d2b57b50bb16 made the modules
required for pod testing part of the standard "develop" dependencies.
There is therefore no purpose to the $POD_TESTING option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ matrix:
           packages:
             - nginx postgresql-9.6-pgtap
       env:
+        - POD_TESTING=1
     - perl: '5.24'
       addons:
         postgresql: '9.5'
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -28,7 +28,6 @@ matrix:
           packages:
             - nginx postgresql-9.6-pgtap
       env:
-        - POD_TESTING=1
```
ehuelsmann: The pod tests act on this value too. I don't think it can be removed.

nick-prater: I'm proposing to remove the check from the two places this variable is used, as per latest commit to this PR. I think it simplifies code and it's better that all checks are consistently applied, given that these pod tests are not resource heavy.

nick-prater:
Looking at the Travis test results, the issue seems to be that test xt/08-pod-coverage.t doesn't work on Perl < 5.20. I guess that's why the option to disable pod testing was added.

At the very least I'll add a comment to that effect, but I'm wondering why it's broken.

PR #3328, opened 2017-12-16
Uniformize SQL-Ledger migration

ylavoie:
This PR aim to show only significant differences between migration from SL 2.8 or 3.0 to LedgerSMB. In order to do so:
1- The schema is now a parameter, so code is more generic
2- Many inconsistencies between the 3 versions were solved.


ehuelsmann:
This looks like a great improvement! I can see one more step to be taken, though: can we make the differences conditional using `<?lsmb IF ?>` statements and use that to de-duplicate the content completely? Then we don't need to keep them in sync at all (and we can replace 2 of the three files by a symlink to the third).

PR #3330, opened 2017-12-17
locale dated setup

ehuelsmann:
Closes #2739

### DIFF
--- a/UI/js-src/lsmb/DateTextBox.js
+++ b/UI/js-src/lsmb/DateTextBox.js
@@ -48,7 +48,6 @@ define([
                params.placeholder =
                   params.placeholder.replace(/M/g, 'm').replace(/y/g, 'yy');
             }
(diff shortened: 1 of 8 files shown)


UI/js-src/lsmb/DateTextBox.js
```diff
@@ -16,15 +17,37 @@ define([
...
+                       || bundle["timeFormat-"+formatLength];
+               }
+               else {
+                  params.placeholder = params.constraints.datePattern
+                       || bundle["dateFormat-"+formatLength];
+               }
+               params.placeholder =
+                  params.placeholder.replace(/M/g, 'm').replace(/y/g, 'yy');
+            }
             this.inherited(arguments);
```
ylavoie: As per documentation, constructors must not call inherited. See https://dojotoolkit.org/reference-guide/1.8/dojo/_base/declare.html#id8

PR #3339, opened 2017-12-21
Simplify new user handling

ylavoie:

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -1024,6 +1024,39 @@ sub skip_coa {
 }
 
 
+=item _render_user
+
+Renders the new user screen. Common functionality to both the
+select_coa and skip_coa functions.
+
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1112,31 +1125,14 @@ sub save_user {
...
-               {id => '0', label => $locale->text('Manage Users')},
-               {id => '1', label => $locale->text('Full Permissions')},
-           );
-           my $template = LedgerSMB::Template->new_UI(
-               $request,
-               template => 'setup/new_user',
-           );
-           $duplicate = $template->render_to_psgi($request);
+           return $duplicate
+               if $duplicate;
```
ehuelsmann: Duplicate will never be false, right?

ylavoie: Not unless something very wrong happens in the rendering.

PR #3350, opened 2017-12-26
Properly compute cleared balance ending

ylavoie:
Substracting a month doesn't give end of last month, specially when we are computing february balances, where 2 or 3 days of transactions won't be accounted for.


ehuelsmann:
I've been thinking about this PR. Instead of doing any kind of month/month-end calculations, how about passing in the report end-date? We'll be returned the reconciled balance as reconciled "up to" the end date of the report.
Basically, since we want reconciliations to be sequentially ordered in time, we don't want to use a date at all here: we simply need to include all reconciled transactions in the report. The closest we can get to using no date, is to use the end-date of the report.

ylavoie:
Using the end date will work only if the report hasn't been submitted, otherwise the opening amount will be wrong.

ehuelsmann:
needs backporting?

PR #3351, opened 2017-12-27
Add skip option for migration corrections

ylavoie:
- Add possibility to skip benign migration problems
- Optimize tooltips and button handling

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -599,7 +599,7 @@ sub _get_linked_accounts {
 }
 
 
-=item upgrade_settings
+=item upgrade_settigs
 
 =cut
 
@@ -782,7 +782,7 @@ sub _failed_check {
     my $hiddens = {
        check => $check->name,
 verify_check => md5_hex($check->test_query),
-    database => $request->{database},
+    database => $request->{database}
     };
     my @skip_keys = grep /^skip_/, keys %$request;
     $hiddens->{@skip_keys} = $request->{@skip_keys};
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -817,16 +817,16 @@ verify_check => md5_hex($check->test_query),
     $sth->finish();
 
     my $heading = { map { $_ => $_ } @{$check->display_cols} };
-    my %buttons_set = map { $_ => 1 } @{$check->buttons};
-    my $buttons;
+    my %buttons = map { $_ => 1 } @{$check->buttons};
+    my $enabled_buttons;
     for (
         { value => 'fix_tests', label => 'Save and Retry', cond => defined($check->{columns})},
         { value => 'cancel',    label => 'Cancel',         cond => 1                         },
         { value => 'force',     label => 'Force',          cond => $check->{force_queries}   },
         { value => 'skip',      label => 'Skip',           cond => $check->skipable          }
     ) {
-        if ( $buttons_set{$_->{label}} && $_->{cond}) {
-            push @$buttons, {
+        if ( $buttons{$_->{label}} && $_->{cond}) {
+            push @$enabled_buttons, {
                  type => 'submit',
                  name => 'action',
                 value => $_->{value},
@@ -855,7 +855,7 @@ verify_check => md5_hex($check->test_query),
            columns            => $check->display_cols,
            rows               => $rows,
            hiddens            => $hiddens,
-           buttons            => $buttons,
+           buttons            => $enabled_buttons,
            include_stylesheet => 'setup/stylesheet.css',
     });
 }
@@ -918,7 +918,7 @@ sub fix_tests{
 
     for my $count (1 .. $request->{count}){
         my @values;
-        push @values, split(/,/,@{$check->id_columns})
+        push @values, @{$check->id_columns}
             if $id_displayed;
         for my $edit (@edits) {
           push @values, $request->{"${edit}_$count"};
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -28,6 +28,7 @@ use File::Temp;
 use HTTP::Status qw( HTTP_OK HTTP_UNAUTHORIZED );
 use List::Util qw( first );
 use Locale::Country;
+use MIME::Base64;
 use Try::Tiny;
 use Version::Compare;
 
@@ -791,8 +792,7 @@ verify_check => md5_hex($check->test_query),
     my $rows = [];
     while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
       my $count = 1+scalar(@$rows);
-      #TODO: A '' will fail with invalid input syntax in fix_tests - YL
-      my $id = join(',',map { $row->{$_} // '' } @{$check->id_columns});
+
       for my $column (@{$check->columns // []}) {
         my $selectable_value = $selectable_values{$column};
         my $name = $column . '_' . $count;
@@ -812,7 +812,8 @@ verify_check => md5_hex($check->test_query),
                    size => 15,
           } };
       };
-      $hiddens->{"id_$count"} = $id;
+      #TODO: A '' will fail with invalid input syntax in fix_tests - YL
+      $hiddens->{"id_$count"} = join(',',map { MIME::Base64::encode(($row->{$_} // ''), '')} @{$check->id_columns});
       push @$rows, $row;
     }
     $hiddens->{count} = scalar(@$rows);
@@ -925,7 +926,7 @@ sub fix_tests{
         for my $edit (@edits) {
           push @values, $request->{"${edit}_$count"};
         }
-        push @values, split(/,/,$request->{"id_$count"})
+        push @values, map { MIME::Base64::decode($_)} split(/,/,$request->{"id_$count"})
            if ! $check->{insert};
 
         my $rv = $sth->execute(@values) ||
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -26,6 +26,7 @@ use warnings;
 use Digest::MD5 qw(md5_hex);
 use File::Temp;
 use HTTP::Status qw( HTTP_OK HTTP_UNAUTHORIZED );
+use List::Util qw( first );
 use Locale::Country;
 use Try::Tiny;
 use Version::Compare;
@@ -1436,13 +1437,11 @@ Force work.  Forgets unmatching tests, applies a curing statement and move on.
 sub force{
     my ($request) = @_;
     my $database = _init_db($request);
-    my $dbinfo = $database->get_info();
 
-    my %checks = map { $_->name => $_ } @{$info_applicable_for_upgrade{$dbinfo}};
-    my $check = $checks{$request->{check}};
-    my $force_queries = $check->{force_queries};
+    my $test = first { $_->name eq $request->{check} }
+                    LedgerSMB::Upgrade_Tests->get_tests();
 
-    for my $force_query ( @$force_queries ) {
+    for my $force_query ( @{$test->{force_queries}}) {
         my $dbh = $request->{dbh};
         $dbh->do($force_query);
         $dbh->commit;
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -800,6 +800,7 @@ verify_check => md5_hex($check->test_query),
            ( defined $selectable_value && @$selectable_value )
            ? { select => {
                    name => $name,
+                   default_values => $row->{$column} // '',
                    id => $count,
                    options => $selectable_value,
                    default_blank => ( 1 != @$selectable_value )
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -735,7 +735,7 @@ sub upgrade {
     my $locale = $request->{_locale};
 
     for my $check (_applicable_upgrade_tests($dbinfo)) {
-        next if $check->skip ne 'Disabled'
+        next if $check->skipable
              && defined $request->{"skip_$request->{check}"}
              && $request->{"skip_$request->{check}"} eq 'On';
         my $sth = $request->{dbh}->prepare($check->test_query);
@@ -786,7 +786,6 @@ verify_check => md5_hex($check->test_query),
     };
     my @skip_keys = grep /^skip_/, keys %$request;
     $hiddens->{@skip_keys} = $request->{@skip_keys};
-    warn np $hiddens;
 
     my $rows = [];
     while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
@@ -824,7 +823,7 @@ verify_check => md5_hex($check->test_query),
         { value => 'fix_tests', label => 'Save and Retry', cond => defined($check->{columns})},
         { value => 'cancel',    label => 'Cancel',         cond => 1                         },
         { value => 'force',     label => 'Force',          cond => $check->{force_queries}   },
-        { value => 'skip',      label => 'Skip',           cond => $check->skip eq 'Off'     }
+        { value => 'skip',      label => 'Skip',           cond => $check->skipable          }
     ) {
         if ( $buttons_set{$_->{label}} && $_->{cond}) {
             push @$buttons, {
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -599,7 +599,7 @@ sub _get_linked_accounts {
 }
 
 
-=item upgrade_settigs
+=item upgrade_settings
 
 =cut
 
@@ -735,9 +735,12 @@ sub upgrade {
     my $locale = $request->{_locale};
 
     for my $check (_applicable_upgrade_tests($dbinfo)) {
+        next if $check->skip ne 'Disabled'
+             && defined $request->{"skip_$request->{check}"}
+             && $request->{"skip_$request->{check}"} eq 'On';
         my $sth = $request->{dbh}->prepare($check->test_query);
         $sth->execute()
-            or die 'Failed to execute pre-migration check ' . $check->name;
+            or die 'Failed to execute pre-migration check ' . $check->name . ', ' . $sth->errstr;
         if ($sth->rows > 0){ # Check failed --CT
              return _failed_check($request, $check, $sth);
         }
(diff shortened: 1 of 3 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -702,6 +702,8 @@ sub _applicable_upgrade_tests {
                   LedgerSMB::Upgrade_Tests->get_tests;
 }
 
+my @checks;
```
ehuelsmann: Our requests shouldn't depend on global state (I'm going through a lot of work to *eliminate* global state, such as App_State). Is there a reason you declare the checks as global?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -44,11 +44,8 @@ Returns the test object with the name.
 
 sub get_by_name {
     my ($self, $name) = @_;
-    my @tests = $self->_get_tests;
-    for my $test (@tests){
-       return $test if $test->name eq $name;
-    }
-    return;
+    my %tests = map { $_->name => $_ } $self->_get_tests;
+    return $tests{$name};
```
ehuelsmann: Why does this need to change? I mean, the old version, even if optimizing a non-bottleneck, wasn't exactly wrong...

Same functionality as the old version, no more values iterated through than required and nicely short:

```perl
use List::Util qw( first );

sub get_by_name {
   my ($self, $name) = @_;

   return first { $_->name eq $name } $self->_get_tests;
}
```

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -122,6 +119,20 @@ C<text> is the textual value to be presented in the UI.
 
 has selectable_values => (is => 'ro', isa => 'HashRef', required => 0);
 
+sub _displayed_key {
```
ehuelsmann: The name suggests this returns displayed keys. However, from reading the code, that's not what it seems to do.

Do we really need this check? If so, maybe add a bit of documentation using `#` comments why this check is invaluable and what the exact check is achieving?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -122,6 +119,20 @@ C<text> is the textual value to be presented in the UI.
...
+    my ($self,@keys) = @_;
+    for my $key (@keys) {
+        die "'$key' not displayed in test '$self->{name}'"
+            if not grep( $key, $self->{display_cols});
+    }
+};
+
+after 'selectable_values' => sub {
+    my $self = shift;
+    $self->_displayed_key(keys %{$self->{selectable_values}})
```
ehuelsmann: Do we need to run this check on *every access* to `selectable_values`? or are you actually only trying to check it when being set?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -150,15 +161,16 @@ Repair query key to set the values if we can repair
 
 =cut
 
-has id_where => (is => 'ro', isa => 'Str', required => 0, default => 'id');
+has id_where => (is => 'ro', isa => 'Str', required => 0, default => '');
 
-=item id_column
+=item id_columns
```
ehuelsmann: You haven't changed all references to `id_column` to the new name (there's one left, as far as I can tell)

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -168,6 +180,12 @@ Repair query columns to run once per result
 
 has columns => (is => 'ro', isa => 'ArrayRef[Str]', required => 0);
 
+after 'columns' => sub {
```
ehuelsmann: Same as with the selectable values:

1. do we really need the check?
2. and if we do: do we really need to do the check on every attribute access?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -214,13 +231,37 @@ Tooltip for each button
...
-        return {
-            'Save and Retry' => marktext('Save the fixes provided and attempt to continue migration'),
-            'Cancel' => marktext('Cancel the <b>migration</b>')
-    }},
-    required => 0);
+    isa => 'Maybe[HashRef[Str]]', required => 0,
+    default => undef,   # Force initializer call
+    initializer => sub {
+        my ( $self, $value, $writer_sub_ref, $attribute_meta ) = @_;
+        for ($value) {
```
ehuelsmann: `$value` is at best an array reference, but definitely not an array.

Please use a local value for iterating the array and explicitly iterate the array:

```perl
for my $btn (@$value) { };
```

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -214,13 +231,37 @@ Tooltip for each button
...
+            die LedgerSMB::I18N::text(q(No button '[_1]' in test '[_2]'),$_,$self->{name})
```
ehuelsmann: Is this error detecting a programming error or providing the user with valuable feedback?

In case of the former, what are the chances of a true user ever seeing it? If the answer to that is "nil" then I don't think we should bother our translators with the need to translate it.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -214,13 +231,37 @@ Tooltip for each button
...
+        }
+        my %defaults = ('Save and Retry' => marktext('Save the fixes provided and attempt to continue migration'),
+                                'Cancel' => marktext('Cancel the <b>whole migration</b>'));
+        for (keys %defaults) {
+            $value->{$_} = $defaults{$_}
+                if not $value->{$_};
+        }
+        $writer_sub_ref->($value);
+    }
+);
```
ehuelsmann: The initializer feels over-engineered. How many more confirmations are we going to be seeing? Can we simply go for listing the confirmations that make sense for a specific test, deferring rendering of the buttons to the test-driver, supporting save(and retry), cancel and skip for now?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -214,13 +231,37 @@ Tooltip for each button
...
+
+=item skip
+
+Skip this test
+
+=cut
+
+enum SkipValues => [qw/ Disabled Off On /];
+has skip => (is =>'ro', isa => 'SkipValues',
```
ehuelsmann: This doesn't feel right: why do we have a special attribute to determine the status of exactly one of possibly infinite buttons?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -896,41 +893,45 @@ sub fix_tests{
...
+        $query = "UPDATE $table SET $setters WHERE $where "
+               . join(' AND ',map {"$_ = ? "} @{$check->id_columns});
     }
     my $sth = $dbh->prepare($query);
 
     for my $count (1 .. $request->{count}){
-        my $id = $request->{"id_$count"};
         my @values;
-        push @values, $id
+        push @values, split(/,/,@{$check->id_columns})
```
ehuelsmann: What's happening here? I thought the `id_columns` are an arrayref? Split is an operation that works on a string, not an array?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -896,41 +893,45 @@ sub fix_tests{
...
-        push @values, $request->{"id_$count"}
+        push @values, split(/,/,$request->{"id_$count"})
            if ! $check->{insert};
 
-        $sth->execute(@values) ||
+        my $rv = $sth->execute(@values) ||
             $request->error($sth->errstr);
+        #TODO: We should abort gracefully
+        die "$rv rows affected instead of only 1"
+            if $rv > 1;
```
ehuelsmann: Nicely defensive. I like it: it's not our own data, after all.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1436,9 +1437,11 @@ Force work.  Forgets unmatching tests, applies a curing statement and move on.
 sub force{
```
ehuelsmann: I understand the Skip, Save and Cancel buttons, but what's the Force button for? I mean, we can't force a square peg in a round hole... If the schema forbids data to be stored, no amount of Force is going to help...?

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Ah! List::Util is core. So we won't pay for another cpan module.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Renamed. Selectable values must be displayed otherwise test can't succeed.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: It should be in the builder, not on every access.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Fixed

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Tests can be skippable. Simplified and renamed

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Sometimes, the peg problem is not the shape but the color. See the `no_businesses` test for an example. Either we force the user to go back in SL to define the missing business or we force remove the references to the missings.

ehuelsmann: Ok. I see now. So, `force` applies a query which looses the (non-conforming) data. That's clear now. Thanks for the `no_businesses` reference.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,66 @@ sub _failed_check {
...
       push @$rows, $row;
-      my $count = scalar(@$rows);
-      $hiddens->{"id_$count"} = $row->{$check->id_column};
     }
     $hiddens->{count} = scalar(@$rows);
     $sth->finish();
 
     my $heading = { map { $_ => $_ } @{$check->display_cols} };
-    my %buttons = map { $_ => 1 } @{$check->buttons};
+    my %buttons_set = map { $_ => 1 } @{$check->buttons};
```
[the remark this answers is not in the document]
ylavoie: Fixed

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Fixed

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
     my $hiddens = {
        check => $check->name,
 verify_check => md5_hex($check->test_query),
-    database => $request->{database}
+    database => $request->{database},
```
ehuelsmann: If there's a style issue in the file (missing commas on last hash key/value pair), please change it in a separate commit. If there's no such style issue, why did this need to change in an otherwise unrelated change?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
...
     };
+    my @skip_keys = grep /^skip_/, keys %$request;
+    $hiddens->{@skip_keys} = $request->{@skip_keys};
```
ehuelsmann: hash slices. Nice!

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
...
 
     my $rows = [];
     while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
-      my $id = $row->{$check->{id_column}};
+      my $count = 1+scalar(@$rows);
+      #TODO: A '' will fail with invalid input syntax in fix_tests - YL
```
ehuelsmann: This comment is too terse for me. What is the problem you're trying to warn the reader for?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
...
-    database => $request->{database}
+    database => $request->{database},
     };
+    my @skip_keys = grep /^skip_/, keys %$request;
+    $hiddens->{@skip_keys} = $request->{@skip_keys};
 
     my $rows = [];
     while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
-      my $id = $row->{$check->{id_column}};
+      my $count = 1+scalar(@$rows);
```
ehuelsmann: `scalar(@$rows)` returns the exact number of rows in the `@$rows` array. Why do we need to add `1` to that?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
...
+        my $name = $column . '_' . $count;
         $row->{$column} =
            ( defined $selectable_value && @$selectable_value )
            ? { select => {
-                   name => $column . "_$id",
-                   id => $id,
+                   name => $name,
+                   id => $count,
                    options => $selectable_value,
                    default_blank => ( 1 != @$selectable_value )
```
ehuelsmann: Shouldn't the value of the dropdown be set, just as with the input field below, based on the value that can be found in `$row->{$column}`?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
...
       for my $column (@{$check->columns // []}) {
         my $selectable_value = $selectable_values{$column};
+        my $name = $column . '_' . $count;
         $row->{$column} =
            ( defined $selectable_value && @$selectable_value )
            ? { select => {
-                   name => $column . "_$id",
-                   id => $id,
+                   name => $name,
+                   id => $count,
```
ehuelsmann: The previous `$id` value was different for each row. The new `$count` value doesn't seem to be. Am I missing something?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
...
 
     my $rows = [];
     while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
-      my $id = $row->{$check->{id_column}};
+      my $count = 1+scalar(@$rows);
+      #TODO: A '' will fail with invalid input syntax in fix_tests - YL
+      my $id = join(',',map { $row->{$_} // '' } @{$check->id_columns});
       for my $column (@{$check->columns // []}) {
         my $selectable_value = $selectable_values{$column};
+        my $name = $column . '_' . $count;
```
ehuelsmann: This makes `$name` to be the same value for all cells in a single column, right?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,65 @@ sub _failed_check {
...
     };
+    my @skip_keys = grep /^skip_/, keys %$request;
+    $hiddens->{@skip_keys} = $request->{@skip_keys};
 
     my $rows = [];
     while (my $row = $sth->fetchrow_hashref('NAME_lc')) {
-      my $id = $row->{$check->{id_column}};
+      my $count = 1+scalar(@$rows);
+      #TODO: A '' will fail with invalid input syntax in fix_tests - YL
+      my $id = join(',',map { $row->{$_} // '' } @{$check->id_columns});
```
ehuelsmann: What happens when the value in the primary key is a character field containing a comma? (Or do we only have numeric p-keys so far?)

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -779,72 +782,66 @@ sub _failed_check {
...
+                           msg => $check->{tooltips}->{$_->{label}}
+                                ? $request->{_locale}->maketext($check->{tooltips}->{$_->{label}})
+                                : undef,
+                           position => 'above'
+                         },
+                 text => $request->{_locale}->maketext($_->{label}),
+                class => 'submit'
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ylavoie: Thanks. After the introduction of the last button, there really wasn't any other choice.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -896,41 +892,45 @@ sub fix_tests{
...
+    my $id_columns = join('|',@{$check->id_columns});
     my $id_displayed = $check->{insert}
-                and grep( /^$check->{id_column}$/, @{$check->display_cols} );
+                and grep( /^($id_columns)$/, @{$check->display_cols} );
 
     my $query;
     if ($check->{insert}) {
         my @_edits = @edits;
-        unshift @_edits, $check->{id_column} if $id_displayed;
+        unshift @_edits, @{$check->id_columns} if $id_displayed;
```
ehuelsmann: I'm unfamiliar with the UI, but why doesn't the UI provide us with these values? Isn't the key to the situation that the row is *new*, so the p-key isn't available in advance?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1436,9 +1436,11 @@ Force work.  Forgets unmatching tests, applies a curing statement and move on.
 sub force{
     my ($request) = @_;
     my $database = _init_db($request);
+    my $dbinfo = $database->get_info();
 
-    my %test = map { $_->name => $_ } LedgerSMB::Upgrade_Tests->get_tests();
-    my $force_queries = $test{$request->{check}}->{force_queries};
+    my %checks = map { $_->name => $_ } @{$info_applicable_for_upgrade{$dbinfo}};
```
ehuelsmann: I think this is wrong. You probably meant to call `applicable_for_upgrade()` instead of evaluating the hash `applicable_for_upgrade`.

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1436,9 +1436,11 @@ Force work.  Forgets unmatching tests, applies a curing statement and move on.
...
+    my $check = $checks{$request->{check}};
```
ehuelsmann: This statement and the one above can be done with `first()` from `List::Util` again.

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -150,15 +163,16 @@ Repair query key to set the values if we can repair
 
 =cut
 
-has id_where => (is => 'ro', isa => 'Str', required => 0, default => 'id');
+has id_where => (is => 'ro', isa => 'Str', required => 0, default => '');
```
ehuelsmann: Now that we support complex primary keys, what value does this attribute still have?

lib/LedgerSMB/Upgrade_Tests.pm
```diff
@@ -200,27 +214,69 @@ Enabled buttons
...
+
+=item BUILD
+
+Called after object building. Ideal to validate object fields for coherence
+
+=cut
+
+sub BUILD {
+    my $self = shift;
+    $self->_validate_displayed_key(keys %{$self->{selectable_values}})
```
ehuelsmann: This check is done at run time (when the user is upgrading). However, it's validating information that's statically available in this file. For example the `no_business_for_customer` check defines all the components required to run this check. So, even without having a database to migrate, we can just check the column-consistency when running our `prove` tests.
And once we have established our `prove` outcome to be correct, why would we still want to run this check during a migration?

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Because the rows are empty initialy and id starts at 1.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: ID of columns are SQL identifiers

lib/LedgerSMB/Scripts/setup.pm
ylavoie: No, `$count` changes everytime we push a line in `$rows`

lib/LedgerSMB/Scripts/setup.pm
ylavoie: See previous comment

lib/LedgerSMB/Scripts/setup.pm
ylavoie: We never did because if the data was presented to the user to fix, then the original value was wrong.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: Fixed

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: Moved to `xt/50-upgrade-tests.t`. Same for correlation between tooltips and buttons.

lib/LedgerSMB/Scripts/setup.pm
ylavoie: id_columns contains the columns id, not data. So, unless we want to protect ourselves in case of someone starts to use quoted ids, which we don't elsewhere, joining the id array with commas is totally acceptable.

ehuelsmann: Sure. But `map { $row->{$_} // '' }` maps the column names into the data for the current record. Which means you can get any content, including commas.

lib/LedgerSMB/Upgrade_Tests.pm
ylavoie: None. Either id_where or id_columns can be used. Both cannot be empty though.

ehuelsmann: I don't understand why we need `id_where` in case we already have `id_columns`

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -896,41 +895,45 @@ sub fix_tests{
...
+          push @values, $request->{"${edit}_$count"};
         }
-        push @values, $request->{"id_$count"}
+        push @values, map { MIME::Base64::decode($_)} split(/,/,$request->{"id_$count"})
            if ! $check->{insert};
 
-        $sth->execute(@values) ||
+        my $rv = $sth->execute(@values) ||
             $request->error($sth->errstr);
+        #TODO: We should abort gracefully
```
ehuelsmann: No need to `die()` here: simply return a PSGI triplet with a body indicating the problem.

PR #3353, opened 2017-12-28
Move comment close to function definition

ylavoie:


sbts:
Is there a reason the comment is still AFTER the function it describes?

ylavoie:
SQL comments are usually after function definition in our code. That one was lost far from its definition.

ehuelsmann:
The reason they're *after* is that the object to comment on doesn't exist before.

PR #3355, opened 2017-12-29
Fix #1960 - Remove pre Perl-5.14 `local $@` protection blocks

### LINKED ISSUE #1960: Remove pre-5.14 `local $@;` protection code blocks
For 1.6, lets drop pre-5.14 support (that is, 5.10 and 5.12) explicitly by dropping the code blocks which surround the localization of `$@` -- which in 5.10 and 5.12 (from a 'supported by us' perspective) worked differently than 5.14+.

nick-prater:
Perl 5.14 changed the handling of `$@` in eval blocks - previously there was a risk that `$@` would be clobbered elsewhere, affecting calling code - 'action at a distance'. See:
https://perldoc.perl.org/perl5140delta.html#Exception-Handling

We no longer support Perl < 5.14, so the code added to protect against this problem is no needed. This PR removes that unnecessary code.

### DIFF
--- a/lib/LedgerSMB/Database.pm
+++ b/lib/LedgerSMB/Database.pm
@@ -232,7 +232,6 @@ sub get_info {
     full_version => undef,
           status => undef,
     };
-    local $@ = undef;
 
     my $dbh = eval { $self->connect({PrintError => 0, AutoCommit => 0}) };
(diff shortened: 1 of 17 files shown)


nick-prater:
Unfortunately this seems to break on perl 5.14 and 5.16. I'll have to set up a development system with those versions to debug further.

PR #3372, opened 2018-01-01
Remove useless dblog file tests from xt/40-dbsetup.t

nick-prater:
xt/40-dbsetup.t attempts to check and clean up various dblog files
from a temporary directory, but these files are never created, hence
the checks and cleanup are useless and are removed by this patch.

This patch is in preparation to enable InputOutput::RequireCheckedOpen
Perl::Critic policy on the tests.

### DIFF
--- a/xt/40-dbsetup.t
+++ b/xt/40-dbsetup.t
@@ -25,7 +25,7 @@ for my $evar (qw(LSMB_NEW_DB LSMB_TEST_DB)){
 }
 
 if ($run_tests){
-        plan tests => 20;
+        plan tests => 19;
         $ENV{PGDATABASE} = $ENV{LSMB_NEW_DB};
 }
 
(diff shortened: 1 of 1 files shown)


xt/40-dbsetup.t
```diff
@@ -16,9 +16,7 @@ HINT:  Set LSMB_NEW_DB environment variable and try running again.');
 
 my $temp = $ENV{TEMP} || '/tmp/';
 my $run_tests = 1;
```
ehuelsmann: I think this variable can be eliminated too.

xt/40-dbsetup.t
```diff
@@ -16,9 +16,7 @@ HINT:  Set LSMB_NEW_DB environment variable and try running again.');
...
-for my $log (qw(dblog dblog_stderr dblog_stdout)){
-    unlink "$LedgerSMB::Sysconfig::tempdir/$log";
-}
+
 for my $evar (qw(LSMB_NEW_DB LSMB_TEST_DB)){
   if (!defined $ENV{$evar}){
       $run_tests = 0;
```
ehuelsmann: And this reference to it too. Because I think the `plan skip_all` below simply cancels execution.

xt/40-dbsetup.t
nick-prater: Agreed... to follow... one step at a time...

PR #3395, opened 2018-01-07
Add GIFI empty description test

ylavoie:
Add test to deal with empty SQL-Ledger GIFI descriptions

### DIFF
--- a/lib/LedgerSMB/Upgrade_Tests.pm
+++ b/lib/LedgerSMB/Upgrade_Tests.pm
@@ -507,6 +507,23 @@ push @tests, __PACKAGE__->new(
 
 
 push @tests, __PACKAGE__->new(
+   test_query => q{select accno,description from gifi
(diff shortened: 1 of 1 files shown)


ehuelsmann:
Hmm. I'm not sure I like this test: It's your opinion they should have a description, but there's nothing in the migration that depends on the descriptions being available. In other words: when they're empty, nothing fails to migrate.

sbts:
I agree that GIFI should have descriptions, however, I agree with @ehuelsmann that it's not really a pre-migration test.
Rather it's a Data Validation test that is probably best run on All Databases during a normal upgrade, and after a migration from an external source

ylavoie:
Migration will fail if there is a missing use gifi and another with a missing description

ehuelsmann:
Right. But the failure is due to missing code, not due to missing description.

PR #3398, opened 2018-01-08
Escape ampersands in URI

ylavoie:
All ampersands should be escaped to `&amp;` in URI to ensure valid syntax.


UI/Contact/divs/files.html
```diff
@@ -19,7 +19,7 @@
 <?lsmb
   FOREACH COL IN columns;
          IF COL.col_id == 'file_name';
-            COL.href_base='file.pl?action=get&file_class=4&id=';
+            COL.href_base='file.pl?action=get&amp;file_class=&amp;id=';
```
nick-prater: new line misses the `4` value for file_class

PR #3416, opened 2018-01-18
Optimize long queries by removing loop invariants and reordering table order by size

ylavoie:
By removing the loop invariants, making sure that the smallest tables are used first and adding a few indexes we get a tenfold speed improvement on the long queries.
This algorythm is O(n2) on a big table, thus we have to be carefull with it.
In order to reduce the time, I
- moved the computation of accno outside the loop and precomputed it at the beginning of the migration routine;
- same for the truncated dates, they were moved in an augmented acc_trans table and computed once;
- moved the type computation in acc_trans so it can show AR, AP or GL, like we do with transactions;
- the query order was modified to use the smallest tables in the beginning and the biggest at the end;
The query time dropped by an order of magnitude (10x)

### DIFF
--- a/sql/upgrade/sl3.0.sql
+++ b/sql/upgrade/sl3.0.sql
@@ -994,11 +994,6 @@ INSERT INTO invoice (id, trans_id, parts_id, description, qty, allocated,
             deliverydate, serialnumber
        FROM :slschema.invoice;
 
-ALTER TABLE :slschema.acc_trans ADD COLUMN lsmb_entry_id integer;
-
-update :slschema.acc_trans
-  set lsmb_entry_id = nextval('acc_trans_entry_id_seq');
-
 INSERT INTO acc_trans (entry_id, trans_id, chart_id, amount, transdate,
                        source, cleared, fx_transaction,
                        memo, approved, cleared_on, voucher_id, invoice_id)
--- a/sql/upgrade/sl3.0.sql
+++ b/sql/upgrade/sl3.0.sql
@@ -227,7 +227,7 @@ create or replace function pg_temp.reconciliation__account_list () returns setof
         coa.accno || ' ' || coa.description as name,
         coa.accno, coa.id as id
     FROM account coa
-         JOIN cr_coa_to_account cta ON cta.chart_id = coa.id
+    JOIN cr_coa_to_account cta ON cta.chart_id = coa.id
     ORDER BY coa.accno;
 $$ language sql;
 
@@ -394,10 +394,7 @@ $$
     SELECT sum(ac.amount) * CASE WHEN c.category in('A', 'E') THEN -1 ELSE 1 END
         FROM account c
         JOIN acc_trans ac ON (ac.chart_id = c.id)
-    JOIN (      SELECT id FROM ar WHERE approved
-          UNION SELECT id FROM ap WHERE approved
-          UNION SELECT id FROM gl WHERE approved
-          ) g ON g.id = ac.trans_id
+        JOIN transactions t ON t.id = ac.trans_id AND t.approved
     WHERE c.id = $1 AND cleared
       AND ac.approved IS true
       AND ac.transdate <= in_report_date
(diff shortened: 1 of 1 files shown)


sql/upgrade/sl3.0.sql
```diff
@@ -227,7 +227,7 @@ create or replace function pg_temp.reconciliation__account_list () returns setof
         coa.accno || ' ' || coa.description as name,
         coa.accno, coa.id as id
     FROM account coa
-         JOIN cr_coa_to_account cta ON cta.chart_id = coa.id
+    JOIN cr_coa_to_account cta ON cta.chart_id = coa.id
```
ehuelsmann: To what extent does this make a performance difference?

sql/upgrade/sl3.0.sql
```diff
@@ -973,22 +994,16 @@ INSERT INTO invoice (id, trans_id, parts_id, description, qty, allocated,
...
                        memo, approved, cleared_on, voucher_id, invoice_id)
- SELECT lsmb_entry_id, acc_trans.trans_id, (select id
-                    from account
-                   where accno = (select accno
-                                    from :slschema.chart
-                                   where chart.id = acc_trans.chart_id)),
-                                    amount, transdate, source,
-        CASE WHEN cleared IS NOT NULL THEN TRUE ELSE FALSE END, fx_transaction,
-        memo, approved, cleared, vr_id, invoice.id
+ SELECT lsmb_entry_id, acc_trans.trans_id,
```
ehuelsmann: You stopped adding the `lsmb_entry_id` by removing the fragment above. Did you test this?

sql/upgrade/sl3.0.sql
```diff
@@ -1119,25 +1134,16 @@ INSERT INTO cr_report(chart_id, their_total, submitted, end_date, updated, enter
...
-        WITH types AS ( SELECT id,'AP' AS type FROM :slschema.ap
-                  UNION SELECT id,'AR'         FROM :slschema.ar
-                  UNION SELECT id,'GL'         FROM :slschema.gl)
-        SELECT DISTINCT ac.trans_id, types.type
-        FROM :slschema.acc_trans ac
-        JOIN types ON ac.trans_id = types.id
-        ORDER BY ac.trans_id
-    ) n ON n.trans_id = a.trans_id
-    ORDER BY post_date,cr.id,n.type,a.source ASC NULLS LAST,a.amount
+SELECT cr.id::INT, cr.end_date, a.source, a.type, a.cleared::TIMESTAMP, a.amount::NUMERIC, a.transdate AS post_date, a.lsmb_entry_id
```
ehuelsmann: The equivalence between the old and new chunks needs some explanation in the commit message, because I'm not seeing it yet...

sql/upgrade/sl3.0.sql
```diff
@@ -1054,7 +1069,7 @@ BEGIN
 END;
 $$ LANGUAGE PLPGSQL;
 
-SELECT payment_migrate(p.id, p.trans_id, cast(p.exchangerate as numeric), p.paymentmethod_id)
+SELECT pg_temp.payment_migrate(p.id, p.trans_id, cast(p.exchangerate as numeric), p.paymentmethod_id)
```
ehuelsmann: Has nothing to do with performance fixes in the migration. But nice catch. I would have put it in its own commit because it's thematically different from the other changes.

sql/upgrade/sl3.0.sql
ylavoie: Restored

sql/upgrade/sl3.0.sql
ylavoie: This specific fix has nothing to do but payment_migrate is the 3rd longuest query, accounting for almost 10% of migration time.

sql/upgrade/sl3.0.sql
ylavoie: This algorythm is O(n2) on a big table, thus we have to be carefull with it.
In order to reduce the time, I
- moved the computation of accno outside the loop and precomputed it at the beginning of the migration routine;
- same for the truncated dates, they were moved in an augmented `acc_trans` table and computed once;
- moved the type computation in `acc_trans` so it can show AR, AP or GL, like we do with `transactions`;
- the query order was modified to use the smallest tables in the beginning and the biggest at the end;
The query time dropped by an order of magnitude (10x)

ehuelsmann: Thanks for the explanation, but it belongs in the commit message, so it is retained even if we have to move from GitHub to another provider some day. This explanation would have made a great commit message.

PR #3417, opened 2018-01-18
Stop skipping all remaining test

ylavoie:
All remaining tests were skipped after a first skip.


ehuelsmann:
Why is this called "stop skipping all remaining tests"? Didn't the condition being changed fail to match *all* skip conditions?

ylavoie:
This fixes the problem we were having with Skip. After doing 1 skip, all remaining tests we also immediately skipped.

PR #3422, opened 2018-01-21
Check reconciliation balanceness

ylavoie:
Prevent submit of an unbalanceed report

### DIFF
--- a/sql/changes/1.6/constrain_cr_report_line_entries.sql
+++ b/sql/changes/1.6/constrain_cr_report_line_entries.sql
@@ -1,0 +1,1 @@
+CREATE UNIQUE INDEX cr_report_line_entry_idx ON cr_report_line(report_id,ledger_id);
--- a/sql/modules/Reconciliation.sql
+++ b/sql/modules/Reconciliation.sql
@@ -133,7 +133,6 @@ BEGIN
         RETURN FALSE;
     END IF;
 
-    --TODO: Why can't we cascade DELETE with a TRIGGER to prevent?
     DELETE FROM cr_report_line
      WHERE report_id = in_report_id
            AND report_id IN (SELECT id FROM cr_report
--- a/sql/modules/Reconciliation.sql
+++ b/sql/modules/Reconciliation.sql
@@ -553,7 +553,7 @@ $$
                               OR ac.fx_transaction IS TRUE)))
                 AND (ac.entry_id > coalesce(r.max_ac_id, 0))
         GROUP BY gl.ref, ac.source, ac.transdate,
-                ac.memo, ac.voucher_id, gl.table,
+                ac.memo, ac.voucher_id, gl.table, ac.trans_id,
                 case when gl.table = 'gl' then gl.id else 1 end
         HAVING count(rl.id) = 0;
 
--- a/sql/modules/Reconciliation.sql
+++ b/sql/modules/Reconciliation.sql
@@ -30,11 +30,17 @@ BEGIN;
 CREATE OR REPLACE FUNCTION reconciliation__submit_set(
         in_report_id int, in_line_ids int[]) RETURNS bool AS
 $$
+DECLARE
+        result NUMERIC;
 BEGIN
-        UPDATE cr_report set submitted = true where id = in_report_id;
-        PERFORM reconciliation__save_set(in_report_id, in_line_ids);
-
-        RETURN FOUND;
+        result = reconciliation__check_balanced(in_report_id, in_line_ids);
+        IF result = 0 THEN
+            UPDATE cr_report set submitted = true where id = in_report_id;
+            PERFORM reconciliation__save_set(in_report_id, in_line_ids);
+            RETURN FOUND;
+        ELSE
+            RAISE EXCEPTION 'Unbalanced report by %', result::NUMERIC(10,2);
+        END IF;
 END;
 $$ LANGUAGE PLPGSQL;
 
@@ -43,6 +49,19 @@ COMMENT ON FUNCTION reconciliation__submit_set(
 $$Submits a reconciliation report for approval.
 in_line_ids is used to specify which report lines are cleared, finalizing the
 report.$$;
+
+CREATE OR REPLACE FUNCTION reconciliation__check_balanced(in_report_id int,
+      in_line_ids int[]) RETURNS NUMERIC AS
+$$
+    SELECT balance FROM (
+      SELECT cr.end_date,COALESCE(SUM(crl.our_balance),0) - cr.their_total AS balance
+        FROM cr_report cr
+        LEFT JOIN cr_report_line crl ON cr.id = crl.report_id
+                                    AND crl.id = ANY(in_line_ids)
+       WHERE cr.id = in_report_id
+    GROUP BY cr.end_date,cr.their_total
+    ) a;
+$$ LANGUAGE SQL;
 
 CREATE OR REPLACE FUNCTION reconciliation__check(in_end_date date, in_chart_id int)
 RETURNS SETOF defaults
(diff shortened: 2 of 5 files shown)


sql/modules/Reconciliation.sql
```diff
@@ -93,6 +111,28 @@ $$Sets which lines of the report are cleared.$$;
...
+    JOIN acc_trans ac ON rl.ledger_id = ac.entry_id
+    WHERE report_id = in_report_id
+      AND cleared_on IS NOT NULL;
+
+    IF FOUND THEN
+        RAISE EXCEPTION 'Transactions already cleared on this report';
+        RETURN FALSE;
+    END IF;
+
+    --TODO: Why can't we cascade DELETE with a TRIGGER to prevent?
```
ehuelsmann: to prevent what?

ylavoie: Having to clean up here.

sql/modules/Reconciliation.sql
```diff
@@ -44,6 +50,18 @@ $$Submits a reconciliation report for approval.
 in_line_ids is used to specify which report lines are cleared, finalizing the
 report.$$;
 
+CREATE OR REPLACE FUNCTION reconciliation__check_balanced(in_report_id int
```
ehuelsmann: This function isn't followed by a `COMMENT ON ...` to document it.

sql/modules/Reconciliation.sql
```diff
@@ -44,6 +50,18 @@ $$Submits a reconciliation report for approval.
...
+      ) RETURNS NUMERIC AS
+$$
+    SELECT COALESCE((
+          SELECT SUM(crl.our_balance) - cr.their_total
```
ehuelsmann: Does a `0` return from this function mean "balanced"? If so, do you mean to return "Yes, it's balanced" when `SUM(crl.our_balance)` is `NULL` (thus leading the difference to be `NULL`)?

sql/modules/Reconciliation.sql
```diff
@@ -44,6 +50,18 @@ $$Submits a reconciliation report for approval.
...
+            FROM cr_report_line crl
+            JOIN cr_report cr ON cr.id = crl.report_id
```
ehuelsmann: if there is a report with no lines, this causes the function to indicate "balanced" (sum of zero lines is NULL, coalesced to zero). Is that the correct behaviour?

ylavoie: That was the intent. In the application, we want only the amount, in tests we want 0 and it is the actual of an empty ledger or report.

sql/modules/Reconciliation.sql
ehuelsmann: Still needs cleanup ...

sql/modules/Reconciliation.sql
```diff
@@ -520,6 +562,18 @@ $$
                 max_ac_id = (select max(entry_id) from acc_trans)
         where id = in_report_id;
 
+    -- How can we get already cleared transactions?
```
ehuelsmann: Please don't put the questions but the answers in the comments :-) !

ylavoie: We are testing (defensively?) for a condition that should never happen in the application. Cleared transactions can only arise from a reconciliation, yet the test data provide some.

sql/modules/Reconciliation.sql
```diff
@@ -93,6 +112,28 @@ $$Sets which lines of the report are cleared.$$;
...
+    -- Make sure that transactions present on this report do not have a
+    -- cleared_on date or they won't ever be accessible anymore.
+    PERFORM rl.id
+    FROM cr_report_line rl
+    JOIN acc_trans ac ON rl.ledger_id = ac.entry_id
+    WHERE report_id = in_report_id
+      AND (ac.cleared_on IS NOT NULL OR ac.cleared);
+
+    IF FOUND THEN
+        RAISE EXCEPTION 'Transactions already cleared on this report';
```
ehuelsmann: But... Shouldn't we prevent cleared transactions from being in the report to begin with?

sql/modules/Reconciliation.sql
ehuelsmann: I'm not sure about your reaction. Is that the response to the question in the code? If so, then shouldn't the response above be in the code?

ehuelsmann: also, if they can't occur in the application, maybe we should adjust the migration so that they can't happen in the migration either?

ylavoie: The problem isn't migration. Test data seems designed to provide an invalid situation, which we took for ok for many years.

sql/changes/1.6/constrain_cr_report_line_entries.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX cr_report_line_entry_idx ON cr_report_line(report_id,ledger_id);
```
ehuelsmann: Maybe we need 

```sql
CREATE UNIQUE INDEX cr_report_line_entry_idx ON cr_report_line(report_id,ledger_id) WHERE clear_date NOT NULL;
```

So that reports can be saved, with lines included in multiple reports. Just can't accept/post two reports which include the same line.

sql/modules/Reconciliation.sql
```diff
@@ -93,6 +112,28 @@ $$Sets which lines of the report are cleared.$$;
...
+BEGIN
+    PERFORM approved FROM cr_report
+    WHERE id = in_report_id AND approved;
+    IF FOUND THEN
+        RAISE EXCEPTION 'Cannot delete approved recon report';
+        RETURN FALSE;
+    END IF;
+
+    -- Make sure that transactions present on this report do not have a
+    -- cleared_on date or they won't ever be accessible anymore.
```
ehuelsmann: Of course the transactions are already cleared?! I mean, we're DELETING the report, right?

ehuelsmann: Or should it be impossible to delete a report which has at least 1 cleared transaction?

sql/modules/Reconciliation.sql
ehuelsmann: Ok. But the original remark remains: don't put questions, but answers in the code comments.

sql/modules/Reconciliation.sql
ylavoie: A transaction, not an entry line, gets cleared only when the reconciliation report is approved. If we ever get an already cleared transaction in a report that we are deleting, it won't ever be reconciliable again because we only select uncleared transactions for entry lines.

sql/changes/1.6/constrain_cr_report_line_entries.sql
ylavoie: Why should we ever want to allow duplication of a transaction in reconciliation reports? That is one of the source of the problems in reconciliations.

sql/modules/Reconciliation.sql
```diff
@@ -93,6 +112,27 @@ $$Sets which lines of the report are cleared.$$;
 CREATE OR REPLACE FUNCTION reconciliation__delete_my_report(in_report_id int)
 RETURNS BOOL AS
 $$
+BEGIN
+    PERFORM approved FROM cr_report
+    WHERE id = in_report_id AND approved;
+    IF FOUND THEN
```
ehuelsmann: Why do you introduce this? We have a trigger which blocks updates and deletes to approved reconciliation reports: `cr_report_block_changing_approved`.

sql/modules/Reconciliation.sql
ehuelsmann: I have no idea why you say the entire transaction will be cleared. If you look at https://github.com/ledgersmb/LedgerSMB/blob/master/sql/modules/Reconciliation.sql#L247-L248 you see that the `acc_trans` lines get set to cleared. Those are the acc_trans lines *included in the report*, which almost certainly means "not the entire transaction". 

Regardless, I don't understand how that's related to my question: if the report is approved, it has all cleared transaction lines. If the report is not approved, no transaction lines should be cleared. So the question then becomes: how is it possible that some lines have been marked as cleared while the report hasn't been approved?

sql/changes/1.6/constrain_cr_report_line_entries.sql
ehuelsmann: Well, maybe I don't have `Delete` rights, but I need to prepare the recon report for my manager, so after submitting one, I need to start a new one because my manager isn't available to delete the first report until the next day at which time he'll be deleting one and accepting the other?

PR #3429, opened 2018-01-25
* Automatically paginate report results when the result exceeds 1000 lines

ehuelsmann:


ehuelsmann:
Comments from the community: time to backport isn't quite right yet.

PR #3430, opened 2018-01-25
* Automatically paginate report results when the result exceeds 1000 lines

ehuelsmann:


ziuq:
Looking at the issue like a team member and seeing that it has not been a problem, I have to think this change will be inconvenient to most current users.   And more so for a first stab at the issue.    My suggestion is that it be a configuration choice.

There are lots of trade-offs to consider;  the demo yields 100 items of about 5000 in 10 seconds versus the entire lot in about 14 seconds.  Any one searching for some specific item  may appreciate this change.  Someone examining a table more generally could see this as an impediment to getting an overview.

sbts:
Ok. First impression, the "too many lines" bit possibly should be shown at the top as well as the bottom
The "Expand Full match" should be along side the <prev | next> and that should also be at the top

That should address the concern ziuq had by making it obvious and easy for a user that wants the kitchen sync to have it.

paginated output is pretty standard for virtually any web or desktop application, I don't see it as an impediment, especially if we are offering a quick way to return the kitchen sync if wanted

I'd say, lets leave it basically as this first iteration for 1.5, and put the improvements such as PGObject limited rowquery into 1.6

nick-prater:
In general, I like this development. I'd personally find it useful for those situations where I forget to set the date range on a search - better to limit the initial results than return my entire GL.

I'd prefer to see this being developed and refined on master, whereas it feels we are rushing to put this into 1.5 without a chance for proper user testing and refinement. I think that is because it is addressing a real-world problem experienced by one of our 1.5 users - but it would reassure me to see their confirmation that this addresses their problem.

I'd prefer to see the message reworded. I think saying `Too many rows match your query` sounds like an error. How about `More than 1000 rows match your query. Only the first 100 lines are shown.`?

`Expand full match` is perhaps better written `Show all rows`?

I'd like to see which 'page number' I'm viewing. I'd like a link to the first and last page. I'd like the option to show all results available on every page, not just when the first page is displayed. 

In time I'd like the option to configure/select how many rows are displayed to suit my screen size. Paging could minimise the need to scroll up and down the page. As it is, to get to the third page of results, I twice have to scroll down to the bottom of each page to click 'next'.

In summary, I think more refinement is needed, but this is a positive step forward.

PR #3487, opened 2018-03-18
Refactor Mailer.pm - remove unused strip param, silence warnings

nick-prater:
* Refactor Message-ID generation into a separate method.
* Handle undefined fields (cc, bcc etc) without triggering warnings during message_id generation.
* Remove unused strip parameter from `attach` method, which caused warnings when not defined.

### DIFF
--- a/lib/LedgerSMB/Mailer.pm
+++ b/lib/LedgerSMB/Mailer.pm
@@ -140,15 +140,12 @@ sub prepare_message {
 }
 
 =head2 $mail->attach(data => $data, file => $file,
-                     filename => $name, strip => $strip)
+                     filename => $name)
 
 Add an attachment to the prepared message.  If $data is specified, use the
 value of that variable as the attachment value, otherwise attach the file
 given by $file.  If both a file and data are given, the data is attached.
 filename must be given and is used to name the attachment.
-
-$strip is an optional string to remove from the filename send with the
-attachment.
 
 =cut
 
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Mailer.pm
```diff
@@ -193,10 +168,9 @@ sub attach {
     # strip path from output name
     my $filename;
     if ($args{filename}) {
-        my $strip = quotemeta $args{strip};
         $filename = $args{filename};
-        $filename =~ s/(.*\/|$strip)//g;
-        }
+        $filename =~ s/(.*\/)//g;
```
hasorli: I suggest to make this more portable (and safer) with File::Basename (core module) and fileparse
http://perldoc.perl.org/File/Basename.html

According to the manual do not use : "basename(1) . It does NOT always return the file name portion of a path as you might expect. To be safe, if you want the file name portion of a path use fileparse() ."

nick-prater: I like that suggestion. Are you able to do a PR?

hasorli: I am not able to test it at the moment, can you use something like (replace line 171 and 172 :  
 $filename = fileparse($args{filename},@suffixlist);

I can merge your commit to master, if the test pass.

nick-prater:
I am set up to test that code at the moment, so I'll make a PR - hopefully later today.
Thanks for making the suggestion :)

lib/LedgerSMB/Mailer.pm
nick-prater: I think actually the filename stripping can and should be removed completely from this method. Mailer should not mangle a parameter that is explicitly given by the caller.

nick-prater:
PR #3490

PR #3492, opened 2018-03-22
Comment Sauce keys

ylavoie:
Comment the now unused Sauce keys to remove Travis warnings.


ehuelsmann:
With these keys disabled and sauce not being used, why do we need to install "sauce_connect"?

ehuelsmann:
Also, why comment and not just delete?

PR #3510, opened 2018-04-04
Separate template `method` parameter

nick-prater:
LedgerSMB::Template should be self-contained. It's object creator   
and properties were, for legacy reasons, being used to pass a 
`method` argument to functions in LedgerSMB::Legacy_Util.

`method` and `media` parameters now removed from LedgerSMB::Template
module and passed directly to LedgerSMB::Legacy_Util::render_template().

All calls to `LedgerSMB::Legacy_Util::render_template()` have been 
checked against this change. `media` parameter was not in use.

### DIFF
--- a/lib/LedgerSMB/Template.pm
+++ b/lib/LedgerSMB/Template.pm
@@ -10,7 +10,7 @@ This module renders templates.
 
 =over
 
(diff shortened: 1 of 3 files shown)


einhverfr:
I like the changes

PR #3509, opened 2018-04-03
Cleanup of LedgerSMB::Template and associated test

nick-prater:
* Corrections to POD
* Removal of unused imports
* Reinstate explicit test file names, prior to further refactoring

LATEX_TESTING tests are broken and have been for some time. This is a step towards fixing and enabling them as part of CI testing.

PR #3513, opened 2018-04-05
Remove unused module LedgerSMB::Template::Elements

nick-prater:
Module LedgerSMB::Template::Elements is not in use - removing. Grepping lsmb code for the methods this module provides turns up no matches. Seems conclusive that it is not in use.


nick-prater:
These test failures will clear once PR #3509 is merged

PR #3514, opened 2018-04-05
Remove sysconfig tempdir option

nick-prater:
The LedgerSMB::Sysconfig::tempdir option is no longer used by
lsmb, so removing this configuration option entirely.


ehuelsmann:
Loving this one!

PR #3538, opened 2018-04-21
Update pod - remove auto-output reference

nick-prater:
The pod for LedgerSMB::Template had a reference to the previously removed
auto-output functionality. Pod updated accordingly.

### DIFF
--- a/lib/LedgerSMB/Template.pm
+++ b/lib/LedgerSMB/Template.pm
@@ -120,9 +120,12 @@ Returns a list of format names, any of the following (in order) as applicable:
 
 =item new_UI($request, template => $file, ...)
 
-Wrapper around the constructor that sets the path to 'UI', format to 'HTML',
-the user to C<$request->{_user}>, the locale to C<$request->{_locale}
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Template.pm
```diff
@@ -120,9 +120,12 @@ Returns a list of format names, any of the following (in order) as applicable:
...
 
-Wrapper around the constructor that sets the path to 'UI', format to 'HTML',
-the user to C<$request->{_user}>, the locale to C<$request->{_locale}
-and leaves auto-output enabled.
+Wrapper around the constructor that sets the following properties:
+
+    path   => 'UI'
+    format => 'HTML',
+    user   => C<$request->{_user}>
+    locale => C<$request->{_locale}
```
ehuelsmann: Missing a terminating `greater than`?

nick-prater: Yes - thanks for spotting. In fact, as that is now an indented code block, the C<> quotes don't work at all there.

PR #3554, opened 2018-04-30
* Improve 'save_grid' semantics

ehuelsmann:
Improves `save_grid` semantics in the sense that it's closer to what is intuitive: taking all parameters from the `grid` declaration, unless overridden in the `save_grid` call.


einhverfr:
seems positive.

PR #3578, opened 2018-05-07
Update t/data/Is_LSMB_running.html after version increment

nick-prater:
File `t/data/Is_LSMB_running.html` needs to be updated after a version increment. The test allows only one line different in lsmb output from this reference file.

Wiki checklist has been updated to include this step at https://github.com/ledgersmb/LedgerSMB/wiki/Release-check-list

### DIFF
--- a/t/data/Is_LSMB_running.html
+++ b/t/data/Is_LSMB_running.html
@@ -30,7 +30,7 @@
             };
        </script>
         <script type="text/javascript" src="js/dojo/dojo.js"></script>
-        <script type="text/javascript" src="js/lsmb/main.js"></script>
+        <script type="text/javascript" src="js-src/lsmb/main.js"></script>
         
 
         
@@ -59,7 +59,7 @@
                         alt="LedgerSMB Logo" />
                 </a>
                 <h2 align="center" style="margin-top: 0">
-                    LedgerSMB 1.6.0-dev
+                    LedgerSMB 1.7.0-dev
                 </h2>
                 <div class="listtop">
                     Database administrator credentials


t/data/Is_LSMB_running.html
```diff
@@ -30,7 +30,7 @@
             };
        </script>
         <script type="text/javascript" src="js/dojo/dojo.js"></script>
-        <script type="text/javascript" src="js/lsmb/main.js"></script>
+        <script type="text/javascript" src="js-src/lsmb/main.js"></script>
```
ehuelsmann: I don't think this `js-src` change is correct. You're likely running from source locally, but travis runs built.

nick-prater:
This is the diff
```--- t/data/Is_LSMB_running.html	2018-05-07 20:28:31.783691569 +0000
+++ /tmp/Is_LSMB_running.html	2018-05-07 22:23:38.484402328 +0000
@@ -33 +33 @@
-        <script type="text/javascript" src="js/lsmb/main.js"></script>
+        <script type="text/javascript" src="js-src/lsmb/main.js"></script>
@@ -62 +62 @@
-                    LedgerSMB 1.6.0-dev
+                    LedgerSMB 1.7.0-dev
```

The test doesn't account for this difference caused by built/unbuilt dojo. The sed filter is too restrictive: `'s|="js-src/d|="js/d|'` Arguably changing that is a better fix.

sbts:
I've raised PR's for master and 1.6 to fix the sed expression and change the number of differences to 0

PR #3584, opened 2018-05-07
Master: Ignore built/unbuilt dojo differences in t/16-schema-upgrade-html.t

nick-prater:
Depending on whether we are running with built or unbuilt dojo,
the javascript and stylesheet paths can be either `js` or `js-src`.

This patch ignores these differences when comparing html output in
tests, but continues to trap other unexpected differences.

### DIFF
--- a/t/16-schema-upgrade-html.t
+++ b/t/16-schema-upgrade-html.t
@@ -33,6 +33,19 @@ sub test_request {
 }
 
 
+sub filter_js_src {
+
+    # Given an array ref of rendered html lines,
+    # perform in-place substitution of javascript paths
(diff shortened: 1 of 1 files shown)


ehuelsmann:
With @sbts's PRs, can we close this one?

nick-prater:
Yes, superseded by @sbts PR #3587. Many thanks.

nick-prater:
I'm wrong... We do need this PR so that `t/16-schema-upgrade-html.t` will run regardless of whether dojo is built or not. Reopening.

I had confused this with another similar PR which affected Is_LSMB_Running.sh.

PR #3590, opened 2018-05-09
* Modify audittrail table to allow migration scripts to change data

ehuelsmann:


ehuelsmann:
@ylavoie and @nick-prater both confirmed the solution in the chat channel. I'm going to merge based on their positive reviews.

PR #3681, opened 2018-06-11
Update README "Environment Variables" and fix broken link

sbts:
- Update README "Environment Variables" section 
- fix broken link to "in depth installation" instructions


sbts:
No Point in running Travis here as it's a doc only update so cancelled the build.
Cancelling the whole build just caused it to re-run, so cancelled each individual job instead

PR #3701, opened 2018-06-17
Pod criticism

ehuelsmann:


ehuelsmann:
As discussed with @sbts, @ylavoie, @freelock, @jame and @nick-prater, merging without objections.

PR #3724, opened 2018-07-02
Update the development PSGI to cope with new location

ylavoie:
Make the development PSGI stay current.

### DIFF
--- a/utils/devel/ledgersmb-server-development.psgi
+++ b/utils/devel/ledgersmb-server-development.psgi
@@ -17,8 +17,6 @@ no lib '.';
 
 use FindBin;
 use lib $FindBin::Bin . '/../..'; # For our 'old code'-"require"s
-use lib $FindBin::Bin . '/../../lib';
-use lib $FindBin::Bin . '/../../old/lib';
 use LedgerSMB::PSGI;
 use LedgerSMB::PSGI::Preloads;
 use LedgerSMB::Sysconfig;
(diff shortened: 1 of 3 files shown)


utils/devel/ledgersmb-server-development.psgi
```diff
@@ -16,9 +16,9 @@ package LedgerSMB::FCGI;
 no lib '.';
 
 use FindBin;
-use lib $FindBin::Bin . '/..'; # For our 'old code'-"require"s
-use lib $FindBin::Bin . '/../lib';
-use lib $FindBin::Bin . '/../old/lib';
+use lib $FindBin::Bin . '/../..'; # For our 'old code'-"require"s
+use lib $FindBin::Bin . '/../../lib';
+use lib $FindBin::Bin . '/../../old/lib';
```
ehuelsmann: Why are these here at all? I mean, in the production version we depend on `-I` being passed to the calling process. I think that would be a good idea for the development infrastructure as well.

ylavoie: Because we have the same in ledgersmb-server.psgi, only the paths were adjusted to cope with `/bin` and `/utils/devel`.

ylavoie: I agree

ehuelsmann:
Will you replace this PR with a removal-PR?

PR #3726, opened 2018-07-02
Make sure undefined in warning messages do not trigger useless warnings

ylavoie:
Remove useless warnings when database version or company is missing.

### DIFF
--- a/lib/LedgerSMB/PSGI/Util.pm
+++ b/lib/LedgerSMB/PSGI/Util.pm
@@ -47,7 +47,8 @@ sub internal_server_error {
     my @body_lines = ( '<html><body>',
                        q{<h2 class="error">Error!</h2>},
                        "<p><b>$msg</b></p>" );
-    push @body_lines, "<p>dbversion: $dbversion, company: $company</p>"
+    push @body_lines, '<p>dbversion: ' . ($dbversion // 'undef') .
+         ', company: ' . ($company//'undef') . '</p>'
         if $company || $dbversion;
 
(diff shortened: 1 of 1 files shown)


sbts:
In general I am happy with the change, if, as @ehuelsmann has asked the error message is made more user friendly. (the error should also be translatable of course)

However, I'm interested in how it can come to pass that the DB version and Company can be undefined in the first place.
I'd have thought we shouldn't be reaching code that requires those without a valid DB session in the first place

ylavoie:
Those are part of our tests.

ehuelsmann:
Hi Yves, Could you be a bit less terse? I don't think David will understand from your response if the tests are actively testing for this, or that they happen to trigger this condition. Nor will he understand whether if -in case of the latter - the tests happen to show a bigger problem, or that this is a regular use-case in our code base.

ylavoie:
The focus of the PR was to remove a useless message in Travis about an undefined variable. After inspection, it was because we were outputting variables regardless of their definition in the error message.
Asking that LedgerSMB::PSGI::Util be internationalized is not the same and should be the focus of another PR, for there are many other routines to do in setup.

I don't mind doing it but I had pressure before to keep my PR short and focused.

ehuelsmann:
If there's no database version, then there's no valid database connection which means no language and thus the translation an identity transform, making that point moot.

The question whether this glosses over a bigger issue by suppressing the warning still stands: is there a valid use case for not having a company name and db version?

PR #3727, opened 2018-07-02
Don't try to open the database if creds are undefined

ylavoie:
We already know the database open is bound to fail later so we might as well say it now.

### DIFF
--- a/lib/LedgerSMB/Middleware/AuthenticateSession.pm
+++ b/lib/LedgerSMB/Middleware/AuthenticateSession.pm
@@ -133,6 +133,8 @@ sub call {
         unless $env->{'lsmb.company'};
 
     my $creds = LedgerSMB::Auth::factory($env)->get_credentials;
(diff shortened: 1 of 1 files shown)


ehuelsmann:
More importantly: if there are no creds, then why doesn't https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Auth/DB.pm#L46 fail? In other words, how is it possible that the changed lines are reached without https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Auth/DB.pm#L46 triggering a `die`?

ylavoie:
Because HTTP_AUTHORIZATION is not always set in https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Auth/DB.pm#L39, so we return before the check.

ehuelsmann:
Ok. Regression due to introducing middleware then.

PR #3728, opened 2018-07-02
Don't click an already opened menu

ylavoie:
Sub-menu folders clicking toggles openness. Make sure we don't close an already opened sub-menu if the browser is slow to answer.

### DIFF
--- a/xt/lib/PageObject/App/Menu.pm
+++ b/xt/lib/PageObject/App/Menu.pm
@@ -122,9 +122,12 @@ sub click_menu {
             ok($label,"Found label $label");
 
             my $submenu = $item->find("//*[\@id='$label']");
-            ok($submenu,"Submenu found " . $submenu->get_text);
+            ok($submenu,"Submenu found '" . $submenu->get_text . "'");
 
-            $submenu->click;
+            #Click toggle the foldable menus. No need if it is already opened
+            my $content = $submenu->find("//*[\@id='$label']/parent::*")->get_attribute('class');
+            $submenu->click
+                if $content !~ 'dijitTreeContentExpanded';
         }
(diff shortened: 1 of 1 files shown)


xt/lib/PageObject/App/Menu.pm
```diff
@@ -122,9 +122,12 @@ sub click_menu {
             ok($label,"Found label $label");
 
             my $submenu = $item->find("//*[\@id='$label']");
-            ok($submenu,"Submenu found " . $submenu->get_text);
+            ok($submenu,"Submenu found '" . $submenu->get_text . "'");
 
-            $submenu->click;
+            #Click toggle the foldable menus. No need if it is already opened
```
ehuelsmann: "click menu" is the specification and that's what it does. Maybe the intent of the caller is to *close* the submenu. From this code, we can't tell. I'm not sure this change is the correct one. Isn't there a better fix we can apply?

ylavoie: We don't have any tests to close sub-menus. This is part of the menu feature test, which validates that each menu is accessible.

ehuelsmann: Even if we don't have any tests depending on the now-removed behaviour, I'm concerned to have a function which doesn't do what it implies.

ylavoie: The major problem it that the click, compounded with the retries, can toggles the submenu, opening and closing until the proper timing makes the validation works.

ehuelsmann:
There's no activity on this PR and my feeling it it addresses a problem the wrong way. Closing. Please discuss and re-open if needed.

PR #3751, opened 2018-08-15
Adds transaction date to purchase history report.

einhverfr:
In 1.4 and higher since all columns are sortable, this provides more information about what was purchased in what timeframes.
    Naturally this is only available in the detail report.


einhverfr:
The major benefit of this feature is it allows someone to look at purchase history over time and use this for more pro-active reaching out to clients.  We already provide other invoice header information so this seems like a good fit here.

sql/modules/Company.sql
```diff
@@ -79,7 +79,8 @@ create type eca_history_result as (
    serialnumber text,
    exchangerate numeric,
    salesperson_id int,
-   salesperson_name text
+   salesperson_name text,
+   transdate date
```
ehuelsmann: There's just one nit with this PR and its that this change to the function signature requires dropping the function of the old signature: if we don't, we end up with two functions of the same name, only discriminated by the argument list. That's fine by me, but confuses the hell out of PGObject :-)
Could you please add the required DROP statement so I can merge the PR?

Thanks!

einhverfr: Will do.  Thanks!

PR #3795, opened 2018-08-31
Inputfield map demo

ehuelsmann:
Please review but do not merge yet.

### DIFF
--- a/lib/LedgerSMB/Scripts/payment.pm
+++ b/lib/LedgerSMB/Scripts/payment.pm
@@ -59,6 +59,7 @@ use LedgerSMB::Magic qw( MAX_DAYS_IN_MONTH EC_VENDOR );
 use LedgerSMB::PGDate;
 use LedgerSMB::PGNumber;
 use LedgerSMB::Report::Invoices::Payments;
+use LedgerSMB::Request::Helper::ParameterMap;
 use LedgerSMB::Sysconfig;
 use LedgerSMB::Template;
 use LedgerSMB::Template::UI;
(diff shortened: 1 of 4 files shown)


lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -163,42 +174,31 @@ sub pre_bulk_post_report {
...
+
+    # parse the flat "request" namespace into a hierarchical structure
+    # as defined by the $bulk_post_map transform
+    my $data = $bulk_post_map->($request);
+
+    # The user interface sets the 'id' field true-ish when the customer
+    # is selected for inclusion in the bulk payment
+    @{$data->{contacts}} = grep { $_->{id} } @{$data->{contacts}};
+    for my $crow (@{$data->{contacts}}) {
+        $crow->{accno} = $request->{ar_ap_accno};
```
nick-prater: I think this could use `$data->{ar_ap_accno}` - would be more consistent with the approach of mapping all wanted request parameters into a hierarchical data structure. Same with the following line.

lib/LedgerSMB/Scripts/reports.pm
```diff
@@ -251,7 +251,10 @@ sub reverse_overpayment {
 
 
 ###TODO-LOCALIZE-DOLLAR-AT
```
nick-prater: Can remove this comment now that `$@` is localised.

ehuelsmann:
This PR needs tests, but should otherwise be fine...

Lets wait and see if I can implement tests soon before merging.

ehuelsmann:
With @nick-prater's recent tests and the fact that the tests succeed, I think this PR now qualifies to be merged.

PR #3800, opened 2018-09-06
Add a shared LedgerSMB::Setting instance to request

nick-prater:
Many parts of the lsmb code use a `LedgerSMB::Setting` instance. Rather than each having to instantiate its own, often copying the entire request object into the new Setting instance just for the sake of a database handle, this PR provides one instance, initialised in a standard way, with a valid database handle, accessed via the `LedgerSMB` request class, which can be used wherever needed.

This further reduces the need for `LedgerSMB::App_State::DBH()` which we want to deprecate, which is currently used by `LedgerSMB::Setting->get()`. That dependency will be removed in a future PR.

### DIFF
--- a/lib/LedgerSMB.pm
+++ b/lib/LedgerSMB.pm
@@ -188,6 +188,7 @@ sub new {
     $self->{_session_id} = $request->env->{'lsmb.session_id'};
     $self->{_create_session} = $request->env->{'lsmb.create_session_cb'};
     $self->{_logout} = $request->env->{'lsmb.invalidate_session_cb'};
+    $self->{setting} = $request->env->{'lsmb.setting'};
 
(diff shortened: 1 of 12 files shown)


lib/LedgerSMB.pm
```diff
@@ -184,6 +188,7 @@ sub new {
     $self->{_session_id} = $request->env->{'lsmb.session_id'};
     $self->{_create_session} = $request->env->{'lsmb.create_session_cb'};
     $self->{_logout} = $request->env->{'lsmb.invalidate_session_cb'};
+    $self->{setting} = $request->env->{'lsmb.setting'};
```
ehuelsmann: Please prefix with an underscore: that hides the value from UI templates.

PR #3810, opened 2018-09-09
* Factor UI template rendering into a separate module

ehuelsmann:
This change is a step in the direction of completely separating document and UI
template processing into their own modules, thereby reducing complexity in both.

The UI template is a singleton pattern, setting up the renderer once (per worker)
and re-using the renderer thereafter (until the worker gets torn down).

This pattern allows faster response generation for those pages/entrypoints which
use it.


lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -141,7 +136,7 @@ sub get_search_criteria {
 
 sub pre_bulk_post_report {
     my ($request) = @_;
-    my $template = LedgerSMB::Template->new(
+    my $template = LedgerSMB::Template::UI->new( # printed document
```
nick-prater: Shouldn't this be `LedgerSMB::Template->new(...`?

PR #3812, opened 2018-09-12
Fix balance sheet report offsetted titles

ylavoie:
Ensure that html tags are properly set in all cases.

### DIFF
--- a/UI/Reports/balance_sheet.html
+++ b/UI/Reports/balance_sheet.html
@@ -49,7 +49,7 @@ END ;
       <th colspan="<?lsmb max_path_depth ?>"
           style="font-weight:bold"> </th>
       <?lsmb FOREACH col IN report.sorted_col_ids -?>
-      <th><?lsmb report.cheads.ids.$col.props.description ?></th>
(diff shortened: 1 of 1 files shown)


UI/Reports/balance_sheet.html
```diff
@@ -49,7 +49,7 @@ <h1><?lsmb text('Balance Sheet') ?></h1>
       <th colspan="<?lsmb max_path_depth ?>"
           style="font-weight:bold"> </th>
       <?lsmb FOREACH col IN report.sorted_col_ids -?>
-      <th><?lsmb report.cheads.ids.$col.props.description ?></th>
+      <th style="white-space:nowrap"><?lsmb report.cheads.ids.$col.props.description ?></th>
```
ehuelsmann: Shouldn't we do this with css?

ylavoie: Yes but we don't want it for all titles.

ehuelsmann: Ok. So we need an extra css class?

PR #3816, opened 2018-09-14
Filter Payments screen field clean-up

nick-prater:
This PR removes spurious fields from the Filter Payments screen for which there is currently no support in the database backend:
    
* categories
* department
* businesses
* date_from
* date_to
    
The underlying `payment_get_all_contact_invoices` database procedure appears to
support filtering on these fields, but on close inspection, the query it runs does not
take notice of these parameters.
    
It is unclear in any case exactly what date the from/to filters should operate on -
due date, tax point, entry date - it's ambiguous.
    
Removing these fields ensures that we present only working functionality which
we are able to test, without the distraction of features yet to be implemented.
They can easily be added back as and when anybody requires these features
and is willing to implement them.

To complement this, the Payments Detail screen now displays the `Date From`,
`Date To` and `Payment Type` fields only if their respective values are defined.

### DIFF
--- a/UI/payments/payments_detail.html
+++ b/UI/payments/payments_detail.html
@@ -106,12 +106,18 @@
                 } ?>
         </div>
 
-        <div id="date_filter_row">
-                <label for="filter_from"><?lsmb text('Filtering From') ?></label>
-                <span id="filter_from"><?lsmb payment.date_from ?></span>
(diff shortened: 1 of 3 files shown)


ehuelsmann:
I like the approach of restricting the UI to what we currently implement. If there's anything missing, we can add issues to the tracker to keep a log of all desired functions not yet implemented.

PR #3826, opened 2018-09-15
Drop LedgerSMB module VERSION property

nick-prater:
The `LedgerSMB` module had both `version` and a `VERSION` property.
Both were explicitly set to the the same value and therefore duplicate
one another.

To remove this duplication, the `VERSION` property has been deleted and
uses of it replaced with `version`.

The lowercase version was retained to be consistent with the other object
properties which are also lowercase.


ehuelsmann:
Nice cleanup. Thanks!

PR #3839, opened 2018-09-21
Cleanup and add tests for Batch_Overview module

nick-prater:
This PR cleans-up and adds test for the module
`LedgerSMB::Report::Unapproved::Batch_Overview`.

* Unused methods `subtotal_cols` and `text` removed
* Unit tests added for `get_rows` method
* Allow explicit `undef` setting of `class_id` property
* Separate `get_rows` database query method from `run_report` for more granular testing
* Reformat and augment pod
* Remove unused module dependencies
* Document subclassing expectations of `LedgerSMB::Report`

### DIFF
--- a/t/testrules.yml
+++ b/t/testrules.yml
@@ -12,12 +12,12 @@ seq:
       - xt/{0,1,2,3}*
   - seq:
     - xt/40-dbsetup.t
-    - xt/45*.t
-    - xt/47*.t
-    - xt/48*.t
     - par:
       - xt/40-database.t
       - xt/41-coaload.t
+      - xt/45*.t
+      - xt/47*.t
+      - xt/48*.t
   - seq:
     - par:
       - xt/{43,44,5,7}*.t
--- a/t/testrules.yml
+++ b/t/testrules.yml
@@ -14,6 +14,7 @@ seq:
     - xt/40-dbsetup.t
     - xt/45*.t
     - xt/47*.t
+    - xt/48*.t
     - par:
       - xt/40-database.t
       - xt/41-coaload.t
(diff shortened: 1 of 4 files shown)


xt/48.1-ledgersmb-report-unapproved-batch_overview.t
```diff
@@ -0,0 +1,137 @@
...
+is($report->approved(undef), undef, 'reset approved filter');
+
+# Query unapproved
+is($report->approved(0), 0, 'set approved filter = false');
+ok($rows = $report->get_rows, 'get_rows() with approved_filter = false');
+is(scalar @{$rows}, 3, 'querying with approved=false returned 3 rows');
+is($report->approved(undef), undef, 'reset approved filter');
+
+
+#TODO Query Amount
```
ehuelsmann: Do we need to create an issue for this in GitHub? If we don't, how are we going to track what items are remaining?

Or is there work in progress and should I hold back on merging?

t/testrules.yml
```diff
@@ -14,6 +14,7 @@ seq:
     - xt/40-dbsetup.t
     - xt/45*.t
```
ehuelsmann: Is there a reason not to add 45-48 to the "par" (parallel) section below?

xt/48.1-ledgersmb-report-unapproved-batch_overview.t
nick-prater: I would say no need to raise an issue in GitHub. We find the issue again by grepping `TODO` in our source code - and there's plenty like it. This isn't a bug or user-affecting. Our current unit test coverage is so patchy, I think it makes no sense to raise GitHub issues for missing unit test coverage until we're nearer 100%.

There's no further work in progress on this test. If you prefer, I could add the missing part, but that makes it more difficult for you in merging MC.

t/testrules.yml
nick-prater: Not that I'm aware of. I'll PR and we'll see what breaks.

PR #3844, opened 2018-09-23
Search Batches UX Improvements

nick-prater:
A number of improvements and fixes to polish the Search Batches results screen:

* Change title s/Search Unapproved Transactions/Search Batches/.
* Change title s/Batch Type/Transaction Type/ to be consistent with previous screen.
* Fix display of `Transaction Type` (the batch class name) by adding trigger to
  Moose class to look up the batch class name whenever the id property is changed.
* Fix display of `Description` - was previously showing non-existent field `Reference`.
* Don't display buttons when showing approved batches - the approve, delete, unlock
  actions they present cannot be used once a batch is approved.
* Removed meaningless `(Locked)` heading field - we don't allow searching on this and it
  was not being populated - it has meaning only for an individual batch, not a result set.

Internal changes:

* Simplify template processing. Don't loop through Batch Class list re-mapping
  variables. Instead specify appropriate key names to use when building the dropdown,
  so that the class list can be used unmodified.
* Expand and correct pod for `list_batches` method.

### DIFF
--- a/UI/Reports/filters/batches.html
+++ b/UI/Reports/filters/batches.html
@@ -1,9 +1,6 @@
 <?lsmb
     PROCESS "elements.html";
     PROCESS 'report_base.html';
-
-    # Empty item at top of batch class list - Allow any class
-    batch_classes.unshift({});
 ?>
 
 <body class="lsmb <?lsmb dojo_theme ?>">
@@ -19,6 +16,7 @@
                 value_attr = "id"
                 text_attr = "class"
                 name = "class_id"
+                default_blank = "true"
             } ?>
         </div>
 
--- a/UI/Reports/filters/batches.html
+++ b/UI/Reports/filters/batches.html
@@ -1,27 +1,23 @@
 <?lsmb
     PROCESS "elements.html";
     PROCESS 'report_base.html';
-    action = 'list_batches';
 
-    FOR CLS IN batch_classes;
-        CLS.text = CLS.class;
-        CLS.value = CLS.id;
-    END;
+    # Empty item at top of batch class list - Allow any class
     batch_classes.unshift({});
 ?>
 
 <body class="lsmb <?lsmb dojo_theme ?>">
 <div id="batch_search">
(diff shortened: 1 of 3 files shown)


UI/Reports/filters/batches.html
```diff
@@ -1,27 +1,23 @@
 <?lsmb
     PROCESS "elements.html";
     PROCESS 'report_base.html';
-    action = 'list_batches';
 
-    FOR CLS IN batch_classes;
-        CLS.text = CLS.class;
-        CLS.value = CLS.id;
-    END;
+    # Empty item at top of batch class list - Allow any class
```
ehuelsmann: This can actually be achieved using the "default_blank" option for the SELECT block (see https://github.com/ledgersmb/LedgerSMB/blob/master/UI/lib/elements.html#L249)

lib/LedgerSMB/Report/Unapproved/Batch_Overview.pm
```diff
@@ -160,49 +185,83 @@ Returns the inputs to display on header.
...
-                    value => 'batch_delete',
-                    class => 'submit',
-                 },{
-                    name  => 'action',
-                    type  => 'submit',
-                    text  => $self->_locale->text('Unlock'),
-                    value => 'batch_unlock',
-                    class => 'submit',
-                }]);
+    $self->get_buttons();
```
ehuelsmann: Just wondering: shouldn't this be a functionality of the main `LedgerSMB::Report` class, as that allows rendering of input elements as well?

nick-prater: Yes, `LedgerSMB::Report` expects that it's `set_buttons` method will be overrided to achieve this. Fixed in commit 6632dd4628b82e1f417051898d23dd8bb4499790

UI/Reports/filters/batches.html
nick-prater: Thanks. Changed in commit 402654bcf9e34fedb9f349efdc68c67862264764

PR #3938, opened 2018-12-17
Travis xenial

ylavoie:
- Move the Travis test to Ubuntu Xenial.
- Remove Perl 5.18 which is now 4 years old and PostGreSQL which is now end-of-life.


ylavoie:
I'm working on it, it fails on my repo too. I published too fast, it seemed so straightforward.

PR #3954, opened 2018-12-28
Show custom account links

nick-prater:
The LedgerSMB schema has the functionality to allow custom account_links - custom flags which can be added to accounts from the Chart of Accounts.

This PR provides an interface to expose this functionality, adding checkboxes to the Edit Account screen so that the custom flags can be added or removed from an account.

BDD and sql tests are extended to cover this functionality.

### DIFF
--- a/UI/accounts/edit.html
+++ b/UI/accounts/edit.html
@@ -257,6 +257,22 @@
               value = '1'} ?>
    </div>
 </div>
+<?lsmb IF form.custom_link_descriptions.size ?>
+<div class="inputline" id="acc-custom-flags-line">
+    <label class="line"><?lsmb text('Custom Flags') ?></label>
+    <?lsmb FOREACH link IN form.custom_link_descriptions ?>
+    <div class="inputgroup">
+        <?lsmb INCLUDE input element_data={
+              name = link.description,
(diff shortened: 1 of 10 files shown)


UI/accounts/edit.html
```diff
@@ -257,6 +257,22 @@
               value = '1'} ?>
    </div>
 </div>
+<?lsmb IF form.custom_link_descriptions.size ?>
+<div class="inputline" id="acc-custom-flags-line">
```
ehuelsmann: does this imply that there are only ever going to be as many checkmarks as fit on a single row? (How many checkmarks are there on the other rows? 6?) Or should we support wrapping to new rows by limiting the number on the row to the same number on the "built-in" rows?

nick-prater: It's a good point, but as things stand, it will wrap, albeit not very prettily.
I think this is a CSS concern.
We can't wrap based purely on the number of checkmarks as (unlike the other fixed checkmarks) we have no control over how long the label is for each.

ehuelsmann: Ok. Good point. Other than that, I think this is a nice and clean PR. Thanks!

PR #3968, opened 2019-01-02
Constrain account_links: account can be summary for at most one descriptor

nick-prater:
This PR adds a database trigger constraint and companion schema change tests to
ensure that accounts can only be a Summary account for a single
account_link_description.
    
This constraint is assumed by the lsmb UI but was not previously enforced.

To provide a user-friendly schema change check interface, this PR extends the
select dropdown HTML template to accept a callback function to define the option
list on a per-row basis, without affecting existing behaviour.

### DIFF
--- a/sql/changes/1.7/limit_summary_account_links.sql.checks.pl
+++ b/sql/changes/1.7/limit_summary_account_links.sql.checks.pl
@@ -10,8 +10,7 @@ check(
         SELECT account.accno AS "Account",
                account.id AS account_id,
                account.description AS "Description",
-               string_agg(account_link.description, ', ') AS "Currently Summary For",
-'' AS "New Summary For"
+               string_agg(account_link.description, ', ') AS "Currently Summary For"
         FROM account
         JOIN account_link ON (
             account.id = account_link.account_id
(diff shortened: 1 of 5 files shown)


sql/changes/1.7/limit_summary_account_links.sql.checks.pl
```diff
@@ -0,0 +1,101 @@
...
+
+An account may only be a summary for a single transaction type, but this
+was not enforced in earlier versions of LedgerSMB.
+
+To continue with the upgrade, the listed accounts must be altered so that
+they are a summary account for only a single transaction type, or changed
+so that they are no longer a summary account.
+
+For each account row below, please select which transaction type it should be
+a summary accout for, or leave the selection blank if the account
```
ehuelsmann: typo ('accout' should be 'account')

sql/changes/1.7/limit_summary_account_links.sql.checks.pl
```diff
@@ -0,0 +1,101 @@
...
+should no longer be considered as a summary accout.
```
ehuelsmann: same typo :-)

PR #3971, opened 2019-01-04
Show custom summary account_link descriptors in UI

nick-prater:
This commit extends the Chart of Accounts edit screen to include
custom summary account_link descriptors in the 'Summary Account For'
dropdown.
    
This dropdown is now populated from the database
`account_link_description` table, rather than being hard-coded (but
required to correspond).
    
Checks in the perl code to limit the number of summary account_link
descriptors linked to an account have been removed. The UI prevents
this and the database now has a constraint to enforce this. The
perl checks are therefore unnecessary.

### DIFF
--- a/UI/accounts/edit.html
+++ b/UI/accounts/edit.html
@@ -274,27 +274,36 @@
 </div>
 <?lsmb END ?>
 <div class="inputline" id="summary-line">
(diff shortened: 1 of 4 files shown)


ehuelsmann:
Very nice cleanup! Thanks!

PR #3991, opened 2019-01-12
Fix #2567

### LINKED ISSUE #2567: Reports->Customer History Ignores Date Filter
Result set is the same regardless of date range entered in "Startdate" or "Transaction Date" although caption at top of page does show requested transaction date range.

ehuelsmann:


ehuelsmann:
@nick-prater you were going to write a few tests for this query right?

PR #4000, opened 2019-01-19
Merge multi currency support

### LINKED ISSUE #560: rewrite of currency and exchangerates
After the accounts, plan on refactoring currency and exchangerate handling.  The plan is to have a buy/sell system which can be better constrained than what we have now and useful for individual transactions as we move into the financial rewrite.

The initial phase would be backwards-compatible largely with previous implementations. 

New tables:
- currency 
  - symbol
  - allowed_variance
  - default

Would have a dojo form for adding currencies in a tabular manner, orderd by symbol

Eschangerate would have an fkey on symbol.

New urls:

.../currency/$symbol/ get: report of currency
 .../currency/$symbol/ put/post, update currency
 ..,/currency/$symbol.json/xml get pr put currency 
 .../currency/$symbol/$bysell/$date get|post|put a new exchangerate (html/http form)
  .../currency/$symbol/$bysell/$date.json|xml get|post|put

### LINKED ISSUE #3410: 1.4: Exchangerate gets copied (unoverridable) when copying an invoice in foreign currency
When copying an invoice in non-local currency from one date to another, the exchange rate gets copied too, without being overridable in the UI. That is: the UI offers an exchange rate entry (pre-filled), but setting a different exchange rate doesn't help.

(Note that this problem is solved on the MC branch...)

ehuelsmann:
Support multiple exchange rates per currency per day by supporting an exchangerate per `acc_trans` line. AR/AP invoices/transactions still have a single rate and currency per transaction, however, GL transactions now support a multitude of currencies and rates in a single transaction due to this change.

Reason to want to allow multiple currencies per transaction include the desire to post an opening balance with multiple currencies or to post a paypal transaction, transferring money from one foreign currency to another (without going through the functional currency).

Fix #3410 
Fix #560


einhverfr:
Awesome!

PR #4013, opened 2019-01-27
* Move role grants to Roles.sql

ehuelsmann:
While loading some database to be repaired while helping a user,
the fact that some GRANTs are in Pg-database.sql means they don't
get applied upon reloading the modules -- with resulting in access
problems.
This may be a problem with restores across PG clusters too, when the
user elects not to load the roles file.

Note that the 'chart' view exists in Pg-database, but gets dropped
during "schema change updates", meaning the grant on it can't be
executed in Roles.sql: by the time it gets there, the view doesn't
exist anymore.

### DIFF
--- a/sql/Pg-database.sql
+++ b/sql/Pg-database.sql
@@ -662,8 +662,6 @@ create table location_class_to_entity_class (
   entity_class int not null references entity_class(id)
 );
 
-GRANT SELECT ON location_class_to_entity_class TO PUBLIC;
-
 COMMENT ON TABLE location_class_to_entity_class IS
 $$This determines which location classes go with which entity classes$$;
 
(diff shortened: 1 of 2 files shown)


sql/Pg-database.sql
```diff
@@ -4897,8 +4893,6 @@ select c.id, c.accno, coalesce(at.description, c.description),
 group by c.id, c.accno, coalesce(at.description, c.description), c.category,
          c.heading, c.gifi_accno, c.contra, c.tax;
 
-GRANT SELECT ON chart TO public;
-
 COMMENT ON VIEW chart IS $$Compatibility chart for 1.2 and earlier.$$;
```
einhverfr: Maybe another PR might be to remove the chart view in 1.7

ehuelsmann: Yea. I've thought of that. However, that would break the "incremental upgrades" in `sql/changes/`, because there's [drop_chart.sql](https://github.com/ledgersmb/LedgerSMB/blob/master/sql/changes/1.6/drop_chart.sql) upgrade script which would fail to apply.

The best way forward is probably by carefully designing a way to integrate the incremental scripts (automatically) into the schema file and making sure the upgrade scripts won't be applied anymore when they're "merged". It's something I've been thinking about, but not had the opportunity yet to write code for. (But, yes, this confused me before, so we seem to be at a point where the incrementals start to be confusing...)

PR #4037, opened 2019-05-03
Fix recon reports for mc

ylavoie:
1- Fix the foreign currency validation check to use buy rate for AP and sell for AR
2- Sum the transation and its FX part in the reconciliation records before removing the FX.

### DIFF
--- a/sql/changes/mc/acc-trans-migration.sql
+++ b/sql/changes/mc/acc-trans-migration.sql
@@ -134,13 +134,13 @@ BEGIN
                           where ltd.id = crl.ledger_id);
   UPDATE cr_report_line crl
      SET our_balance = our_balance
-                     + (select our_balance from cr_report_line cl
+                     + (select sum(our_balance) from cr_report_line cl
                            join lines_to_delete ltd on cl.ledger_id = ltd.id
-                          where ltd.id = crl.ledger_id or ltd.assoc = crl.ledger_id),
+                          where ltd.assoc = crl.ledger_id),
          their_balance = their_balance
-                       + (select their_balance from cr_report_line cl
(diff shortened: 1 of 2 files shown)


sql/changes/mc/acc-trans-migration.sql
```diff
@@ -132,6 +132,16 @@ BEGIN
          AND NOT EXISTS (select 1 from cr_report_line cl
                            join lines_to_delete ltd on cl.ledger_id = ltd.assoc
                           where ltd.id = crl.ledger_id);
+  UPDATE cr_report_line crl
+     SET our_balance = our_balance
+                     + (select our_balance from cr_report_line cl
+                           join lines_to_delete ltd on cl.ledger_id = ltd.id
+                          where ltd.id = crl.ledger_id or ltd.assoc = crl.ledger_id),
```
ehuelsmann: Are you sure the `assoc` is correct here? It's not the line being deleted, so I would expect either the sum to be incorrect (sometimes) or multiple lines to be returned, which isn't allowed with this nested select.

sql/changes/mc/acc-trans-migration.sql
```diff
@@ -132,6 +132,16 @@ BEGIN
...
+         their_balance = their_balance
+                       + (select their_balance from cr_report_line cl
+                           join lines_to_delete ltd on cl.ledger_id = ltd.id
+                          where ltd.id = crl.ledger_id or ltd.assoc = crl.ledger_id)
```
ehuelsmann: same here.

ylavoie:
Simplified, even though the previous version was ok

PR #4043, opened 2019-06-23
* Before building new 'content' attribute, assert it's valid content

ehuelsmann:
While at it, factor out the similar check from 'wait_for_content'.


ehuelsmann:
CircleCI is a non-error: it can't find the build config -- which is correct.

The Travis error is simply because Travis tripped on the Sauce Labs tests which we don't care about at this point.

PR #4045, opened 2019-06-23
* Fix the warning about UTF-8 output in the BDD tests

ehuelsmann:


ehuelsmann:
Doesn't actually work.

PR #4046, opened 2019-06-24
Add class tag to confirm menu are loaded

ylavoie:
Add a CSS tag to signal that the menu tree is completely loaded.
BDD tests must rely on a loaded menu to exercise it properly.


ehuelsmann:
This PR is - as discussed with @ylavoie - no longer required for tests to succeed.

PR #4050, opened 2019-07-02
Updated pr4042

ylavoie:
Add latest updates on CircleCI branch

PR #4067, opened 2019-07-14
Circleci final

ehuelsmann:
Move coverage testing to CircleCI from TravisCI.

Closes #4042 and #4050 

The maximum run-time for a test job on TravisCI is 50 minutes. Our coverage test jobs run 40 minutes (when run off a cached earlier run -- meaning sometimes longer than 40 minutes). Jobs on CircleCI can run *much* longer (experience so far: 10 hours).
The move to CircleCI gained us a lot of insight in parts of our test suite that were working "only by accident" -- as in that they needed fixing to work correctly on CircleCI, which should make our tests much more robust. Additionally, we will win the opportunity to test (more easily) with other browsers than PhantomJS (FF and Chrome are the immediate candidates).


ehuelsmann:
Weird. The circleci job has completed succesfully, but didn't report back.

PR #4069, opened 2019-07-14
* Fix setup.pl discarding the need to re-authenticate

ehuelsmann:
Browsers may choose not to send credentials
and expect to receive a 403 unauthenticated back;
we however simply discard the failed authentication
and proceed until we crash somewhere further down
the line. Instead of failing, check the creds and
request re-authentication.


ehuelsmann:
Testing failed; no real indication why. Closing for now.

PR #4070, opened 2019-07-15
Set PERL5LIB for perlbrew

ylavoie:
Set up PERL5LIB for perlbrew for CircleCI tests


ylavoie:
Still leaves issues with globally installed perl

PR #4078, opened 2019-07-21
Remove PERL5OPT in favor of the less limited HARNESS_PERL_SWITCHES

ylavoie:
HARNESS_PERL_SWITCHES does not limit the Perl switch that can be used, contrary to PERL5OPT, so we can remove Deparse warnings and "Module will be removed from core" info.
The objective is to reduce the test noise to show errors and warnings relevant to LedgerSMB only.


ylavoie:
HARNESS_PERL_SWITCHES is not propagated to inners invocations

PR #4118, opened 2019-08-26
Complete move to port 5672 for PSGI

ylavoie:
- Adjust the plackup systemd service file for 5762
- Make sure that Travis checks for 5762 as default value


ehuelsmann:
@ylavoie shouldn't this PR also include a change to .circleci/config.yml ?

PR #4143, opened 2019-10-06
Update payment table documentation

### LINKED ISSUE #4142: Document the payment and payment_links tables
The documentation for these tables isn't complete as the full intent of the tables and how to use them, isn't included in the documentation.

The documentation should include both design considerations and use-case examples.

ehuelsmann:
Closes #4142

### DIFF
--- a/sql/changes/1.8/update-payment-tables-documentation.sql
+++ b/sql/changes/1.8/update-payment-tables-documentation.sql
@@ -1,0 +1,132 @@
+
+COMMENT ON TABLE payment IS
+$$ This table tracks header data for payments.
+
+Its purpose is to maintain references to a gl transaction (overpayments)
+and purchase/sales orders (prepayments).
+
+Together with the payment_links table, it assembles data spread across multiple
+transactions into a single payment for reconciliation purposes.
+
+This table exists to record as a single payment 'acc_trans' lines spread over
+multiple AR/AP transactions: payment lines use the same 'trans_id' as the
+AR/AP item which they are payments to. This means that there is a need for
+a secondary table to track the various parts of a payment. E.g. when a 30USD
+payment is entered for 2 invoices (INV1 @ 10USD, INV2 @ 20USD), one might
+expect it to be recorded as
+
+  +----------+--------+-------+---------+------+
+  | trans_id | amount | DB/CR | account | ref  |
+  +----------+--------+-------+---------+------+
+  |  25      |  30.00 | DB    | Bank    |      |
+  |  25      |  10.00 | CR    | AR      | INV1 |
+  |  25      |  20.00 | CR    | AR      | INV2 |
+  +----------+--------+-------+---------+------+
+
+However, the payment is actually recorded in the 'acc_trans' table  as
+below (assuming INV1 has trans_id 1 and INV2 has trans_id 2):
+
+  +----------+--------+-------+---------+------+
+  | trans_id | amount | DT/CR | account | ref  |
+  +----------+--------+-------+---------+------+
+  |    1     |  10.00 | DT    | Bank    |      |
+  |    1     |  10.00 | CR    | AR      | INV1 |
+  +----------+--------+-------+---------+------+
+  |    2     |  20.00 | DT    | Bank    |      |
+  |    2     |  20.00 | CR    | AR      | INV2 |
+  +----------+--------+-------+---------+------+
+
+To track the lines spread across the two transactions into a single payment,
+the 'payment_links' table references each of the lines as well as a line in
+the 'payment' table.  Each line in the 'payment_links' table has a 'type',
+which has one of the three values below. For non-overpayment lines, the type
+is always '1'.
+
+Overpayments are entered into the ledger as a GL transaction of two lines; one
+line posts against the cash account; the other against the overpayment account.
+The overpayment gets entered into this table with payment_links of type 2 to the
+associated acc_trans lines. E.g.
+
+  +----------+---------+-------+---------+--------------+--------+
+  | trans_id |  amount | DT/CR | account | payment type | eca_id |
+  +----------+---------+-------+---------+--------------+--------+
+  |   15     |   40.00 |  DT   | Bank    |   2          | 21     |
+  |   15     |   40.00 |  CR   | Prepay  |   2          | 21     |
+  +----------+---------+-------+---------+--------------+--------+
+
+Note that the 'payment type' is the 'type' field in the 'payment_links' table
+and that the 'eca_id' field is the 'entity_credit_id' from the 'payment'
+table.
+
+
+Allocations of overpayments to invoices are entered as 2 lines of an AR/AP
+transaction, linking the associated acc_trans lines to a payment using
+payment_links of type 0. In case the payment involves a foreign currency gain
+or loss, the allocation consists of 3 lines. All three lines are included
+in the payment_links.
+A 4th line may be involved when an early payment discount is in order.
+
+E.g. when an overpayment gets used to pay the two invoices from the example
+above, the transactions will look like this (when the payment is payment_id
+'20' and the overpayment is payment_id '15' -- as created above):
+
+  +----------+--------+-------+---------+------+------------+--------+
+  | trans_id | amount | DT/CR | account | ref  | payment_id | ovp_id |
+  +----------+--------+-------+---------+------+------------+--------+
+  |    1     |  10.00 | DT    | Prepay  |      |    20      |   15   |
+  |    1     |  10.00 | CR    | AR      | INV1 |    20      |   N/A  |
+  +----------+--------+-------+---------+------+------------+--------+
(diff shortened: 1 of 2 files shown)


ehuelsmann:
@ylavoie Submitted a new description with my latest understanding of the payments table's purpose. I hope this addresses your concern about the description originally mostly being about overpayments.

ehuelsmann:
> Being only somewhat proficient with SQL and having a different mental model of bookkeeping than lsmb shows, I might not be your best reviewer.

Or exactly the one we're looking for: one who needs a lot of context and documentation and obviously documentation which is concisely written, without vagueness.

> But I've found it worthwhile, thanks.

Thanks for the review!

> Some of what you write here would be better put into a programmer's intro to lsmb. That is colored by the table names which I have found challenging with the 150 or so of them. A few
> example user transactions, a map of the tables concerned and how they are effected would be good.

Very nice idea. Indeed. I'll be looking where to write that and add the same there too.

> If this is the table to start looking for my sent or received cheque and its effects, that could be said.

It's the table to find that, yes. However, it's not just that: it's also any cash receipts/payments, credit card transactions and bank account transactions. I'll add something to this effect. The effects you don't really see in this table: the (balance sheet) effects are stored in the `acc_trans` table.

> Lines 109-111: These seem to presume that there is some way to know what the gain/loss would be known when payment is received.

There are cases where a payment to a foreign currency invoice involves a specific fx rate: when a USD invoice gets paid to a EUR bank account, the bank applies an fx rate. Originally, the invoice was posted with a different rate, meaning a loss/gain needs to be calculated.

> In the general case, converting foreign currency to functional cannot be connected to a payment received.

The way it currently works is that any payment to a foreign currency invoice gets a gain/loss posted, assuming the rate at which the payment happens is different from the rate at which the invoice was posted.

>  How do the periodic FX adjustments fix in with this?

The above "pattern"/idea is inherited from SL. Neither SL nor LSMB have periodic revaluation (for LSMB: yet). When there will be periodic revaluation, it's possible that part of the gain/loss gets reversed due to changes in rates, or selection of different base rates.

> I don't understand lines 5-6: the parenthetical "over-/prepayment" is meant to clarify what?

gl transactions in case of overpayments ; and orders in case of prepayments.

> Line 9: Parentheses are not necessary, but "reconciliation" is too narrow. I might just want to know what USD came in or went out today.

But wouldn't a query on your bank account, petty cash account and credit card account tell you that? I mean, in my bank account, not all transactions are recorded as payments: I have bank transactions I post directly into the books (e.g. VAT charged that I need to transfer to the tax authorities); those transactions *do* constitute outflows. But maybe that's not what you mean?

The "for reconciliation purposes" means to say that the individual lines in the different transactions are sufficient to report cashflows on, but for recon purposes should be considered a single payment. That single payment ("single number") isn't recorded in the books, as it is now.

> Line 11-38: "work-around" I know there has been a lot of issues with payments, but I think it would
> be clearer to just describe the purpose and implementation first and address the quirks of legacy problems if necessary.

Very good point. I'll adjust the text.
 
> Pedantically, I say payments get "applied to" some other journal document, not "spread over" (which is clear enough).

Right. I think I'd normally do that too. But the point of this sentence is that the lines of the payment don't "exist" in a payment transaction. Instead, the lines which were added due to the payment, are associated with the transaction id of the invoices they pay: they are fully subsumed into those transactions. Hence my wording "spread over" to signal this fact.

> Between lines 18-56 you have three table snippets:
> "amount" is misspelled in the first.
> I am expecting your 1st and 2nd table snippets to refer to the payment table (per line 14); but I guess they are just made up.

Actually, the tables at lines 18 and 29 are the same table in close resemblance of the `acc_trans` table. They are showing content which you would expect (first table) contrasting that with what will actually be recorded (second table). It's impossible to plainly use the `acc_trans` table, because it doesn't have columns such as "account name", "debit/credit" or "reference": those fields come from other tables, but are required to make the example table meaningful.

> That "trans_id" seems to be an 'id' in one and a FK in the other is confusing.

I'm not sure what you mean. Tables 1 and 2 go together. Table 3 shows a completely independent use-case. Does that need further clarification in the text?

> Line 45: "Overpayments are entered into the ledger as a GL transaction", they should be in journal

My bad: I should have called it "general journal".

> 4 - receipts or 5-payments (as of 1.7.0).
> 
> The three types of payments seem redundant: monies in a prepay, overpaid, or on-deposit account are not assets (from seller's view) so the source and sink accounts and 'ref' are sufficient, or just the transactions' ids. Already etched, I think.

Is this a remark with respect to the third table? If that's the case, the point of that table is to show what the allocation of a pre-/overpayment to invoice(s) looks like. The money in a prepayment and overpayment accounts is a liability. I'm not sure if that matters for table 3 though.

> Your "AR/AP" mentions are essential. Someone may start reading and think "payments" have to do with "paying" and another might think of processing "payments" aka "cash receipts".

Good. This table actually has more to do with allocating incoming money (receipts) or available money (pre- & overpayments)  from various accounts to open "AR" items -- invoices.

ziuq:
> > Some of what you write here would be better put into a programmer's intro to lsmb.

I encourage you to put that in your schedule.   A explication of some basic transactions might invite more participation.  Another conversation.

> > If this is the table to start looking for my sent or received cheque and its effects
> ... yes ... it's also any cash receipts/payments, credit card transactions and bank account transactions.

Understood.

> I'll add something to this effect. The effects you don't really see in this table: the (balance sheet) effects are stored in the `acc_trans` table.

Which I take is the core of all journal entry lines.
 
> > How do the periodic FX adjustments fix in with this?

I led with the FX issue because it seems  flawed.  You are pointing out that there are businesses that never see the FX that they billed out.  I agree that Lsmb may work for them ( I have not thought much about the A/P side).  But this is a big limitation.

> >... I might just want to know what USD came in or went out today.
> But wouldn't a query on your bank account, petty cash account and credit card account tell you that?

> The "for reconciliation purposes" means to say that the individual lines in the different transactions are sufficient to report cashflows on, but for recon purposes should be considered a single payment. That single payment ("single number") isn't recorded in the books, as it is now.


> > Pedantically, I say payments get "applied to" 
> Right. I think I'd normally do that too. But the point of this sentence is that the lines of the payment don't "exist" in a payment transaction. Instead, the lines which were added due to the payment, are associated with the transaction id of the invoices they pay: they are fully subsumed into those transactions. Hence my wording "spread over" to signal this fact.

A payment necessarily  invades other "transactions" to adjust their balance but it still needs to be viewed as a separate "transaction".  

> > Between lines 18-56 you have three table snippets:
> > "amount" is misspelled in the first.
> > I am expecting your 1st and 2nd table snippets to refer to the payment table (per line 14); but I guess they are just made up.
> 
> Actually, the tables at lines 18 and 29 are the same table in close resemblance of the `acc_trans` table. They are showing content which you would expect (first table) contrasting that with what will actually be recorded (second table). It's impossible to plainly use the `acc_trans` table, because it doesn't have columns such as "account name", "debit/credit" or "reference": those fields come from other tables, but are required to make the example table meaningful.
> 
> > That "trans_id" seems to be an 'id' in one and a FK in the other is confusing.
> 
> I'm not sure what you mean. Tables 1 and 2 go together. Table 3 shows a completely independent use-case. Does that need further clarification in the text?

Yes.  I noted the inconsistency or change but didn't know if it was intentional.  The formatting at first made me think these were actual query results.

This style would have been clearer to me, as being closer to the implementation.

Table A                    Table B
id  fld1  fld2  fldN -->  id  ...
----------------------        ------------
1   cr    1.00          20        ....

> > Line 45: "Overpayments are entered into the ledger as a GL transaction", they should be in journal
 
> ... I should have called it "general journal".

Not my point, they should be in the a/p or a/r journal.

> > 4 - receipts or 5-payments (as of 1.7.0).
> > The three types of payments seem redundant: monies in a prepay, overpaid, or on-deposit account are not assets (from seller's view) so the source and sink accounts and 'ref' are sufficient, or just the transactions' ids. Already etched, I think.

> Is this a remark with respect to the third table? If that's the case, the point of that table is to show what the allocation of a pre-/overpayment to invoice(s) looks like. The money in a prepayment and overpayment accounts is a liability. I'm not sure if that matters for table 3 though.

Yes, that was more general design comment, a view-like representation might have clarified for
me.
 
> > Your "AR/AP" mentions are essential. ...

> Good. This table actually has more to do with allocating incoming money (receipts) or available money (pre- & overpayments) from various accounts to open "AR" items -- invoices.

I assume you just mean that A/R activity often outnumbers A/P.  Someone selling desktops, workstations, and larger horses custom assembled could easily have more A/P transactions than A/R.

I meant to mention yesterday:  I would expect either or both of your first two table examples.  I see those in manual books and automated G/Js.   The first is the usual textbook example, but often it is easier to just enter offsets to one account than total them into one line.

ehuelsmann:
Ok. Only comments to the actual design remain. I'll factor those out later into separate issue(s) our as comments into existing refactoring issues.

Meeting documentation as it is now.

PR #4155, opened 2019-10-10
Keep the accounts properties for GIFI on Balance Sheet and Income Statement

ylavoie:
GIFI only change the account number and description, in order to standardize reporting to governments. All other account properties have to be preserved when producing Balance Sheet or Income Statement reports


ehuelsmann:
I think this has been backported yesterday already, right? (By me, that is?)

PR #4207, opened 2019-11-11
Remove test noise

ylavoie:
1- Remove notices in COGS to remove test noise.
2- Set PostgreSQL minimum messages in pgTAG test files to prevent notices coming from PostgreSQL code about text-search queries which don't contain lexemes searched.


ehuelsmann:
Thanks!

PR #4269, opened 2019-12-08
Reduce two more uses of LedgerSMB::App_State::DBH

ehuelsmann:


ehuelsmann:
Close PR until we have the branch fixed.

PR #4370, opened 2020-01-18
Fix Add Currency 'Action not defined: save_currency' error

nick-prater:
On System->Currency->Edit Currencies screen, clicking 'Add'
produced an error because the form action was not correctly
defined.


ehuelsmann:
Yay! Welcome back!

PR #4374, opened 2020-01-20
Fix makeblacklist misses some functions for testing

nick-prater:
`makeblacklist` is meant to generate a list of all database functions
defined in the `sql/modules` directory. These are then checked by
tests to ensure that they have not been overloaded, which our object
mapper does not expect.

I discovered that `makeblacklist` was ignoring functions for which the
opening parameter brace appeared on a new line. As a result many
functions were being skipped.

This patch now identifies functions whose declaration is split across
lines and therefore includes those functions in the tests.


ehuelsmann:
Wow. What a list!

ehuelsmann:
@huntorjohny, care to explain why you are approving PRs in our project?

ehuelsmann:
@huntorjohny, don't get me wrong: I like that you are checking in our project, but maybe there is a more effective way to contribute.

PR #4386, opened 2020-01-26
Use proper key for migration of assembly table

ylavoie:
Use proper key for SQL-Ledger assembly migration


ehuelsmann:
Hmm. There's really very little purpose in running the entire test suite for this commit, right? Nothing in the test suite tests this specific functionality.

ehuelsmann:
But maybe that's something you can work on? Obviously, you have test sets you can draw inspiration from...

PR #4387, opened 2020-01-26
Migrate sl32

ylavoie:
Migrate SQL-Ledger 3.2.

This PR adds support for 3.2 for everything but kits and payroll. It:
- takes advantage of multi-currency capabilities to stop limiting SL to 1 FX transaction per day;
- pushes the stable migration target to 1.7 instead of 1.6;
- adds tests to prevent migration of kits

### DIFF
--- a/sql/changes/LOADORDER
+++ b/sql/changes/LOADORDER
@@ -58,7 +58,7 @@
 # Note: the schema as created up to here, is the one as required by
 #   the migration scripts in sql/upgrade/. Don't insert any change
 #   scripts before this point, because it breaks the migrations from
-#   both LedgerSMB 1.2 & 1.3 as well as from SQL 2.8 and 3.0
+#   both LedgerSMB 1.2 & 1.3 as well as from SQL 2.8, 3.0 and 3.2
 # 1.6 changes
 1.6/drop_chart.sql
 !1.6/drop_arap_cols.sql
--- a/sql/changes/LOADORDER
+++ b/sql/changes/LOADORDER
@@ -54,7 +54,6 @@
 1.5/fixed-assets-not-null-constraints.sql
 1.5/template_menu-v2.sql
 1.5/abstract_tables.sql
-#tag: migration-target
 # Note: the schema as created up to here, is the one as required by
 #   the migration scripts in sql/upgrade/. Don't insert any change
 #   scripts before this point, because it breaks the migrations from
@@ -99,6 +98,7 @@ mc/new-constraints.sql
 mc/dropped-objects.sql
 mc/views.sql
 mc/delete-migration-validation-data.sql
+#tag: migration-target
 # 1.7 changes
 1.7/drop-custom-catalog.sql
 1.7/alter-menu-attributes.sql
--- a/sql/upgrade/sl3.0.sql
+++ b/sql/upgrade/sl3.0.sql
@@ -1,8 +1,8 @@
 --Setup
 
 -- With help of a few conditional statements handled by the Template toolkit,
--- this migration file handles migration from all SQL-Ledger version up to 3.0
--- to all Ledgersmb up to 1.6
+-- this migration file handles migration from all SQL-Ledger version up to 3.2
+-- to all Ledgersmb up to 1.8
 
 -- When moved to an interface, these will all be specified and preprocessed.
 \set default_country '''<?lsmb default_country ?>'''
@@ -14,6 +14,7 @@
  */
 \set slschema '<?lsmb slschema ?>'
 \set lsmbversion '<?lsmb lsmbversion ?>'
+\set slversion '<?lsmb slversion ?>'
 
 BEGIN;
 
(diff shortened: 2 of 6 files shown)


sql/changes/LOADORDER
```diff
@@ -54,11 +54,10 @@
 1.5/fixed-assets-not-null-constraints.sql
 1.5/template_menu-v2.sql
 1.5/abstract_tables.sql
-#tag: migration-target
```
ehuelsmann: The text block below belongs to the `migration-target` tag. You're moving the tag, but editing the text of the textblock without moving it with the tag?

Also, since this change implements SL3.2 migration, shouldn't the "old" tag stay where it is (to support the old migrations)?

sql/upgrade/sl3.0.sql
```diff
@@ -458,12 +465,21 @@ UPDATE :slschema.vendor SET entity_id = (SELECT id FROM entity WHERE 'V-' || ven
 
 UPDATE :slschema.customer SET entity_id = coalesce((SELECT min(id) FROM entity WHERE 'C-' || customernumber = control_code), entity_id);
 
+--TODO: Do we need that for MC?
+INSERT INTO defaults(setting_key,value)
+    SELECT 'curr',array_to_string(array_agg(curr),':')
+    FROM :slschema.curr;
```
ehuelsmann: Yes, we need a currency setting in the defaults table, but not like this: the company's functional currency is in the defaults table after MC. The other currencies are in the currency table.

sql/upgrade/sl3.0.sql
```diff
@@ -903,60 +910,71 @@ ALTER TABLE gl ENABLE TRIGGER gl_audit_trail;
...
---TODO: Handle amount_tc and netamount_tc
 insert into ar
-(entity_credit_account, person_id,
+        (entity_credit_account, person_id,
         id, invnumber, transdate, taxincluded,
         amount_bc, netamount_bc,
         amount_tc, netamount_tc,
 <?lsmb IF VERSION_COMPARE(lsmbversion,'1.6') < 0; ?>
         paid, datepaid,
 <?lsmb END; ?>
```
ehuelsmann: The following SELECT doesn't have a conditional block like this. Do you expect this PR to be backported to 1.6? If not, wouldn't it be better not to have any conditionals on the target schema? (It's complex enough as it is when you want to support multiple origin schemas...)

sql/upgrade/sl3.0.sql
```diff
@@ -903,60 +910,71 @@ ALTER TABLE gl ENABLE TRIGGER gl_audit_trail;
...
-        ar.id, invnumber, transdate, ar.taxincluded, amount, netamount,
-        CASE WHEN exchangerate IS NOT NULL THEN amount/exchangerate ELSE 0 END,
-        CASE WHEN exchangerate IS NOT NULL THEN netamount/exchangerate ELSE 0 END,
+        ar.id, invnumber, transdate, ar.taxincluded,
+        amount, netamount,
+        CASE WHEN exchangerate IS NOT NULL AND exchangerate <> 1 THEN amount/exchangerate ELSE 0 END,
+        CASE WHEN exchangerate IS NOT NULL AND exchangerate <> 1 THEN netamount/exchangerate ELSE 0 END,
 <?lsmb IF VERSION_COMPARE(lsmbversion,'1.6') < 0; ?>
         paid, datepaid,
 <?lsmb END; ?>
```
ehuelsmann: oh. The conditional *is* there. The question about different versions of schemas still applies though.

sql/upgrade/sl3.0.sql
```diff
@@ -903,60 +910,71 @@ ALTER TABLE gl ENABLE TRIGGER gl_audit_trail;
...
+        terms, description)
 SELECT
         vendor.credit_id,
         (select entity_id from :slschema.employee
                 WHERE id = ap.employee_id),
         ap.id, invnumber, transdate, ap.taxincluded, amount, netamount,
-        CASE WHEN exchangerate IS NOT NULL THEN amount/exchangerate ELSE 0 END,
-        CASE WHEN exchangerate IS NOT NULL THEN netamount/exchangerate ELSE 0 END,
+        CASE WHEN exchangerate IS NOT NULL AND exchangerate <> 1 THEN amount/exchangerate ELSE 0 END,
+        CASE WHEN exchangerate IS NOT NULL AND exchangerate <> 1 THEN netamount/exchangerate ELSE 0 END,
```
ehuelsmann: Should I conclude that you're setting `amount_tc` and `netamount_tc` to zero when the transaction is in the default currency? If that's the case, that's not how LedgerSMB's own schema migration should work: when the transaction is in the functional currency, the `netamount_bc == netamount_tc`.

sql/upgrade/sl3.0.sql
```diff
@@ -903,60 +910,71 @@ ALTER TABLE gl ENABLE TRIGGER gl_audit_trail;
...
 <?lsmb IF VERSION_COMPARE(lsmbversion,'1.6') < 0; ?>
         paid, datepaid,
 <?lsmb END; ?>
-        duedate, invoice, ordnumber, ap.curr, ap.notes, quonumber,
-        intnotes,
+        duedate, invoice, ordnumber,
+        CASE WHEN exchangerate IS NOT NULL THEN ap.curr ELSE NULL END,
```
ehuelsmann: Actually, `curr` should always be filled. When `exchangerate IS NULL`, it should be the functional currency indicator.

ehuelsmann:
More than 30 days without activity...

PR #4425, opened 2020-02-14
Improve waiting for pages

ehuelsmann:
This improvement is required for the refactored authentication branch to be
able to complete its tests successfully; apparently, there are new timing
issues in that branch exposing existing timing issues in the BDD tests.


nick-prater:
As it stands, and as you suggested in chat, this PR illustrates the need for a variant of "When I press ..." which waits for new content to be loaded. That would be cleaner and more natural than adding all the extra `wait for the page to load` steps.

How does this PR address the timing issues that you found?  Presumably these issues were causing false failures in tests?

PR #4470, opened 2020-02-29
Constrain transdate and approved fields on transaction tables

nick-prater:
This PR enforces stricter data integrity and fixes an issue on 1.8 with reconciliation reports
displaying a zero opening balance for older (pre-1.4) reports.

It constrains `transdate` and `approved` columns to be `NOT NULL` for the
following tables:

  * gl
  * ar
  * ap
  * acc_trans
  * transactions

Migration checks are used to ensure these constraints can be applied:

* Ensure gl, ar, ap and acc_trans tables do not contain null values for
  `transdate` or `approved`. This has never been seen in a company database,
  but if found, the upgrade process offers a way to fill-in the missing data.

* Offer to automatically fill any missing `transaction` approval state, from
  the associated gl, ar, ap tables.

* Offer to automatically clear orphaned invoice rows. These are invoice lines
  having no date and being linked to no transaction. As such they have no
  financial impact and cannot be displayed through the user interface.
  The schema has unfortunately allowed such rows to exit and they must be
  removed to enforce a `NOT NULL` constraint on `transactions.transdate`.

* Offer to automatically clear orphaned `transactions` table rows. These are
  empty rows having no links to anything else. As they contain no data,
  they can be removed without financial impact.


ehuelsmann:
Thanks! Very nice cleanup!

PR #4471, opened 2020-02-29
Drop `queue_payments` configuration setting

nick-prater:
Removes non-functioning queue_payments setting from the `defaults`
table and the LedgerSMB::DBObjext::Payment::post_bulk()
method. This was the only place the setting was referenced and the
sql code to support it did not exist.

### DIFF
--- a/old/lib/LedgerSMB/DBObject/Payment.pm
+++ b/old/lib/LedgerSMB/DBObject/Payment.pm
@@ -706,22 +706,7 @@ partial payment (C<paid == 'some'>).
 
 sub post_bulk {
     my ($self, $data) = @_;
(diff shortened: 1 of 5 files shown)


old/lib/LedgerSMB/DBObject/Payment.pm
```diff
@@ -750,16 +735,12 @@ sub post_bulk {
         }
         $self->{transactions} = $invoice_array;
         $self->{source} = $contact->{source};
-        if ($queue_payments){
-            $self->{batch_class} = BC_PAYMENT;
```
ehuelsmann: `BC_PAYMENT` is now unused and can be removed y too.

PR #4472, opened 2020-03-01
Fix missing transaction date creating bulk payment batches

nick-prater:
Creating bulk payment batches was failing when tighter constraints
are being applied to the database because `acc_trans` cannot have a null
`transdate`.

This patch ensures that a payment date (usually same as the batch date)
is passed to the database when creating a payment batch and ensures
that the `acc_trans` rows have a valid `transdate`.

The patch makes explicit which parameters are being passed to the
database and avoids polluting the request object with data private
to the post_bulk method.

### DIFF
--- a/old/lib/LedgerSMB/DBObject/Payment.pm
+++ b/old/lib/LedgerSMB/DBObject/Payment.pm
@@ -707,7 +707,6 @@ partial payment (C<paid == 'some'>).
 sub post_bulk {
     my ($self, $data) = @_;
 
-    #$self->{payment_date} = $self->{datepaid};
     for my $contact (grep { $_->{id} } @{$data->{contacts}}) {
         my $invoice_array = "{}"; # Pg Array
         for my $invoice (@{$contact->{invoices}}) {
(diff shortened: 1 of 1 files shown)


old/lib/LedgerSMB/DBObject/Payment.pm
```diff
@@ -733,11 +732,21 @@ sub post_bulk {
...
-        $self->{source} = $contact->{source};
 
-
-        $self->call_dbmethod(funcname => 'payment_bulk_post');
+        $self->call_dbmethod(
+            funcname => 'payment_bulk_post',
+            args => {
+                transactions => $invoice_array,
+                batch_id => $self->{batch_id},
+                source => $contact->{source},
```
ehuelsmann: While I understand what you are doing here, I would say the intent of `call_dbmethod` is to add database-side "methods" to the object. To that extent, I would expect that we should only provide the `transactions` and `source` keys as additional parameters.

nick-prater: Thanks for the feedback. I hadn't appreciated that some parameters can be provided via `args` while others can be taken from object properties. That makes perfect sense, but I'd seen it as an either/or situation. Patch applied...

ehuelsmann: I haven't either, but that's definitely how it should work. If it doesn't then I/we should fix PGObject::Simple.

PR #4490, opened 2020-03-07
Constrain transactions.table_type field

nick-prater:
Stricter database constraints to ensure that the `transactions`.`table_type`
field has to be one of the valid values (`ap`|`ar`|`gl`) and cannot be `NULL`.


nick-prater:
Mixes up table_type/table_name in file naming and description.
Rejecting and submitting a new PR.

PR #4493, opened 2020-03-08
Recon cleanup

nick-prater:
More clean-up, simplification, documentation and separation of responsibilities for
Reconciliation Report generation.

Standardise initialisation, encapsulate calculations within DBObject mapper, encapsulate
display and formatting concerns within Script module.

### DIFF
--- a/lib/LedgerSMB/Scripts/recon.pm
+++ b/lib/LedgerSMB/Scripts/recon.pm
@@ -215,23 +215,17 @@ sub _display_report {
     _set_sort_options($recon, $request);
 
     $recon->get;
-
-    my $neg_factor = ($recon->{account_info}->{category} =~ /^[AE]/) ? -1 : 1;
-    $recon->{their_total} *= $neg_factor;
+    _process_upload($recon, $request) unless $recon->{submitted};
+    $recon->build_totals;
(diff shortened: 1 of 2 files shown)


lib/LedgerSMB/Scripts/recon.pm
```diff
@@ -215,43 +215,31 @@ sub _display_report {
...
         }
     }
 
-    for my $field (qw/ cleared_total outstanding_total statement_gl_calc their_total variance /) {
-        $recon->{$field} = $recon->{$field}->to_output(money => 1);
-    }
-
-    for my $field (qw/ our_total beginning_balance /) {
-        $recon->{$field} ||= LedgerSMB::PGNumber->from_db(0);
+    for my $field (qw/ cleared_total outstanding_total statement_gl_calc their_total variance our_total beginning_balance /) {
```
ehuelsmann: It's not the end of the world, but it'd be nice (for code that's not explicitly deprecated) to try and stick to 80-character width. If nothing else, many brail readers have that width, even today. (The something else would be: I've always understood that the 80-character width fits better with the human brain...)

ehuelsmann:
Merging. I like the cleanup!

PR #4494, opened 2020-03-08
Upgrade TravisCI testing to Xenial

ehuelsmann:


ehuelsmann:
Bionic runs our test-suite just fine. We'll upgrade to that.

PR #4511, opened 2020-03-15
Remove the need for UPDATE permissions on acc_trans for payment reversal

ehuelsmann:
Note that this issue was found because the BDD test user doesn't have
UPDATE permissions on acc_trans (which means no other user will have it
either, because that user has full permissions...).

Note also that by changing the order in which we do things, the UPDATE
permission isn't required and we can insert the correct data the first
time around.


ehuelsmann:
Tested on 1.7, but this doesn't seem to be a problem there. (I didn't verify the code to see if it runs an update statement.)

PR #4518, opened 2020-03-21
Before, we left enabling/disabling "the back button" (caching) to the user

ehuelsmann:
Lets decide for the user that they truely don't want to cache our pages,
because they are application responses, not just static HTML pages. These
application responses potentially contain confidential information. Lets
err on the safe side instead of leaving it to our users.


ehuelsmann:
@ylavoie points out to regularly use the functionality that the pages *are* cached.

The fact that  that would be required, is something we should work on...

PR #4524, opened 2020-03-31
Turn off autocomplete values on login form

ylavoie:
Turn off autocomplete attributes for the moment.
Autocomplete attributes help password managers to infer the purpose of a field in a form, saving them from accidentally saving or autofilling the wrong data.
“username” and “current-password” should be used if we turn on autocomplete.
[skip_ci]

### DIFF
--- a/UI/login.html
+++ b/UI/login.html
@@ -58,7 +58,8 @@
                          title=text('User Name')
                          value = login
                          accesskey="l"
-                         tabindex=1 } #' ?>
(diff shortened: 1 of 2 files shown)


ehuelsmann:
@ylavoie hi. were you going to add the password inputs of the "create new user" and "reset password" pages of setup.pl (and the password reset of the HR>Employee>User page of the main  app)?

PR #4533, opened 2020-04-12
Implement LedgerSMB::PGTimestamp and fix single-item timecards

ehuelsmann:
The blocking item for fixing timecards has been the missing Timestamp type all along. Finally implement it to fix (single item) timecards.

### DIFF
--- a/lib/LedgerSMB/PGTimestamp.pm
+++ b/lib/LedgerSMB/PGTimestamp.pm
@@ -50,13 +50,10 @@ sub _stringify {
 
 =head1 CONSTRUCTOR SYNTAX
 
-LedgerSMB::PgDate->new({ date => DateTime->new(year => 2012, day => 31, month =>
-12)});
+Note the constructor is private, and not intended to be called directly.
 
-Note the constructor here is private, and not intended to be called directly.
-
-Use from_db and from_input methods instead since these handle appropriately
-different formats and handle construction differently.
+Use from_db and from_input methods instead since these handle different
+formats appropriately and handle construction
 
 =cut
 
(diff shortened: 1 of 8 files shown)


lib/LedgerSMB/PGTimestamp.pm
```diff
@@ -0,0 +1,223 @@
...
+=head1 SUPPORTED FORMATS
+
+### TO DOCUMENT
+
+=cut
+
+
+=head1 CONSTRUCTOR SYNTAX
+
+LedgerSMB::PgDate->new({ date => DateTime->new(year => 2012, day => 31, month =>
```
nick-prater, line 53: Should this be LedgerSMB::PGTimestamp rather than LedgerSMB::PGDate?

lib/LedgerSMB/PGTimestamp.pm
```diff
@@ -0,0 +1,223 @@
...
+sub from_input{
+    my ($self, $input) = @_;
+
+    local $@ = undef;
+    return $input if eval {$input->isa(__PACKAGE__)} && $input->is_date;
+
+    return __PACKAGE__->new()
+        if ! $input; # matches undefined as well as ''
+
+    my $dt = eval { strptime('%F %T', $input) }
```
nick-prater, line 155: The pod for this function says:

`Parses this from an input string according to the user's dateformat,
unless C<$format> has been specified to override it.`

But here we're parsing to a hard-coded format and don't seem to pay attention to `$format`.

lib/LedgerSMB/PGTimestamp.pm
```diff
@@ -0,0 +1,223 @@
...
+        $fmt = '%F';
+    }
+    $fmt = $formats->{uc($fmt)}->[0] if defined $formats->{uc($fmt)};
+
+    $fmt .= ' %T' if $self->is_time();
+    $fmt =~ s/^\s+//;
+
+    my $formatter = DateTime::Format::Strptime->new(
+             pattern => $fmt,
+              locale => 'en_US',
```
nick-prater, line 189: Is it right that we're hard-coding a `en_US` locale?

lib/LedgerSMB/PGTimestamp.pm
```diff
@@ -0,0 +1,223 @@
...
+sub to_output {
+    my ($self) = @_;
+    return '' if not $self->is_date();
+    my $fmt;
+    if (defined $LedgerSMB::App_State::User->{dateformat}){
+        $fmt = $LedgerSMB::App_State::User->{dateformat};
+    } else {
+        $fmt = '%F';
+    }
+    $fmt = $formats->{uc($fmt)}->[0] if defined $formats->{uc($fmt)};
```
nick-prater, line 182: $formats->{$fmt} gives a list of possible patterns, but we're only trying the first of them here. Should we be testing all of the options?

For example 'DD-MM-YYYY' maps to  ['%d-%m-%Y', '%d-%m-%y'], which I tests for both 4 and 2 digit years. Albeit that is rather surprising behaviour and I'm not sure that we should be supporting 2 digit years as input.

ehuelsmann: Completely agreed. I think 2-digit years are too confusing. I'll PR for PGDate seperately, along the same lines.

lib/LedgerSMB/PGTimestamp.pm
ehuelsmann: As the constructor is private anyway: simply removed the invocation docs -- you're not supposed to invocate it anyway.

lib/LedgerSMB/PGTimestamp.pm
ehuelsmann: Yup. Thanks for pointing that out. I've removed the remark about the user's date format, replacing it by what we *do* support.

lib/LedgerSMB/PGTimestamp.pm
ehuelsmann: To document here: we concluded that there's no problem with the locale as we only support numeric input formats and the locale is used to lookup the names of the months.

PR #4537, opened 2020-04-13
Remove <?lsmb ... ?> expansion in defaults variables

ehuelsmann:
Due to dropping this functionality (an invoice number which says
'<?lsmb DATE ?>-0000123-VOID', anyone?), the split_date function
can be dropped too.


old/lib/LedgerSMB/Form.pm
```diff
@@ -3132,89 +3122,9 @@ sub update_defaults {
             $incnum = $num + 1;
         }
 
-        s/$num/$incnum/;
+        $dbvar =~ s/$num/$incnum/;
```
nick-prater, line 3125: This maintains existing behaviour, but we ought only to be replacing the last number in the reference, otherwise this will have some odd corner-case failures.

If $dbvar = 'INV-100-100'
we expect $incnum to be 'INV-100-101', but it will instead be 'INV-101-100'

nick-prater: Maybe make the substitution regex something like `$dbvar =~ s/$num(\D*)$/$incnum$1/`

ehuelsmann: Replaced with a more elaborate matching criterion for which I hope there to be exactly 1.

PR #4545, opened 2020-04-15
Doc templates no escape

ehuelsmann:
(Document) templates should do their own escaping

### DIFF
--- a/lib/LedgerSMB/Template/LaTeX.pm
+++ b/lib/LedgerSMB/Template/LaTeX.pm
@@ -105,6 +105,7 @@ sub initialize_template {
 
     my %options = ( format => $config->{_format} );
     Template::Plugin::Latex->new($template->context, \%options);
+    $template->context->define_filter('escape', \&escape);
 
     return undef;
 }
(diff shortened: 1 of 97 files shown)


ehuelsmann:
@nick-prater I wonder how to adjust the tests in t/04-template-handling, now that `LedgerSMB::Template` and `LedgerSMB::Template::UI` are different beasts with their own TT invocations. Could you review and advise?

lib/LedgerSMB/Template/LaTeX.pm
```diff
@@ -105,6 +105,7 @@ sub initialize_template {
 
     my %options = ( format => $config->{_format} );
     Template::Plugin::Latex->new($template->context, \%options);
+    $template->context->define_filter('escape', \&escape);
```
nick-prater, line 108: Rather than using our own LaTeX escape routine, can we use the latex_encode filter thats part of Template::Plugin::Latex?

ehuelsmann:
@nick-prater , @ylavoie : based on our discussions on the Matrix chat, I'm concluding that this PR is one bridge too far. The benefit (on our side) insufficiently outweighs the cost (on our users' side) and we have no way to mitigate the cost on the user, since we haven't found a good way to automate the change that needs to be implemented.

ehuelsmann:
therefore I think this PR should be closed.

ehuelsmann:
Closing: too complex.

PR #4559, opened 2020-04-25
Fix typo originating from the mass CSS restructuring

ehuelsmann:


UI/css/ledgersmb-common.css
```diff
@@ -159,7 +159,7 @@ a.listheading:visited { color: var(--a-listheading-link-clr); font-weight: bold;
 
 .listrow1,
 table.dynatable tr,
-.listrow0 { color: var(--listrow1-bg-clr); font-size: var(--listrow1-font-size); vertical-align: top;
+.listrow0 { color: var(--listrow1-clr); font-size: var(--listrow1-font-size); vertical-align: top;
```
ehuelsmann, line 162: these refer to "listrow1-clr" and "listrow1-font-size" yet the class is named listrow0. Is that correct???

ylavoie: According to the naming standard I used, variable-clr is for color and variable-bg-clr for background, so it definetely was an error

PR #4574, opened 2020-05-02
Resolve 'TODO passed' at the end of test runs

ehuelsmann:
Payment.pm has had its copyright fixed long ago; lets acknowledge that.


ehuelsmann:
Fails in some cases and not in others... (Hash key randomization??)

PR #4583, opened 2020-05-03
Run fewer combinations on a single test run on Travis CI

ehuelsmann:
By running fewer Perl versions (we hardly ever get different outcomes
on any of them anyway), we keep more resources for running multiple
integration runs in quick succession.


ehuelsmann:
@ylavoie confirmed on Matrix that he agrees with fewer build-combinations.

PR #4600, opened 2020-05-09
Refactor migrations

ehuelsmann:
The goal is to simplify the the overall code, but also to move the migration code out of setup.pl in order to be able to add a command-line utility to perform the same function.


ehuelsmann:
I'll agree to that, but does that cause problems for the migration?

ylavoie:
> I'll agree to that, but does that cause problems for the migration?

I would tend to remove the 150 days check, that should go into sanity. I am not comfortable in bringing invalid data through migration if cleared date is before transaction date. In our reconciliation dialog, we would (should?) refuse that situation.

ehuelsmann:
@ylavoie, but is it the task of our migration to clean data too? Or is it the task to migrate what's there, unless that's technically impossible due to data constraints? In other words: the fact is that those transactions *are* reconciled. Deleting that fact it's the onus on the user to correct data which likely already was approved by the accountant.....

PR #4606, opened 2020-05-14
Remove undefined errors in logger

ylavoie:
Make sure that the logger won't output undefined errors if called before `script_name` or `action_name` is initialized.


ehuelsmann:
When does that trigger? This should not happen.

PR #4639, opened 2020-05-23
Fix 4615

ehuelsmann:


sbts:
Can I suggest we also add upgrade scripts that will correctly add data to `business_unit_ac` for any entries in `business_unit_inv` so that legacy entries work correctly with reporting

ehuelsmann:
Instead of doing that now, I think we should do that when we remove the `business_unit_inv` table, e.g. for 1.9.

PR #4672, opened 2020-06-01
Add originally published migration script according to versioning protocol

ehuelsmann:
Follow-up to b095dcef04ae38cb782e2cc9f2fecc631e46f47f


ehuelsmann:
Fixes #4674

PR #4725, opened 2020-06-13
Make images references relatives to domain

ylavoie:
Quiet linter invalid image references by making them domain relative instead or current directory relative. HTML spec specifies that the latter is relative to the html source and linters enforce that, even though most browsers are tolerant.

### DIFF
--- a/UI/login.html
+++ b/UI/login.html
@@ -25,7 +25,7 @@
         <div class="login" align="center">
           <a href="http://www.ledgersmb.org/"
              target="_top">
-            <img src="images/ledgersmb.png"
+            <img src="/images/ledgersmb.png"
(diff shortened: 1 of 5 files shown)


ehuelsmann:
Is this commit required? I'm really not very happy that we would be required to use absolute paths in our rendered HTML: we've been able to do without, thereby being agnostic of any reverse-proxying that might be going on on a non-root path -- thus without configuration of our LedgerSMB server/Starman.

sbts:
I agree with @ehuelsmann all links should be relative.

Also, there are a number of changes in the PR that are not related to the PR title.
At least one of them changes double quotes to single, which is inconsistent with other uses in the src and PR

PR #4729, opened 2020-06-14
Follow-up to domReady removal

ehuelsmann:
Sloppy PR needed domReady-removal follow-up...


ehuelsmann:
domReady removed through a PR converting login.js to loginForm.js

PR #4784, opened 2020-07-07
Webpack production

ylavoie:
This PR completes the webpack conversion from Dojo standard building.
- Code is minified and splitted in chunks to speedup page loadings
- HTML page header is compiled and optimized, while still allowing server-side processing
- All CSS are compiled and minimized


ylavoie:
I'll reopen in a few days to keep the noise down while I'm fixing Travis

PR #4785, opened 2020-07-08
SHA proper Dojo/Dijit directories

ylavoie:
Make sure that we take the proper Dojo/Dijit directories into account with Webpack


ylavoie:
Still leaves previous Travis invalid caches

PR #4787, opened 2020-07-11
Webpack production

ylavoie:
This PR completes the webpack conversion from Dojo standard building.

Code is minified and splitted in chunks to speedup page loadings
HTML page header is compiled and optimized, while still allowing server-side processing
All CSS are compiled and minimized


lib/LedgerSMB/Template/UI.pm
```diff
@@ -78,7 +78,7 @@ sub new_UI {
         if (!defined $engine) {
             $engine = Template->new(
                 ### TODO: These should be configurable absolute paths
-                INCLUDE_PATH => [ 'UI/', 'UI/lib/' ],
+                INCLUDE_PATH => [ 'UI/js', 'UI/', 'UI/lib/' ],
```
ehuelsmann, line 81: Which line-change requires the extra `UI/js` path to be in there?

t/16-schema-upgrade-html.t
```diff
@@ -147,27 +157,32 @@ $out = html_formatter_context {
...
 <head>
     <title></title>
     <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
-    <link rel="stylesheet" href="js/dojo/resources/dojo.css" type="text/css" />
-    <link rel="stylesheet" href="js/dijit/themes/claro/claro.css" type="text/css" />
-    <link rel="stylesheet" href="css/ledgersmb.css" type="text/css" />
-    <link rel="stylesheet" href="css/setup.css" type="text/css" />
+    <link href="js/css/claro.css" rel="stylesheet">
+    <link href="js/css/ledgersmb.css" rel="stylesheet">
+    <link href="js/css/setup.css" rel="stylesheet">
```
ehuelsmann, line 170: why is `js/css/` now a directory? it's ugly to put `css/` in the `js/` namespace, don't you think?

t/16-schema-upgrade-html.t
```diff
@@ -147,27 +157,32 @@ $out = html_formatter_context {
...
+            packages: [{"name":"lsmb","location":"../lsmb"}],
+            mode: "$mode"
         };
         var lsmbConfig = {
         };
     </script>
-    <script src="js/lsmb/main.js" charset="utf-8"></script>
+    } .
+    ($mode eq "production"
+        ? q{<script src="js/manifest.js"></script><script src="js/npm.dijit.js"></script><script src="js/npm.dojo.js"></script><script src="js/npm.dojo-webpack-plugin.js"></script><script src="js/gnome~gnome2~ledgersmb~ledgersmb-blue~ledgersmb-brown~ledgersmb-common~ledgersmb-purple~ledgersmb-re.js"></script><script src="js/main.js"></script>}
```
ehuelsmann, line 183: IMO, it would be nice if these would be on separate lines for readability. Is that an option?

webpack.config.js
```diff
@@ -179,18 +191,88 @@ var pluginsProd = [
...
+    }),
 
+    new HtmlWebpackPlugin({
+        inject: false, // Tags are injected manually in the content below
+        minify: false, // Adjust t/16-schema-upgrade-html.t if prodMode is used,
+        filename: "ui-header.html",
+        excludeChunks: [...Object.keys(lsmbCSS)],
+        templateContent: ({ htmlWebpackPlugin }) =>
+            `<!-- prettier-disable -->\n` +
+            `[%#\n` +
```
ehuelsmann, line 208: With this template here, there's little point to have it in `UI/lib`; maybe we can read it from `UI/lib` instead?

t/16-schema-upgrade-html.t
ylavoie: All CSS are now compiled.

webpack.config.js
ylavoie: Nope, we'll need a Perl TT enabled loader for that

t/16-schema-upgrade-html.t
ylavoie: Nope, generated by html webpack plugin which inline its css and js outputs

t/16-schema-upgrade-html.t
ylavoie: Our UI/js is dist in webpack world, i.e. where all compiled artefacts are put. Except for most HTML files, that's the case now.

t/16-schema-upgrade-html.t
ylavoie: By having it in the config, I am able to put it in the compile bundle directly without needing a bundle and allow `html-webpack-plugin` to fix the CSS & JS required, which are hashed named to allow browser caching on user installation to work properly.

ylavoie: See https://github.com/ledgersmb/LedgerSMB/pull/4787/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR232 & https://github.com/ledgersmb/LedgerSMB/pull/4787/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR262 for injections points

t/16-schema-upgrade-html.t
ehuelsmann: ok. I still don't like the idea that all our assets are in `js/` (including css and images). Maybe we need to restructure the `UI/` directory to separate out the static assets from the UI templates?

t/16-schema-upgrade-html.t
ehuelsmann: ok. if that's how it works, then that's how it works.

webpack.config.js
ehuelsmann: it took a bit before I understood that remark.

different question: maybe we need to remove it from `UI/lib/` so there's no confusion about which content is the actual content?

t/16-schema-upgrade-html.t
ylavoie: Sources remained where they were, only compiled artefacts go there. Why should we care about their internals?

Webpack default to `/dist` (UI/js, for us, to prevent Web proxy modifications) and packs everything compiled there.

lib/LedgerSMB/Template/UI.pm
ylavoie: Producing `UI/js/ui-header.html` from the `html-webpack-plugin` template, to be able to load it with the hashed JS & CSS. See https://github.com/ledgersmb/LedgerSMB/pull/4787/commits/167f34504071e4d6b1b267ba7ebac419ecadfba7#diff-11e9f7f953edc64ba14b0cc350ae7b9dR193
Compiled files, when exist, must be picked instead of the non-compiled sources.

t/16-schema-upgrade-html.t
ehuelsmann: Hmm. Maybe we should set up a `dist/` outside of the UI directory?

ehuelsmann: or at least rename `js/` to `dist/` ?

ylavoie: It has to be inside, to let the web proxy reach it and I kept it as UI/js for the same purpose. No problem to set it as /dist as soon as html get handled by webpack.

ehuelsmann: ok.

webpack.config.js
ylavoie: If it is limited to setup, login and logout, we don't need it anymore

ylavoie: > it took a bit before I understood that remark.
> 
> different question: maybe we need to remove it from `UI/lib/` so there's no confusion about which content is the actual content?

We still inject values from Perl TT and we need Webpack to adjust the included CSS & JS file names, which are hashed for long-term caching. I cannot load a file directly because validate the syntax and abort on TT statements. But I can inject code directly and that the solution I picked for the moment.

ehuelsmann: > If it is limited to setup, login and logout, we don't need it anymore

I think it applies to all pages in `UI/setup/`, because we are not using the SPA-approach for setup.pl yet.

ylavoie: Yes it does.

PR #4790, opened 2020-07-11
Cache Travis CPAN

ylavoie:
Perl code and modules seldom change yet we reinstall everything on Travis on each build.
This PR introduces caching of the local perl5 directory to speed up builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,6 +116,7 @@ cache:
   directories:
     - $HOME/locallib
     - $HOME/dojo_archive
+    - $HOME/perl5
 
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -116,6 +116,7 @@ cache:
   directories:
     - $HOME/locallib
     - $HOME/dojo_archive
+    - $HOME/perl5
```
ehuelsmann, line 119: This is the perlbrew directory that our Travis image pulls in. We're not installing anything into that directory. I'm pretty sure this commit is incorrect. We install everything into `$HOME/locallib`.

PR #4839, opened 2020-07-30
Use only language code if country variant is not in list

ylavoie:
User preferences defaults to first entry if the user uses a country variant which is not in the database. Default to the language only if the variant is not found.


ylavoie:
System/Languages and Preferences are not showing the same list.

PR #4859, opened 2020-08-16
Database handling cleanup

ehuelsmann:
Originally started as an effort to clean up the connection-instantiation code paths, this PR cleans up all direct use of DBI (in the web application), instead connecting through LedgerSMB::Database.

With a centralized database connection setup, we can also use the database connection private storage to set up configuration data to be passed on to LedgerSMB::PGObject. This then eliminates the need for LedgerSMB::Sysconfig in the lower levels of the code (deep down the call stack). This is a goal, because eventually, *all* configuration should be passed outside-in to maintain good separation of concerns.


ehuelsmann:
Please merge this PR using a squash-merge: the individual commits are there only for review.

PR #4862, opened 2020-08-18
Use translated countries in preferences.

ylavoie:
Use user locale to show translated countries


cpanfile
```diff
@@ -26,6 +26,34 @@ requires 'JSON::MaybeXS';
...
+requires 'Locale::CLDR::Locales::Ms';
+requires 'Locale::CLDR::Locales::Nb';
+requires 'Locale::CLDR::Locales::Nl';
+requires 'Locale::CLDR::Locales::Pl';
+requires 'Locale::CLDR::Locales::Pt';
+requires 'Locale::CLDR::Locales::Ru';
+requires 'Locale::CLDR::Locales::Sv';
+requires 'Locale::CLDR::Locales::Tr';
+requires 'Locale::CLDR::Locales::Uk';
+requires 'Locale::CLDR::Locales::Zh';
```
ehuelsmann, line 56: Is there an "all" placeholder? Explicitly naming them all is likely to be a maintenance issue.

old/lib/LedgerSMB/DBObject/User.pm
```diff
@@ -101,22 +102,24 @@ sub get_option_data {
...
+
+    my %regions = %{$locale->all_regions}; # Localized countries
+    my %languages = %{$locale->all_languages()}; # Localized languages
 
-    foreach my $key ( all_language_codes() )
-    {
+    # Localize the list, making sure to keep the language key as is
+    foreach my $row ( @rows ) {
+        my ($language,$region) = split /_/, $row->{code};
+        # Use the language_country localized version if available
```
ehuelsmann, line 114: Maybe you can add a bit more explanation of what codes come from which sources and how you need to compensate each code to be able to compare them, with examples? That explains why the code does what it does.

ehuelsmann (changes requested):

cpanfile
```diff
@@ -37,6 +65,7 @@ requires 'MooX::Types::MooseLike::Base';  # for Email::Sender::Transport::SMTP w
 requires 'Moose';
 requires 'Moose::Role';
 requires 'Moose::Util::TypeConstraints';
+requires 'MooseX::ClassAttribute';
```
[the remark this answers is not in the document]
ylavoie: Locale::CLDR complains loudly if it not there

cpanfile
ylavoie: I certainly use Bundle::Locale::CLDR::Everything but that will pull at least 180 Locale::CLDR::Locales that we don't use

ehuelsmann: Ok. In that case, I think we need to add cross references to each place we store them, mentioning that adding an element in one place requires adding it in .. and .. and ... other places.

cpanfile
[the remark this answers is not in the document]
ehuelsmann: Ok. Understand it now: It depends on MooX::ClassAttribute, but our classes are Moose classes which means MooX::* gets upgraded to MooseX::* ; we should mention that that's the reason the requirement is upon *us* (instead of a failure from Locale::CLDR).

old/lib/LedgerSMB/DBObject/User.pm
ylavoie: Done

cpanfile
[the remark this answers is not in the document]
ylavoie: Set in cpanfile

cpanfile
ylavoie: Documented in Pg-database.sql and cpanfile

CPBen:
I'm working on porting these into OpenBSD right now.
Once that's done, it will be very easy to add additional languages.

Chris Bennett

PR #4880, opened 2020-08-24
Language code country cldr

ylavoie:
Move from deprecated Locale::Country and Locale::CLDR for country & language names.

### DIFF
--- a/lib/LedgerSMB/Database/Upgrade.pm
+++ b/lib/LedgerSMB/Database/Upgrade.pm
@@ -17,7 +17,7 @@ use LedgerSMB::Upgrade_Tests;
 
 use File::Temp;
 use List::Util qw( first );
-use Locale::Country;
+use Locale::CLDR;
 use Scope::Guard qw( guard );
 use Template;
 use Version::Compare;
@@ -185,11 +185,15 @@ sub _linked_accounts {
 }
 
 sub _available_countries {
+    # user locale?
+    my %regions = %{Locale::CLDR
+                    ->new(LedgerSMB::Sysconfig::language())
+                    ->all_regions};
     return [
         sort { $a->{text} cmp $b->{text} }
         map { +{ value => uc($_),
-                 text  => code2country($_) }
-        } all_country_codes()
+                 text  => $regions{uc($_)} }
+        } (keys %regions)
         ];
 }
 
--- a/lib/LedgerSMB/Database/Upgrade.pm
+++ b/lib/LedgerSMB/Database/Upgrade.pm
@@ -12,12 +12,12 @@ LedgerSMB::Database::Upgrade - upgrade routines factored out of setup.pm
 use strict;
 use warnings;
 
+use LedgerSMB::I18N;
 use LedgerSMB::Sysconfig;
 use LedgerSMB::Upgrade_Tests;
 
 use File::Temp;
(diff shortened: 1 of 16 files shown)


lib/LedgerSMB/Database/Upgrade.pm
```diff
@@ -185,11 +185,15 @@ sub _linked_accounts {
 }
 
 sub _available_countries {
+    # user locale?
```
ehuelsmann, line 188: What do you mean by this remark? Can you either elaborate or remove? (I mean, if there is a TODO here, please mention that. If there is a consideration not taken into account, but should be in the future, it's good to make that explicit.)

old/lib/LedgerSMB/DBObject/TaxForm.pm
```diff
@@ -156,6 +158,12 @@ sub get_metadata
     @{$self->{countries}} = $self->call_dbmethod(
                 funcname => 'location_list_country'
     );
+    my %regions = %{Locale::CLDR
+                    ->new($self->{_user}->{language})
+                    ->all_regions};
+    foreach (@{$self->{countries}}) {
+      $_->{name} = $regions{$_->{short_name}}
+    }
```
ehuelsmann, line 166: I see this pattern in multiple places. Would it be an idea to factor it out and put it somewhere central instead?

ylavoie: It would be a great idea! I'm just wondering where would be the best place

lib/LedgerSMB/Database/Upgrade.pm
ylavoie: I'd rather use the user's instead of Sysconfig. Just wondering what is the best for setup

ehuelsmann: Setup should use the admin's preconfigured value from Sysconfig. However, libraries this deep in the call chain should be ambivalent. If we need a language, I guess we need an initarg to specify one.

old/lib/LedgerSMB/DBObject/TaxForm.pm
ehuelsmann: How about LedgerSMB::I18N ?

ylavoie: Good! Fixing.

lib/LedgerSMB/Database/Upgrade.pm
```diff
@@ -154,7 +154,7 @@ my %migration_required_vars = (
 my %required_vars_values = (
     default_ar      => sub { _linked_accounts($_[1], 'AR') },
     default_ap      => sub { _linked_accounts($_[1], 'AP') },
-    default_country => \&_available_countries,
+    default_country => sub { LedgerSMB::I18N::get_country_list($_[0]) },
```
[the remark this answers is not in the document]
ehuelsmann: Also: what is the value of argument 0 that's being passed in? `_available_countries()` didn't take any arguments...

lib/LedgerSMB/Database/Upgrade.pm
ylavoie: Fixed

PR #4923, opened 2020-08-30
Use JSON::PP version 4.03 to suppress a warning about JSON::PP::Boolean

ehuelsmann:
Please note that Travis CI runs with this setup for ages already. Adding this to
CircleCI was much overdue.


ehuelsmann:
This change is CircleCI only. Merging without Travis completion.

PR #4926, opened 2020-08-30
Eliminate warning about sort being interpreted as a function call

ehuelsmann:
Due to the fact that sort is followed by parens, it's "upgraded" to
a function call, instead of being its special operator.


ehuelsmann:
Due to the removal of Locale::Country and Locale::Language on master, this code does now only exist on 1.8 (1.7 has yet another variant of it).

PR #5147, opened 2020-09-28
Preferences menu

### LINKED ISSUE #950: Preference change (to format or language) requires a reload to take effect
More discussion at https://sourceforge.net/p/ledger-smb/bugs/1376/

the question is whether we want to fix this by documenting it or by trying to engineer a fix at present.

I am thinking documentation is best right now because the problem and solution will change as we go to a more service-oriented interface.

ylavoie:
This PR fixes the race in updating the menus after saving preferences by triggering the update after the save is confirmed. Fixes #950.

### DIFF
--- a/UI/js-src/lsmb/MainContentPane.js
+++ b/UI/js-src/lsmb/MainContentPane.js
@@ -11,7 +11,8 @@ define([
     "dojo/request/xhr",
     "dojo/query",
     //   "dojo/request/iframe",
-    "dojo/dom-class"
+    "dojo/dom-class",
+    "dojo/_base/connect"
 ], function (
     ContentPane,
     declare,
@@ -23,7 +24,8 @@ define([
     xhr,
     query,
     //   iframe,
-    domClass
+    domClass,
+    connect
 ) {
     return declare("lsmb/MainContentPane", [ContentPane], {
         last_page: null,
@@ -124,6 +126,7 @@ define([
             domStyle.set(this.domNode, "visibility", "visible");
             domStyle.set(this.domNode, "opacity", "1");
             domClass.replace(this.domNode, "done-parsing", "parsing");
+            connect.publish("lsmb/page-fresh-content", [{}]);
         },
         set: function () {
             var newContent = null;
--- a/UI/js-src/lsmb/MainContentPane.js
+++ b/UI/js-src/lsmb/MainContentPane.js
@@ -12,7 +12,7 @@ define([
     "dojo/query",
     //   "dojo/request/iframe",
     "dojo/dom-class",
-    "dojo/_base/connect"
+    "dojo/topic"
 ], function (
     ContentPane,
     declare,
@@ -25,7 +25,7 @@ define([
     query,
     //   iframe,
     domClass,
-    connect
+    topic
 ) {
     return declare("lsmb/MainContentPane", [ContentPane], {
         last_page: null,
@@ -57,7 +57,7 @@ define([
         },
         set_main_div: function (doc) {
             var self = this;
-            var body = doc.match(/<body[^>]*>([\s\S]*)(<\/body>)?/i);
+            var body = doc.match(/<body[^>]*>([\s\S]*)(<\/body>)+?/i);
 
             if (!body) {
                 this.report_error(
@@ -126,7 +126,7 @@ define([
             domStyle.set(this.domNode, "visibility", "visible");
             domStyle.set(this.domNode, "opacity", "1");
             domClass.replace(this.domNode, "done-parsing", "parsing");
-            connect.publish("lsmb/page-fresh-content", [{}]);
+            topic.publish("lsmb/page-fresh-content");
         },
         set: function () {
             var newContent = null;
--- a/UI/users/preferences.html
+++ b/UI/users/preferences.html
@@ -70,7 +70,8 @@
                                   options = user.dateformats
                                   default_values = [ user.prefs.dateformat ]
                                   text_attr = 'format'
-                                  value_attr = 'format'
+                                  value_attr = 'format',
+                                  required = 1
                                 } %]</td>
                           </tr>
                           <tr>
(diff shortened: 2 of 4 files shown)


UI/js-src/lsmb/MainContentPane.js
```diff
@@ -124,6 +126,7 @@ define([
             domStyle.set(this.domNode, "visibility", "visible");
             domStyle.set(this.domNode, "opacity", "1");
             domClass.replace(this.domNode, "done-parsing", "parsing");
+            connect.publish("lsmb/page-fresh-content", [{}]);
```
ehuelsmann, line 129: From https://dojotoolkit.org/reference-guide/1.10/dojo/connect.html:

> As of Dojo 1.7, the preferred way of handling events is to use the new lightweight dojo/on module.

Any reason to use `connect`?

UI/users/preferences.html
```diff
@@ -80,17 +82,18 @@
...
                           <tr>
                             <th align="right"><label for="language">[% text('Language') %]</label></th>
-                            <td>[% PROCESS select_language element_data={
+                            <td>[% PROCESS select element_data={
                                 name = 'language',
                                 options = user.language_codes,
                                 default_values = [ user.prefs.language ],
                                 text_attr = 'label',
                                 value_attr = 'id',
-                                default_blank = 1,
```
ehuelsmann, line 93: Why isn't `default_blank = 1` a good idea? As long as no language has been selected, this will result in the first item being selected. I think offering an empty selection, failing to submit, is better than silently saving the first choice.

UI/users/preferences.html
```diff
@@ -80,17 +82,18 @@
...
+                                required = 1
```
ehuelsmann, line 96: Why is it required? When the user doesn't choose anything, the system default will be selected. Isn't that just fine?

UI/users/preferences.html
```diff
@@ -100,7 +103,8 @@
                                   options = user.cssfiles
                                   default_values = [ user.prefs.stylesheet ]
                                   text_attr = 'file'
-                                  value_attr = 'file'
+                                  value_attr = 'file',
+                                  required = 1
```
ehuelsmann, line 107: Same.

UI/users/preferences.html
```diff
@@ -100,7 +103,8 @@
                                   options = user.cssfiles
                                   default_values = [ user.prefs.stylesheet ]
                                   text_attr = 'file'
-                                  value_attr = 'file'
+                                  value_attr = 'file',
```
ehuelsmann, line 106: None of the other lines ends in a comma. Why should this one?

UI/users/preferences.html
```diff
@@ -69,7 +70,8 @@
                                   options = user.dateformats
                                   default_values = [ user.prefs.dateformat ]
                                   text_attr = 'format'
-                                  value_attr = 'format'
+                                  value_attr = 'format',
```
ehuelsmann, line 73: None of the other lines ends in a comma. Why should this one?

UI/users/preferences.html
```diff
@@ -80,17 +82,18 @@
                                   default_values = [ user.prefs.numberformat ],
                                   text_attr = 'format',
                                   value_attr = 'format',
+                                  required = 1
                                 } %]</td>
                           </tr>
                           <tr>
                             <th align="right"><label for="language">[% text('Language') %]</label></th>
-                            <td>[% PROCESS select_language element_data={
```
ehuelsmann, line 87: Does this mean that the select_language dropdown should be replaced everywhere?

ehuelsmann (changes requested):
Please check the remarks and adjust where appropriate.

UI/js-src/lsmb/MainContentPane.js
ylavoie: Old topic syntax. Switched to topic instead of connect.

UI/users/preferences.html
ylavoie: Agreed. Fixed

UI/users/preferences.html
ylavoie: System default can be anything. We support about 40 over the 200+ possible choices

UI/users/preferences.html
ylavoie: Fixed.

UI/users/preferences.html
ehuelsmann: Yes. It can be anything. I think that was my point before when I told you that we only have 40 languages and that we now offer a list of 200 options. So I'm not sure what impact either has on the point that I tried to make that very often a system wide configured language (in ledgersmb.conf) is likely to fit at least 80% of the users. Or that the system wide default (`en`) is going to suit a lot of users in our target market in general. Anyway: I don't think that setting the first language in the list is a good default when there's no configured language in the user's settings -- which is what the above achieves.

UI/users/preferences.html
ylavoie: Restored

UI/users/preferences.html
ylavoie: Fixed

UI/users/preferences.html
ylavoie: Fixed

UI/users/preferences.html
ylavoie: 1.8 was offering 200+ languages. 1.9 limits them to the languages listed in the database and for that, it should be required and not defaulted to whatever locale may be used on the user system or manually put in ledgersmb.conf.

ehuelsmann: Ok. But on my test system, I make new users all the time. These users use the hardcoded default language "en" because they are created from setup.pl. In the preference page, the first language in the list is selected. If I submit that page without explicitly selecting a language, the language setting changes to Basque. I was happy with the system default, but I'm not happy with Basque. What to do about this?

ylavoie: Git is _not_ showing it in this conversation but _default_blank_ is restored since your first comment.

ehuelsmann: Sorry I didn't notice that! Done hint on part of GitHub would have been practical!

PR #5178, opened 2020-10-16
Proposal on a CLI for administration

ehuelsmann:


lib/LedgerSMB/Admin/Command.pm
```diff
@@ -0,0 +1,144 @@
...
+              ));
+    return @_;
+}
+
+sub dispatch {
+    my ($self, $command, @args) = @_;
+
+    $command //= 'help';
+    die "Unknown command '$command'"
+        unless $command =~ m/^([a-zA-Z]+|run)$/;
```
ylavoie, line 54: Aren't we testing that more precisely below?

ehuelsmann: Not completely the same condition, no. I've replaced the code with something more clear, as per our discussion in the chat.

ylavoie (approved):

PR #5210, opened 2020-11-09
Move configuration of CoA and GIFI from SQL to structured data (XML)

ehuelsmann:
Moving to structured data adds the benefit of consistency checking using
the XML schema verification.
Additionally, in a later iteration, it would be possible to parse (part)
of the content of the configuration file to extract a description which
could be presented to the user.

Also, since GIFI codes are in the CoA files (they exist in the account
creation SQL), there's a fixed combination between GIFI and CoA, so,
don't offer the choice anymore: GIFI are now incorporated into the
'setup XML'.


lib/LedgerSMB/Database.pm
```diff
@@ -545,30 +545,17 @@ sub load_modules {
     return 1;
 }
 
-=head2 $db->load_coa({ country => '2-char-country-code',
-                       chart => 'name-of-chart',
-                       gifi => 'name-of-gifi',
+=head2 $db->load_sic({ country => '2-char-country-code',
                        sic => 'name-of-sic' })
 
 Loads the chart of accounts (and possibly GIFI/SIC tables) as specified in
```
sbts, line 551: Is this comment still correct?
Function name has changed to only encompass SIC....

lib/LedgerSMB/Database/Config.pm
```diff
@@ -112,7 +112,7 @@ List of available files defining a Standard of Industry Codes
 
 sub charts_of_accounts {
     ###TODO: Define a parameter to the SQL directory!!
```
sbts, line 114: Perhaps this todo should be dealt with in this change as well?

lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -1009,33 +1011,42 @@ sub select_coa {
...
-    }
 
-    if ($request->{coa_lc}){
-        if ($request->{chart}){
+        if ($request->{chart}) {
             my ($reauth, $database) = _get_database($request);
             return $reauth if $reauth;
 
-            $database->load_coa(
+            my $c = LedgerSMB::Company->new(
```
sbts, line 1027: Shouldn't instantiation of a new company happen only once and outside any conditionals?

lib/LedgerSMB/Database.pm
ylavoie: ISO-3166 alpha-2 is currently used. Should we provision for upcoming alpha-3?

locale/coa/ca/French-Quebec_General.xml
```diff
@@ -0,0 +1,893 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration xmlns="http://ledgersmb.org/xml-schemas/configuration">
+  <gifi-list>
```
ylavoie, line 3: TODO: Handle GIFI validity periods.

locale/coa/ar/Personal.xml
```diff
@@ -0,0 +1,269 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration xmlns="http://ledgersmb.org/xml-schemas/configuration">
```
ylavoie, line 2: Shouldn't the configuration be in the project? We are using ledgersmb.org as a CDN.

locale/coa/ar/Personal.xml
```diff
@@ -0,0 +1,269 @@
...
+  <coa>
```
ylavoie, line 3: Could the language be in the XML instead of only in the path? At least descriptions can be spell checked.

lib/LedgerSMB/Scripts/setup.pm
ehuelsmann: As discussed in the chat; yes, there's an entire rewrite of setup. Ok desired. This would be part of that rewrite. At the moment though, the company doesn't provide access to all functions that setup requires. So at this point, it doesn't make sense yet.

locale/coa/ca/French-Quebec_General.xml
ehuelsmann: Let's make an issue for that and discuss what that would take in that issue.

locale/coa/ar/Personal.xml
ehuelsmann: Actually, the indication in the path is supposed to be a region or jurisdiction, not a language name. That manifests in e.g. the chart for Canada which has the same chart, but in 2 languages.
But yes, we can add the language to the xml. I'll need to create an attribute for that then. Do we know the language of each of the provided CoAs?

locale/coa/ar/Personal.xml
ehuelsmann: The URL for a schema need not hold a document, but first and foremost needs to be unique. Using our own domain as the basis guaranteed that we are the only ones using the namespace.

bin/scan-coa-to-xml
```diff
@@ -0,0 +1,365 @@
...
+        my $code = strip_sq($m{ACCNO});
+        my $h = first { $code gt $_->{code} } @h;
+        $h->{elem}->appendChild( $a );
+    }
+    else {
+        die;
+    }
+}
+
+while ($orig_content =~ m/insert\s+into\s+currency\s*\(\s*((?&LITT_PAT))\s*,\s*((?&LITT_PAT))\s*\)\s*
```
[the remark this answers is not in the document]
ehuelsmann: There most likely is: there's often text or comments in the remainder of the file. This script was committed, but I'm not sure it should stay: it was developed to migrate or existing schema to xml, but wasn't intended to work robustly for every coa: with each coa added to the scope of the migration had to tweak the script to parse the intricacies of that file (while retaining support for the others)

lib/LedgerSMB/Database.pm
ehuelsmann: Do we want/need to move to the 3-char codes? (I'm fine with that, just asking if there's a necessity.)

Note that although these are country codes, they are used as "jurisdiction codes" in the sense that SIC can be different between jurisdictions. (Or, in case of the EU, the same across multiple jurisdictions.)

bin/scan-coa-to-xml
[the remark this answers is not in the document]
ehuelsmann: Should we remove this script from the final merge? Or move it to `utils/`? Or ...?

locale/coa/ar/Personal.xml
ehuelsmann: The configuration *is* in the project, really: https://github.com/ledgersmb/LedgerSMB/blob/master/doc/company-setup/configuration.xsd

ehuelsmann: (Could you mark this conversation "resolved" if that answers your question/addresses your point?)

lib/LedgerSMB/Database/Config.pm
ehuelsmann: It needs to be done and soon, but I think that as this change moves away from the SQL directory, this isn't for this change.

ehuelsmann:
Lets continue discussion on further refinements here and in the chat channel. I'm going to merge as there's nothing obviously wrong with the current PR as it stands.

PR #5222, opened 2020-11-15
Fix node module versions1

ylavoie:
This PR makes sure that the CI use the same exact node modules as were used in development.
It also removes fsevents which is only need on MACOS but was generating warnings when absent.


ehuelsmann (approved):
Thanks! Looks fine.

PR #5261, opened 2020-12-18
Selenium hub GitHub actions

ylavoie:
Allows testing with Chrome, Firefox, Opera & PhantomJS on CircleCI & GitHub

Note: Testing on Firefox required new Weasel & Weasel-Dojo-Widget for W3C compatibility and to alllow objects behind their label to be clickable

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,12 +6,12 @@ name: CI
 on:
   # Triggers the workflow on push or pull request events but only for the master branch
   push:
-    branches: '*'
+    branches: ['*']
   pull_request:
     branches: '*'
 
   # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
+  # workflow_dispatch:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
@@ -24,8 +24,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        perl: [ "5.24", "5.28", "5.32" ]
+        include:
+          - perl: "5.32"
+            BROWSER: "chrome"
+            SELENIUM: "selenium/standalone-chrome:3.141.59"
+          - perl: "5.28"
+            BROWSER: "firefox"
+            SELENIUM: "selenium/standalone-firefox:3.141.59"
+          - perl: "5.24"
+            BROWSER: "phantomjs"
+            SELENIUM: "ylavoie/standalone-phantomjs:latest"
     services:
+      # Label used to access the service container
       postgres:
         # Docker Hub image
         image: ledgersmb/ledgersmb-dev-postgres
@@ -58,7 +68,7 @@ jobs:
           # Version Spec of the version to use.  Examples: 12.x, 10.15.1, >=10.15.0
           node-version: 15.x
 
-      - name: Cache node modules
+      - name: Restore cached node modules
         uses: actions/cache@v2
         with:
           path: ~/.npm
@@ -66,7 +76,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-node-
 
-      - name: Cache cpan modules
+      - name: Restore cached cpan modules
         uses: actions/cache@v2.1.3
         with:
           # A list of files, directories, and wildcard patterns to cache and restore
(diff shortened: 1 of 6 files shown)


ehuelsmann:
There will be more review items, but in general: since TravisCI isn't doing any testing anymore for us, shouldn't we remove `.travis-ci.yml` instead of updating it?

t/12-dbconfig.t
```diff
@@ -4,10 +4,10 @@ use File::Spec;
 use Test::More;
 
 
-use LedgerSMB::Database::Config;
 use LedgerSMB::Sysconfig;
 
 LedgerSMB::Sysconfig->initialize( $ENV{LSMB_CONFIG_FILE} // 'ledgersmb.conf' );
+use LedgerSMB::Database::Config;
```
ehuelsmann, line 10: Is the move of this line necessary? If so, why?

xt/lib/PageObject/App/Invoices/Line.pm
```diff
@@ -75,7 +75,7 @@ sub field_value {
                       \@name="$field_map{$label}_${id}"]});
     die "Invoice line column $field_map{$label}_${id} not found"
         if not defined $field;
-    my $rv = $field->value;
+    my $rv = $field->get_attribute('value');
```
ehuelsmann, line 78: `$field->value` means to use the Dojo/HTML widget's `value()` method. Why change it to a more low-level method?

xt/lib/Pherkin/Extension/pageobject_steps/invoice_steps.pl
```diff
@@ -279,7 +279,7 @@
         my $invoice = S->{ext_wsl}->page->body->maindiv->content;
         my @lines = grep { $_->is_empty } @{$invoice->lines};
         my $empty = $lines[$empty_line_no - 1];
-        my $part_selector = $empty->get_property('partnumber');
+        my $part_selector = $empty->get_attribute('partnumber');
```
ehuelsmann, line 282: Unrelated to your change?

xt/lib/PageObject/App/Main.pm
```diff
@@ -55,7 +55,8 @@ sub _wait_for_valid_content {
     $self->session->wait_for(
         sub {
             my $class = $self->get_attribute('class');
-            return scalar( grep { $_ eq 'done-parsing' }
+            return defined($class)
```
ehuelsmann, line 58: Nice find, but unrelated. Maybe PR separately?

.github/workflows/main.yml
```diff
@@ -50,6 +50,10 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
+      - name: Cancel Previous Runs
```
ehuelsmann, line 53: Why do we need this?

UI/main.html
```diff
@@ -21,6 +21,21 @@
         </a>
       </div>
       <div id="version_info">[% text("LedgerSMB [_1]", version) %]</div>
+      <script>
```
ehuelsmann, line 24: It would be nice to understand why we need this. I think we managed to move all JS out of the various HTML pages to dojo widgets; this moves us away from that again. Can it be done in a/the dojo widget?

.github/workflows/main.yml
```diff
@@ -75,44 +92,68 @@ jobs:
...
 
       - name: Other preparations
         run: |
-          mkdir screens logs
+          mkdir -p screens logs
           cp doc/conf/ledgersmb.conf.default ledgersmb.conf
 
       - name: Starting 'starman'
         run: starman --preload-app --pid starman.pid --workers 2 -Ilib -Iold/lib --port 5762 bin/ledgersmb-server.psgi &
+        # Use plackup to add reloading capability & debug
```
ehuelsmann, line 129: This comment seems way out of place? Surely our CI machines don't need reloading capability?

ylavoie:
> There will be more review items, but in general: since TravisCI isn't doing any testing anymore for us, shouldn't we remove `.travis-ci.yml` instead of updating it?

Fixed, TravisCI configuration has been removed in another PR

xt/lib/PageObject/App/Invoices/Line.pm
ylavoie: Leftover. Restored.

xt/lib/Pherkin/Extension/pageobject_steps/invoice_steps.pl
ylavoie: Reversed.

xt/lib/PageObject/App/Main.pm
ylavoie: Moved to a new PR

UI/main.html
ylavoie: LGTM is fixed. This should indeed be moved out of the page but, at least for a while, there should be a way to identify on which browser was the application running.

ehuelsmann: Maybe we should show it with a special URL query string argument?

ylavoie:
We moved away from Github Actions.

PR #5322, opened 2021-01-17
Migrate sl30 after mc

ylavoie:
Multi-currencies fixes must be applied before migrations


sql/changes/LOADORDER
```diff
@@ -98,6 +97,7 @@ mc/new-constraints.sql
 mc/dropped-objects.sql
 mc/views.sql
 mc/delete-migration-validation-data.sql
+#tag: migration-target
```
ehuelsmann, line 100: @ylavoie is the SL migration the only migration using this tag? Because if not, then you broke the other migrations, right?

PR #5325, opened 2021-01-18
Fix 1.3 migration

ehuelsmann:


sql/upgrade/1.3-1.5.sql
```diff
@@ -644,8 +644,6 @@ SELECT
      FROM lsmb13.defaults WHERE setting_key = 'curr')
   FROM lsmb13.jcitems
  WHERE project_id IN (select id from lsmb13.project);
-INSERT INTO custom_table_catalog SELECT * FROM lsmb13.custom_table_catalog;
```
ylavoie, line 613: custom_table_catalog was only dropped in 1.7. See #4310

ehuelsmann: Yes, but in 1.8 it was removed from pg-database.sql; that's the inmiddels part. It should still have been in the base schema.

PR #5326, opened 2021-01-18
Allow dummy trigger to actually run without causing failures

ehuelsmann:
Follow-up to 19b66405b35fe7416caf1c1340e40517fb66679e


ehuelsmann:
Decided it's better to just load the triggers module upon schema creation.

PR #5331, opened 2021-01-22
Migrate sl30 mc

ylavoie:
Allow SQL-Ledger to have a different migration start point than LedgerSMB to cope with the the fact that it does support multi-currencies since version 2.8, whereas LedgerSMB got them in 1.7. 
So LedgerSMB migration start point is before multi-currencies whereas SQL-Ledger needs to migrate with the MC fixes in place

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -27,6 +27,7 @@ Removed functionalities
 * Configuration setting 'Only Timeout Locks' removed
 
 Bugs fixed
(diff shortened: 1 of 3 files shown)


ylavoie:
No CircleCI checks?

ehuelsmann (commented):
No CI checks because they run on the branch.

PR #5347, opened 2021-01-28
Fix #5298: Spurious gutter in HTML template main table

### LINKED ISSUE #5298: 1.7 and 1.8 templates add a spurious gutter column to main table
This gutter column can cause Off-by-One column issues once the templates are processed.

**The following is from a [#ledgersmb](https://matrix.to/#/!qyoLumPqusaXqFJNyK:matrix.org/$1609747391488505ZUxWv:matrix.org?via=matrix.org&via=matrix.freelock.com&via=t2bot.io) discussion

> **[dcg]**
If you take a look at....
- https://github.com/ledgersmb/LedgerSMB/blob/1.7.25/templates/demo/letterhead.html
- https://github.com/ledgersmb/LedgerSMB/commit/2814c8d7eddfe659d26fd10bcb2179bb39737779#diff-becfc0e6e93f814b7c02a6b4f07e63124efd17897c7c9b26618d93550130b786
￼
> Line 2 of letterhead is....
```html
<td width=10>&nbsp;</td>
```

> **[ehuelsmann]**
yes

> **[dcg]**
In invoice.html the end result is adding
```html
<td> </td>
```
> to every toplevel `<tr>`
￼
> In the commit I linked to, all of the <td> </td> lines were removed from purchase_order.html
This broke printing of the PO (the header and PO were in different table columns)
as far as I can tell, the only reason that empty column exists in leterhead.html is to forma gutter (margin) to the left of the document

> If that is the case, I would strongly recommend we remove it from all html templates
￼If the gutter is really desired, then we should simply add a left margin to the outer table

> **[ehuelsmann]**
that's indeed the purpose of the original use of it indeed.
I did a lot of clean up on the HTML documents, I think post-1.8. maybe this has been resolved too.

> **[dcg]**
It's still present in 1.8
￼It's also present in 1.9
ehuelsmann: so, at the moment, we need to do one of the followning....
- revert the commit to purchase_order.html
- remove the gutter column from all other templates

> I think the correct solution would be "remove from other templates"

> **[ehuelsmann]
Agreed. We need to remove everywhere.

ehuelsmann:
Remove the gutter from all templates. I guess the original change where
I removed the gutter from the purchase order was based on misunderstanding.

### DIFF
--- a/templates/demo/ap_transaction.html
+++ b/templates/demo/ap_transaction.html
@@ -27,13 +27,11 @@ releases. No explicit versioning was applied before 2021-01-04.
   <table width="100%">
     <?lsmb INCLUDE letterhead ?>
     <tr>
-      <td width="10">&nbsp;</td>
       <th colspan="3">
         <h4 class="c1"><?lsmb text('AP Transaction') ?></h4>
       </th>
     </tr>
     <tr>
-      <td>&nbsp;</td>
       <td>
         <table width="100%" cellspacing="0" cellpadding="0">
           <tr valign="top">
@@ -88,7 +86,6 @@ releases. No explicit versioning was applied before 2021-01-04.
       <td></td>
     </tr>
     <tr>
-      <td>&nbsp;</td>
       <td>
         <table>
           <?lsmb FOREACH account ?><?lsmb loop_count = loop.count - 1 ?>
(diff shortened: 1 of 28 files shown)


ehuelsmann:
I'm unable to deduce the problem that @tonka128 might have had when we introduced the initial difference in gutter margins. Nevertheless, I can see the problem as reported by @sbts in our current code base back to 1.7.

PR #5364, opened 2021-02-23
Yath

ylavoie:
Move testing from `prove` to `yath (Test2-Harness)`, which allows a richer testing platform.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,20 +19,19 @@ commands:
     steps:
       - run:
           command: |
-             if [ "x$COVERAGE" == "x1" ]
-             then
-               export HARNESS_PERL_SWITCHES="$HARNESS_PERL_SWITCHES -MDevel::Cover=-ignore,^/|^utils/|^x?t/|\.lttc$,-blib,0"
-             fi
-             echo $PERL5OPT
-             echo $HARNESS_PERL_SWITCHES
-             echo $PWD
-             # Coverage explicitly without parallellism to prevent resource contention
-             # when Devel::Cover writes coverage database on process-exit
-             prove -j2 --recurse \
-                   --pgtap-option dbname=$PGDB \
-                   --pgtap-option username=$PGUSER \
-                   --pgtap-option psql=.circleci/psql-wrap \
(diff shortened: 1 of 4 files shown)


ehuelsmann:
I'm seeing that the screenshots are missing from the uploaded artifacts. Also, the lines in the html output suddenly have ”MATCH MODE” prepended.

ehuelsmann (changes requested):
Overall, I like this change. I have two points I'd like to make:

1. I don't like the creation of the x2 directory  
    First of all, especially on CI infrastructure, the `89-dropdb.t` test doesn't add any value to the mix. Instead of moving it to a separate directory in order to prevent it from running: let's remove it all together.  
   Second of all, I don't like the fact that the `40-dbsetup.t` test needs to be specialized like this; instead of solving the creation as part of a test script, can't we (simply?) use the new `bin/ledgersmb-admin` script to create a valid database for our testing requirements?
2. Should a CI system really be using a persistent runner? Especially since there doesn't seem to be any pre-loading going on. Along the same lines: why did you choose 5 jobs? We are running this on CircleCI who provide us with two(?) VCPUs, meaning that 5 should be way more than we've been allocated.

ylavoie:
> 1. I don't like the creation of the x2 directory
>    First of all, especially on CI infrastructure, the `89-dropdb.t` test doesn't add any value to the mix. Instead of moving it to a separate directory in order to prevent it from running: let's remove it all together.

x2 is a standard test directory for `Test2` compatibles tests. Nevertheless, `40-dbsetup` & `89-dropdb` have now been replaced by `ledgersmb/admin` in another PR.

> 2. Should a CI system really be using a persistent runner? Especially since there doesn't seem to be any pre-loading going on. Along the same lines: why did you choose 5 jobs? We are running this on CircleCI who provide us with two(?) VCPUs, meaning that 5 should be way more than we've been allocated.

`yath` separated test parsing and execution so yes, sending all tests to a persistent runner allows to preload all execution dependencies only once. It does imply that we need to rely on yath job too, instead of only PID to allow tests running on a single database to run properly.

ehuelsmann (approved):

PR #5377, opened 2021-03-20
Restore COA and DB testing

ylavoie:
Define COA_TESTING & DB_TESTING to enable tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,9 +185,7 @@ executors:
           POSTGRES_PASSWORD: test
       - image: ledgersmb/ledgersmb_circleci-<< parameters.browser >>
     environment:
-      COA_TESTING: 1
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -185,7 +185,9 @@ executors:
           POSTGRES_PASSWORD: test
       - image: ledgersmb/ledgersmb_circleci-<< parameters.browser >>
     environment:
+      COA_TESTING: 1
```
ehuelsmann, line 188: These were disabled for coverage test runs, because coverage tests already run very long. Maybe it's an idea to keep these disabled on the coverage run? (Same for the DB_TESTING parameter.)

ylavoie: Moved each to a specific job

PR #5384, opened 2021-03-20
Admin destroy

ylavoie:
Implement `ledgersmb-admin destroy` to replace `bin/drop-roles`


lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,110 @@
...
+        connect_data => $connect_data,
+        );
+    my $dbh = $db->connect;
+    $dbh->do(q{SELECT setting__set('role_prefix',
+                             coalesce((setting_get('role_prefix')).value, ?))},
+                 {},
+                 "lsmb_$existing_db->{dbname}__")
+        or die $dbh->errstr;
+    $dbh->commit;
+    $dbh->disconnect;
```
ehuelsmann, line 38: Why are you *setting* the value of `role_prefix` if you're about to delete the database?

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,110 @@
...
+        $self->config->get('connect_data')->%*,
+        $existing_db->%*,
+        dbname => ($self->config->get('admindb') // 'postgres'),
+    };
+    $self->db(
+        LedgerSMB::Database->new(
+            connect_data => $connect_admin,
+        ));
+    $dbh = $self->db->connect;
+    my $sth = $dbh->prepare("SELECT rolname FROM pg_roles WHERE rolname LIKE 'lsmb_$existing_db->{dbname}__%';");
```
ehuelsmann, line 54: But what if the database had `role_prefix` set? In that case, the roles you're trying to query for here, don't exist. (Mind you: it's not guaranteed that the roles are shared with another database, even when role_prefix is set...)

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,110 @@
...
+=head1 METHODS
+
+=head2 run(@args)
+
+Runs the C<destroy> command, according to the C<LedgerSMB::Admin::Command>
+protocol.
+
+=head1 LICENSE AND COPYRIGHT
+
+Copyright (C) 2020 The LedgerSMB Core Team
```
ehuelsmann, line 106: Maybe we should update that line :-)

ehuelsmann (changes requested):

lib/LedgerSMB/Admin/Command/destroy.pm
ylavoie: Indeed. That had to be a get only.

lib/LedgerSMB/Admin/Command/destroy.pm
ylavoie: Fixed

lib/LedgerSMB/Admin/Command/destroy.pm
ylavoie: Rewritten to delete only if roles are not used anymore.

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,176 @@
...
+        $existing_db->%*,
+        dbname => ($self->config->get('admindb') // 'postgres'),
+    };
+    $db = LedgerSMB::Database->new(
+            connect_data => $connect_admin,
+        );
+    my $dbh = $db->connect;
+
+    # Scan databases which could be using the same profile
+    my $sth = $dbh->prepare("SELECT datname FROM pg_database WHERE NOT datistemplate");
```
ehuelsmann, line 52: This causes the failure of the critic tests: there is no interpolation, which means you need to use single quotes.

ylavoie: Fixed

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,172 @@
...
+        dbname => ($self->config->get('admindb') // 'postgres'),
+    };
+    $db = LedgerSMB::Database->new(
+            connect_data => $connect_admin,
+        );
+    my $dbh = $db->connect;
+
+    # Scan databases which could be using the same profile
+    my $sth =
+        $dbh->prepare('SELECT datname FROM pg_database WHERE NOT datistemplate');
```
ehuelsmann, line 53: The `$db` object has a `list_dbs` method. Any reason not to use that? (Sorry to notice after my previous review).

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,172 @@
...
+        $self->config->get('connect_data')->%*,
+        $existing_db->%*,
+    };
+    my $db = LedgerSMB::Database->new(
+        connect_data => $connect_data,
+        );
+    my $role_prefix = _get_role_prefix($db);
+
+    $db->drop
+        or die 'Cannot drop database';
```
ehuelsmann, line 38: `drop` never returns false. It returns true or throws an exception.

lib/LedgerSMB/Admin/Command/destroy.pm
ylavoie: Because list_dbs doesn't filter out templates

lib/LedgerSMB/Admin/Command/destroy.pm
ylavoie: Fixed

lib/LedgerSMB/Admin/Command/destroy.pm
ehuelsmann: ok. that's far from ideal, but not something to resolve in this PR.

ehuelsmann (approved):

PR #5386, opened 2021-03-25
Fix "Import Goods" from csv file.

Lianto2020:
Thank you for working on the LedgerSMB code base and wanting to create
a pull request.

For the easiest processing and the best maintenance of the series of
commits ('line of history'), we ask you to take a few things into
consideration. Please:

 * Deal with a single topic in a single PR
   To clarify: fix a single bug or address a single problem in one PR
 * Whenever related, group changes to multiple files into a single commit
   That means: a single search-and-replace action across all files go into
   a single commit
 * Your commits will be become part of the line of commits of the project
   Please create and group your commits in a way that is helpful when
   someone later needs to research 'where does this line come from and why
   did it change'
 * Write tests to validate your change and make sure it continues to work
   in the future
 * Our code base uses a maximum line-width of 80 characters.

By default, all PRs will be tested using Travis CI. For documentation changes,
these tests may be skipped by putting `[skip ci]` somewhere in the commit
message.

For more coding standards and community guidelines, please consult the
links presented above this PR form.

### DIFF
--- a/lib/LedgerSMB/Scripts/import_csv.pm
+++ b/lib/LedgerSMB/Scripts/import_csv.pm
@@ -436,8 +436,9 @@ sub _process_inventory_multi {
 sub _process_parts {
     my ($request, $entries, $columns, $acc_types) = @_;
     my %table_columns =
-        map { $_ => ($_ =~ s/_accno$/_accno_id/r) }
-        map { $_ eq 'partsgroup' ? 'partsgroup_id' : $_ }
+        #LZY map { $_ => ($_ =~ s/_accno$/_accno_id/r) }
+        #LZY map { $_ eq 'partsgroup' ? 'partsgroup_id' : $_ }
+        map { $_ => ($_ =~ s/(_accno|partsgroup)$/$1_id/r )}
         grep { $_ ne 'taxaccnos' }
         @{$columns};
     my %column_placeholders =
@@ -450,7 +451,8 @@ sub _process_parts {
                                               AND al.description = ? ))'
                   : ($_ eq 'partsgroup' ?
                      '(SELECT id FROM partsgroup WHERE partsgroup = ?)'
-                     : $_))}
+        #LZY             : $_))}
+                    : '?'))}
         keys %table_columns;
 
     my $stmt =
@@ -531,7 +533,8 @@ our $process = {
     timecard        => \&_process_timecard,
     inventory       => \&_process_inventory,
     inventory_multi => \&_process_inventory_multi,
-    goods           => \&_process_goods,
+    #LZY goods           => \&_process_goods,
+    parts           => \&_process_goods,
     services        => \&_process_services,
     overhead        => \&_process_overhead,
 };


ehuelsmann:
@Lianto2020 Hi! Welcome to the LedgerSMB project!

Thank you for submitting this PR! I have a few questions. I'll submit those in a second.

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -450,7 +451,8 @@ sub _process_parts {
                                               AND al.description = ? ))'
                   : ($_ eq 'partsgroup' ?
                      '(SELECT id FROM partsgroup WHERE partsgroup = ?)'
-                     : $_))}
+        #LZY             : $_))}
```
ehuelsmann, line 454: Please don't leave the old (faulty) line in place: we have git to help us track the history of our changes (and we can always find back who changed what). In this case, it also made it harder to see the actual difference.

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -436,8 +436,9 @@ sub _process_inventory_multi {
 sub _process_parts {
     my ($request, $entries, $columns, $acc_types) = @_;
     my %table_columns =
-        map { $_ => ($_ =~ s/_accno$/_accno_id/r) }
-        map { $_ eq 'partsgroup' ? 'partsgroup_id' : $_ }
+        #LZY map { $_ => ($_ =~ s/_accno$/_accno_id/r) }
+        #LZY map { $_ eq 'partsgroup' ? 'partsgroup_id' : $_ }
+        map { $_ => ($_ =~ s/(_accno|partsgroup)$/$1_id/r )}
```
ehuelsmann, line 441: I think this is not really a change that fixes anything, right? The code is semantically 99% the same?

This change introduces an important difference (bug?) though: `s/partsgroup$/partsgroup_id/r` is not the same as `$_ eq 'partsgroup' ? 'partsgroup_id' : $_` because the `s///` form also matches `mydummy_partsgroup` whereas the latter does not.

Am I missing that this change does not really contribute to the fix? If so, would you please revert this change? (If you don't know how to do that, no problem, please contact us either here in the ticket or at https://app.element.io/#/room/#ledgersmb:matrix.org.

lib/LedgerSMB/Scripts/import_csv.pm
Lianto2020: yes, will remove the old line

lib/LedgerSMB/Scripts/import_csv.pm
Lianto2020: I found out that this:
         map { $_ => ($_ =~ s/_accno$/_accno_id/r) }
         map { $_ eq 'partsgroup' ? 'partsgroup_id' : $_ }
will make the end hash result with 
{ partsgroup_id => 'partsgroup_id' } while it should be
{ partsgroup => 'partsgroup_id' }
Do you have suggestion of other ways to fix it?

I also messed up how to use branch. Now my master is the same as fix-import-goods branch.
I am gonna have to cancel this pull and fix my branch first

PR #5390, opened 2021-04-02
New invoice content in AP/AR tests

ylavoie:
Make sure we get the desired customer or vendor before declaring the page loaded to remove some test brittleness.

### DIFF
--- a/xt/lib/PageObject/App/AP/Invoice.pm
+++ b/xt/lib/PageObject/App/AP/Invoice.pm
@@ -40,7 +40,7 @@ sub select_vendor {
     $elem->clear;
     $elem->send_keys($vendor);
 
-    $self->update($elem);
+    $self->update;
 }
 
 
--- a/xt/lib/PageObject/App/AP/Invoice.pm
+++ b/xt/lib/PageObject/App/AP/Invoice.pm
@@ -41,6 +41,7 @@ sub select_vendor {
     $elem->send_keys($vendor);
 
     $self->update;
+    $self->session->page->body->maindiv->wait_for_content(replaces => $elem);
 }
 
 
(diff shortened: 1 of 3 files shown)


xt/lib/PageObject/App/AP/Invoice.pm
```diff
@@ -41,6 +41,7 @@ sub select_vendor {
     $elem->send_keys($vendor);
 
     $self->update;
+    $self->session->page->body->maindiv->wait_for_content(replaces => $elem);
```
ehuelsmann, line 44: Nice find!

(However, I think it makes more sense to do this here: https://github.com/ledgersmb/LedgerSMB/blob/4d02bed8491eb6b4244a09b6244541a7e65d360c/xt/lib/PageObject/App/Invoice.pm#L27)

ehuelsmann (changes requested):

xt/lib/PageObject/App/Invoice.pm
```diff
@@ -21,10 +21,10 @@ sub _verify {
 }
 
 sub update {
-    my ($self) = @_;
+    my ($self,$elem) = @_;
 
     $self->find("*button", text => "Update")->click;
-    $self->session->page->body->maindiv->wait_for_content;
+    $self->session->page->body->maindiv->wait_for_content(replaces => $elem);
```
ehuelsmann, line 27: I understand what you're doing here, but the "Update" button could serve the same purpose, no?

ylavoie: Yes, if the whole page changes. What if only partial updates are done? Only the Customer/Vendor field seems to be updated.

ylavoie: Indeed. Simplified

ylavoie:
Please squash on merge.

PR #5394, opened 2021-04-07
Optimize roles transactions

ylavoie:
The PG optimizer has trouble with functions in WHERE clauses because they are considered out of scope for each query.
This PR makes sure that roles functions are called once to compensate.

### DIFF
--- a/sql/modules/Roles.sql
+++ b/sql/modules/Roles.sql
@@ -10,13 +10,18 @@ DROP FUNCTION IF EXISTS lsmb__create_role(text);
 CREATE OR REPLACE FUNCTION lsmb__create_role(in_role text) RETURNS bool
 LANGUAGE PLPGSQL AS
 $$
+DECLARE
+  t_in_role text;
 BEGIN
-  PERFORM * FROM pg_roles WHERE rolname = lsmb__role(in_role);
+  -- Make sure to evaluate the role once because the optimizer
+  -- uses it as a filter on every row otherwise
+  SELECT lsmb__role(in_role) INTO t_in_role;
+  PERFORM rolname FROM pg_roles WHERE rolname = t_in_role;
   IF FOUND THEN
      RETURN TRUE;
   END IF;
 
-  EXECUTE 'CREATE ROLE ' || quote_ident(lsmb__role(in_role))
+  EXECUTE 'CREATE ROLE ' || quote_ident(t_in_role)
   || ' WITH INHERIT NOLOGIN';
 
   RETURN TRUE;
@@ -95,8 +100,11 @@ CREATE OR REPLACE FUNCTION lsmb__grant_menu
 RETURNS BOOL
 LANGUAGE PLPGSQL SECURITY INVOKER AS
 $$
+DECLARE
+  t_in_role text;
 BEGIN
-   PERFORM * FROM pg_roles WHERE rolname = lsmb__role(in_role);
+   SELECT lsmb__role(in_role) INTO t_in_role;
+   PERFORM rolname FROM pg_roles WHERE rolname = t_in_role;
    IF NOT FOUND THEN
       RAISE EXCEPTION 'Role not found';
    END IF;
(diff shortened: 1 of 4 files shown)


sql/modules/Roles.sql
```diff
@@ -95,8 +100,11 @@ CREATE OR REPLACE FUNCTION lsmb__grant_menu
 RETURNS BOOL
 LANGUAGE PLPGSQL SECURITY INVOKER AS
 $$
+DECLARE
+  t_in_role text;
 BEGIN
-   PERFORM * FROM pg_roles WHERE rolname = lsmb__role(in_role);
+   SELECT lsmb__role(in_role) INTO t_in_role;
```
ehuelsmann, line 106: Could you repeat the comment you added above, here too?

sql/modules/admin.sql
```diff
@@ -26,11 +26,13 @@ CREATE OR REPLACE FUNCTION admin__add_user_to_role(in_username TEXT, in_role TEX
         a_role name;
         a_user name;
         t_userid int;
+        t_in_role TEXT;
     BEGIN
 
         -- Issue the grant
+        SELECT lsmb__role(in_role) INTO t_in_role;
```
ehuelsmann, line 33: And here?

sql/modules/admin.sql
```diff
@@ -139,6 +144,7 @@ CREATE OR REPLACE FUNCTION admin__get_roles_for_user(in_user_id INT) returns set
             END IF;
         END IF;
 
+        SELECT lsmb__role_prefix() INTO t_role_prefix;
```
ehuelsmann, line 147: And here?

sql/modules/admin.sql
```diff
@@ -203,6 +210,7 @@ CREATE OR REPLACE FUNCTION admin__get_roles_for_user_by_entity(in_entity_id INT)
             END IF;
         END IF;
```
ehuelsmann, line 212: And here?

sql/modules/admin.sql
```diff
@@ -463,13 +471,15 @@ DROP FUNCTION IF EXISTS admin__get_roles();
 create or replace function admin__get_roles () returns setof pg_roles as $$
 DECLARE
    u_role pg_roles;
+   t_role_prefix TEXT;
 begin
+     SELECT lsmb__role_prefix() INTO t_role_prefix;
```
ehuelsmann, line 476: And here?

sql/modules/Roles.sql
```diff
@@ -1280,10 +1289,11 @@ IF TG_OP = 'DELETE' THEN
...
    -- don't need access enforcement
-   IF pg_has_role((select rolname
-                     from pg_database db inner join pg_roles rol
-                       on db.datdba = rol.oid
-                    where db.datname = current_database()),
+   IF pg_has_role((SELECT rolname
+                     FROM pg_database db
+               INNER JOIN pg_roles rol
+                       ON db.datdba = rol.oid
+                    WHERE db.datname = current_database()),
```
ehuelsmann, line 1296: By the same reasoning, should we not cache the value of the `current_database` function?

ehuelsmann (changes requested):
Very good you found the other affected locations!

sql/modules/Roles.sql
ylavoie: This code is seldom used and the optimizer seems to behave correctly for its internal STABLE functions. Splitting the code in 2 statements actually makes it longer.

ehuelsmann: Ok. Then this is a non-issue.

ehuelsmann (approved):

sql/modules/admin.sql
ylavoie: Done

sql/modules/Roles.sql
ylavoie: Done

sql/modules/admin.sql
ylavoie: Done

sql/modules/admin.sql
ylavoie: Done

sql/modules/admin.sql
ylavoie: Done

PR #5395, opened 2021-04-08
Yath duration tags

ylavoie:
- Set YATH duration tags in tests to optimize running sequence.
- Allow failed test to be retried 2 times to compensate for random failure of feature tests


ehuelsmann:
Hmm. It looks like running times actually got longer onCCI

ylavoie:
Not delivering expectations.

PR #5396, opened 2021-04-11
Make sure that preferences uses a fresh database

ylavoie:
Test with a fresh database

### DIFF
--- a/xt/66-cucumber/01-basic/preferences.feature
+++ b/xt/66-cucumber/01-basic/preferences.feature
@@ -5,7 +5,7 @@
 Feature: Check correct operation of Preferences screen
 
 Background:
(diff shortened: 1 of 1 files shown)


xt/66-cucumber/01-basic/preferences.feature
```diff
@@ -5,7 +5,7 @@
 Feature: Check correct operation of Preferences screen
 
 Background:
-  Given a standard test company
+  Given a fresh standard test company
```
ehuelsmann, line 8: This means that each iteration in the examples creates its own test database. Is that a good idea? Or should we make sure the settings get reset at the end of each iteration to prevent the need to create new databases?

PR #5400, opened 2021-04-25
Admin destroy

ehuelsmann:
Follow-up to @ylavoie's PR #5384 which has been marked failed for too long now.


lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,177 @@
...
+use Syntax::Keyword::Try;
+
+use LedgerSMB::Admin::Command;
+use LedgerSMB::Database;
+
+use Moose;
+extends 'LedgerSMB::Admin::Command';
+use namespace::autoclean;
+
+my $logger;
```
ehuelsmann, line 22: @ylavoie why not simply declare the logger internal to the `run()` method?

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,177 @@
...
+
+    # Scan databases which could be using the same profile
+    my $sth =
+        $dbh->prepare('SELECT datname FROM pg_database WHERE NOT datistemplate');
+
+    $sth->execute
+        or die $dbh->errstr;
+
+    my @lsmb_databases;
+    while (my $database = $sth->fetchrow) {
```
ehuelsmann, line 59: @ylavoie why first put them in a list and then ...

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,177 @@
...
+        push @lsmb_databases, $database;
+    }
+    $sth->finish;
+    $dbh->disconnect;
+
+    my $role_prefix_usage = 0;
+    for my $database (@lsmb_databases) {
```
ehuelsmann, line 66: ... iterate over *that* list to connect to the databases?

I'm not seeing which other role the list plays in this function, so I don't see its value.

lib/LedgerSMB/Admin/Command/destroy.pm
```diff
@@ -0,0 +1,177 @@
...
+                    connect_data => $connect_database,
+                );
+            my $database_role_prefix = _get_role_prefix($db);
+            if ( defined $database_role_prefix && $role_prefix eq $database_role_prefix ) {
+                $role_prefix_usage++;
+                $logger->debug("'$role_prefix' used by $database");
+            }
+        }
+        catch ($e) {
+            $logger->error("ERROR: $e");
```
ehuelsmann, line 83: We need to prefix the error message with "ERROR:" even when already using the `error` logger???

ehuelsmann (commented):
Some remarks with respect to the code.

lib/LedgerSMB/Admin/Command/destroy.pm
ylavoie: It is used in other functions

lib/LedgerSMB/Admin/Command/destroy.pm
ylavoie: Redundant. Fixed

ehuelsmann:
Closing this one as comments have been merged into #5384

PR #5445, opened 2021-05-23
Ignore CI triggers for changes on the other one

ylavoie:
Make sure that CI will ignore changes done on the other CI configuration

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,14 @@ aliases:
 
 # Reuseable commands
 commands:
+
+  check-changed-files-or-halt:
+    parameters:
+      pattern:
+        type: string
+    steps:
+      - run: git show -m HEAD --name-only --pretty="" | egrep -q -v '<< parameters.pattern >>' || circleci step halt
+
   prove:
     steps:
       - run:
@@ -97,6 +105,9 @@ commands:
     steps:
 
       - checkout
+
+      - check-changed-files-or-halt:
+          pattern: ^.github/*$
 
       - run:
           name: "Monitor CPU & MEMORY"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,9 +8,13 @@ on:
   push:
     branches-ignore:
       - 'master'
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -13,6 +13,14 @@ aliases:
 
 # Reuseable commands
 commands:
+
+  check-changed-files-or-halt:
+    parameters:
+      pattern:
+        type: string
+    steps:
+      - run: git show -m HEAD --name-only --pretty="" | egrep -q -v '<< parameters.pattern >>' || circleci step halt
```
ehuelsmann, line 22: Will this cancel the run if there is no file changed with the content `<< parameters.pattern >>` ? If so, I don't think this is a correct condition: that'll be the case in most situations.

Maybe a better condition is: filter out all files in `.github` and check if there are any files left which have changed?

.github/workflows/main.yml
```diff
@@ -8,9 +8,13 @@ on:
   push:
     branches-ignore:
       - 'master'
+    paths-ignore:
+      - '.circleci'
```
ehuelsmann, line 12: This feels like a much more solid condition indeed!

.circleci/config.yml
ylavoie: - git show list all files involved in the last PR
- grep -q -v set its exit status to success if there are no files matching `<< parameters.pattern >>` and the step will halt

ehuelsmann: Right. And the pattern you're using is `^.github/`; so, if no files match that, the step will halt. But when I only change one of the workflow scripts (e.g. `lib/LedgerSMB/Scripts/setup.pm`) then the build will be cancelled for not matching `.github/`, right?

ylavoie: The build is cancelled only if .github is the only thing in the modified files.

ehuelsmann: ah! sorry about misunderstanding your code! I see I overlooked the `-v` switch's effects.

ehuelsmann (approved):

PR #5469, opened 2021-05-28
Compile README.md to include dynamic elements

ylavoie:
Compile the `README.md` file from a template to allow to:
- dynamically insert the supported browsers used to compile the Javascript UI
- insert the current LedgerSMB version

### DIFF
--- a/doc/sources/_README.md
+++ b/doc/sources/_README.md
@@ -33,7 +33,7 @@ customers or RFQs (request for quotation) to your vendors with PDF attachments.
 
 # System requirements
 
-Note that these are the system requirements for LedgerSMB 1.9.0-dev, the current
+Note that these are the system requirements for LedgerSMB <<LSMB_VERSION>>, the current
 development version. Please check the system requirements for [the 1.7 old stable
 version](https://github.com/ledgersmb/LedgerSMB/tree/1.7#system-requirements)
 and [the 1.8 version](https://github.com/ledgersmb/LedgerSMB/tree/1.8#system-requirements).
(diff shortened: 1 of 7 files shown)


doc/sources/_README.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Client
+
+A `Dojo` compatible web browser is all that's required on the client;
+it includes the recent versions of `Chrome`, `FireFox`, `Opera`,
+`MS Internet Explorer` and a wide range of mobile browsers.
+Please note that LedgerSMB
+explicitly doesn't make any attempts to support Internet Explorer.
+
+The tables below show the browsers currently supported. their earliest date
```
ehuelsmann, line 59: small typo: `show` -> `shows`

ehuelsmann: oh, and I think you'll want to change the dot to a comma.

doc/sources/_README.md
```diff
@@ -0,0 +1,269 @@
...
+and a range of versions.
+
+<<browsers_list>>
+
```
ehuelsmann, line 63: At the bottom of the list, I'd add a remark that the data comes from "caniuse.com" and that it doesn't have information on older mobile browsers except for Safari, because browser updates are pushed out to all devices when they come out, meaning that the number of old browsers in negligible.

ehuelsmann:
Do we need "all maintained Node versions" as part of our supported browser list? Because if we don't, I think we won't get these "not supported on Node.JS" warnings. And maybe we really don't: we don't run this code on the server, after all...

ehuelsmann (changes requested):
Please check out the typos I found.

webpack.config.js
```diff
@@ -24,6 +25,7 @@ const VirtualModulePlugin = require("virtual-module-webpack-plugin");
 const { CleanWebpackPlugin } = require("clean-webpack-plugin"); // installed via npm
 
 const argv = require("yargs").argv;
+const { nextTick } = require("process");
```
ehuelsmann, line 28: LGTM says "not used".

doc/sources/_README.md
ylavoie: Period changed. No typo, tables are plural.

ehuelsmann: duh! sorry about that.

doc/sources/_README.md
ylavoie: Done

ylavoie:
> Do we need "all maintained Node versions" as part of our supported browser list? Because if we don't, I think we won't get these "not supported on Node.JS" warnings. And maybe we really don't: we don't run this code on the server, after all...

Node versions removed. Warnings will stay because of Safri 6, imposed by PhantomJS 2.1

ehuelsmann:
Right. We need to think when we want to stop using PhantomJS as a testing platform.

ehuelsmann (approved):

PR #5493, opened 2021-05-30
Mouse-over highlighting for reconciliation screen

ehuelsmann:


ylavoie (changes requested):
Styles have to be defined before selectors can be added in order for tem to be effective

ylavoie (approved):

PR #5516, opened 2021-06-05
Improve FilteringSelect by turning off auto-completion

aungzwin10:
The dijit/form/FilteringSelect autoComplete property is set to true by default. So whenever user type in the textbox, automatically override  user current text with the first option of the filtering result.
When it sets to false,  it will no longer override user current text.


aungzwin10:
lsmb/accounts/AccountSelector also set autoComplete to false. Please see below.

https://github.com/ledgersmb/LedgerSMB/blob/25d369d58ffab314869c5a388c6dcbfad3af4339/UI/js-src/lsmb/accounts/AccountSelector.js#L21

PR #5525, opened 2021-06-07
Fix trial balance report

ylavoie:
Completes #5373 in Trial Balance

### DIFF
--- a/sql/modules/trial_balance.sql
+++ b/sql/modules/trial_balance.sql
@@ -169,9 +169,8 @@ BEGIN
         group by account_id) cp ON cp.account_id = a.id
     LEFT JOIN (SELECT trans_id, description
                  FROM account_translation at
-              INNER JOIN user_preference up ON up.language = at.language_code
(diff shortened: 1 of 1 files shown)


sql/modules/trial_balance.sql
```diff
@@ -169,9 +169,8 @@ BEGIN
         group by account_id) cp ON cp.account_id = a.id
     LEFT JOIN (SELECT trans_id, description
                  FROM account_translation at
-              INNER JOIN user_preference up ON up.language = at.language_code
-              INNER JOIN users ON up.id = users.id
-                WHERE users.username = SESSION_USER) at ON a.id = at.trans_id
+               WHERE language_code = preference__get('language')) at
```
ehuelsmann, line 172: two inner joins replaced by a single (constant-in-the-scope-of-the-query) function call. Nice!

ylavoie: I copied your code. ;-)

PR #5587, opened 2021-06-18
Mark all lines in a transaction 'approved' upon approval

### LINKED ISSUE #5582: 1.9: AP/AP transaction lines are not saved as approved in Post
AR/AP transaction lines are not saved as approved=true when post.

in ar/ap table, the approved status is true when post. But in acc_trans, the approved status is false.

ehuelsmann:
Fixes #5582: Since 1.9, the default value on rows is to be posted "not
approved"; it's important to mark rows approved on transaction approval now.

### DIFF
--- a/sql/modules/Drafts.sql
+++ b/sql/modules/Drafts.sql
@@ -109,6 +109,10 @@ begin
                 approved_at = now()
         WHERE id = in_id;
 
+        UPDATE acc_trans
+        SET approved = 't'::boolean
+        WHERE trans_id = in_id;
+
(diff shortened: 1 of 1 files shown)


aungzwin10:
The 'Post' button now saved the transaction with approved-true, right?

ehuelsmann:
@aungzwin10 they're saved "not approved" (before 1.9 they were often saved "approved"); however, when they're consistently not approved when saved, then it's important to mark them "approved" when approving the transaction itself.

PR #5589, opened 2021-06-18
Fix #5583: UTF-8 characters incorrectly encoded on journal entry screen

### LINKED ISSUE #5583: Unicode character are not correctly showed in Journal Entry
In other screen such as AR/AP transaction and Payment, unicode character are correctly saved and rendered

See below for journal entry
Screen Shot 2021-06-18 at 4 15 43 PM

ehuelsmann:
When clicking 'Update' on a journal entry screen which contains non-ascii
characters, these characters get mangled.


ehuelsmann:
@ylavoie why was this check cancelled? Is it because I entered another PR on a different branch only minutes later?

PR #5598, opened 2021-06-19
Exclude unreleased versions from the table of supported browsers

ehuelsmann:


README.md
```diff
@@ -64,36 +64,35 @@ and a range of versions.
...
 are pushed out to the devices as soon as they are releases, so the amount of
 old browsers is negligible. **
-
 # Quick start (`Docker compose`)
 
 The quickest way to get the `Docker` image up and running is by using the
 docker-compose file available through the `GitHub` repository at:
 
-https://github.com/ledgersmb/ledgersmb-docker/blob/1.9/docker-compose.yml
+https://github.com/ledgersmb/ledgersmb-docker/blob/1.8/docker-compose.yml
```
ylavoie, line 95: This is brought back by doc/sources/_README.md, which is the template from which 'webpack-compile-markdown` generates the actual `README.md`

ehuelsmann: I'm wondering: as long as 1.9 hasn't been released, is 1.8 not the correct value to report?

PR #5599, opened 2021-06-20
Clean up webpack

ylavoie:
CLeanup Dojo:
- Remove remnants of old dojo builder
- Remove unused innerStyle variable in parts templates
- Remove ES6 polyfils because we now require browsers to be ES6 compliants
- Fix webpack analyzer tool to help optimizing


ehuelsmann:
@ylavoie quick question: how did you check the innerStyle isn't being used? (Also, it's on the parts*selector* ; which is not used in the parts screens, but on the invoice and order screens; I was thrown off by your wording about the "parts template")

ylavoie:
We always empty `innerStyle` to void a pre-existing condition which is not assigned  in any of our code nor in dojo, dijit or any other node_modules.

ehuelsmann:
ok. thanks for that explanation!

ehuelsmann (approved):
Yup. Looking good. Nice to be able to drop polyfills!

PR #5618, opened 2021-06-23
Fix node_modules caching

ylavoie:
Also consider HTML files in deciding if Dojo module cache validity

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,15 +50,16 @@ jobs:
           key: |
             modules-${{ hashFiles('**/package-lock.json') }}
 
+      - name: Install dependencies
+        if: steps.cache.outputs.cache-node_modules != 'true'
+        run: npm ci --ignore-scripts
+
       - name: Cache Dojo
         id: cache-dojo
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -50,15 +50,16 @@ jobs:
...
       - name: Cache Dojo
         id: cache-dojo
         uses: actions/cache@v2
         with:
           path: UI/js
-          key: dojo-${{ hashFiles('UI/{js-src,css}/**') }}
-
-      - name: Install Node modules Dependencies
-        run: npm ci
+          key: dojo-${{ hashFiles('UI/{js-src,css}/**') }}-${{ hashFiles('UI/**/*.html') }}
```
ehuelsmann, line 62: You can use multiple patterns with a single hashfiles invocation too: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-with-multiple-patterns

ylavoie: Thanks for the heads-up, that was forbibben before

PR #5627, opened 2021-06-26
Payment fixes

ehuelsmann:


lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1317,6 +1317,10 @@ sub post_payment {
     @array_options = $Payment->get_entity_credit_account();
     my $discount_account_id = $array_options[0]->{discount};
     @array_options = $Payment->get_open_invoices();
+
+    die q|Applicable discounts can't be posted due missing discount account in credit account|
+        if any { $_->{discount_tc} != 0 } @array_options;
```
aungzwin10, line 1322: How that check know discount account is not set in the vendor/customer credit account? I think the discount account is taken at line 1318 and @array_options is replaced by another set of value?

ehuelsmann: The check was changed to https://github.com/ledgersmb/LedgerSMB/pull/5627/files#diff-07704721ae554fe0bb4ef1b2deac3c01d932127e31f54cb25b4671a2c5715362R1322-R1323

PR #5630, opened 2021-06-27
Perl 5.34

ylavoie:
Move Coverage on both CIs to Perl 5.34 to quiet Devel::Cover logs.


lib/LedgerSMB/PSGI.pm
```diff
@@ -107,9 +107,6 @@ sub old_app {
                 catch ($e) {
                     exit;
                 }
-                finally {
-                    exit;
-                }
```
ehuelsmann, line 109: actually, that's not equivalent.

```plain
try {
...;
}
finally {
   exit;
}
```

can be replaced with:

```plain
try {
   ...;
}
catch ($e) {
}
exit;
```

because finally happens in both the error and the non-error cases.

ehuelsmann (changes requested):

lib/LedgerSMB/PSGI.pm
ylavoie: Fixed

ehuelsmann (approved):

PR #5647, opened 2021-06-29
Fix 5646, add blank memo if the discount applied

### LINKED ISSUE #5646: Memo are not posted correctly in multiple invoice payment
According to these [lines](https://github.com/ledgersmb/LedgerSMB/blob/55cac878c22555298aa8b8fadb5becef940694d3/lib/LedgerSMB/Scripts/payment.pm#L1294-L1305), each column are stored as array when posting payments.

When the discount applied on the invoice, the length of `@memo` array and other array are not the same.

Screen Shot 2021-06-29 at 10 38 24 AM

According to my above example screenshot, the `@amount` and `@transaction_id` arrays have length of 6(1 cash line, 1 discount line) since there is 3 invoices. But for the `@memo` array, the length is 3. So, while looping all the arrays using the same index, the memo are posted to wrong invoice.

aungzwin10:
Closes #5646

### DIFF
--- a/lib/LedgerSMB/Scripts/payment.pm
+++ b/lib/LedgerSMB/Scripts/payment.pm
@@ -1360,7 +1360,7 @@ sub post_payment {
                 push @amount, $temporary_discount;
                 push @cash_account_id, $discount_account_id;
                 push @source, $locale->text('Applied discount');
-                push @memo, "";
+                push @memo, undef;
                 push @transaction_id, $array_options[$ref]->{invoice_id};
             }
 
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1360,6 +1360,7 @@ sub post_payment {
                 push @amount, $temporary_discount;
                 push @cash_account_id, $discount_account_id;
                 push @source, $locale->text('Applied discount');
+                push @memo, "";
```
ehuelsmann, line 1363: Could you push `undef` instead of an empty string? That way, in the database a NULL value gets stored instead of an empty string.

For future reference: our tests verify that strings without "variables inside" should be using single quotes. E.g. "string with $variable interpolation" vs 'string constant'. Thus the above should have read `''`.

aungzwin10: the updated commit change to undef

ehuelsmann (approved):

PR #5671, opened 2021-07-05
Fix 5656 and 5657

### LINKED ISSUE #5656: Payable Total is not correctly calculated for foreign currency when number format applied
Number format is "1,000.00"

Screen Shot 2021-06-30 at 4 32 02 PM

### LINKED ISSUE #5657: System default decimal places are not applied for AR/AP Transaction
Although User set decimal places for money in System > Default, AP/AR transaction are not applied with the setting and stick it to 2 decimal places.

Screen Shot 2021-06-30 at 4 46 10 PM

aungzwin10:
The default currency amount for payable should be calculated without format.
The decimal places should always stick to the user defined value and never used constant value in code unless it is intentional.

Closes #5656 
Closes #5657

### DIFF
--- a/old/bin/aa.pl
+++ b/old/bin/aa.pl
@@ -690,7 +690,7 @@ $form->open_status_div($status_div_id) . qq|
 
         # format amounts
         $form->{"amount_$i"} =
-          $form->format_amount( \%myconfig,$form->{"amount_$i"}, 2 );
+          $form->format_amount( \%myconfig,$form->{"amount_$i"}, LedgerSMB::Setting->new(%$form)->get('decimal_places') );
 
         $project = qq|
       <td align=right><select data-dojo-type="dijit/form/Select" id="projectnumber_$i" name="projectnumber_$i">$form->{"selectprojectnumber_$i"}</select></td>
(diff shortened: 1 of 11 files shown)


old/bin/aa.pl
```diff
@@ -792,8 +792,8 @@ sub form_header {
       <th align=left>$form->{invtotal}</th>
      <td>| . (($form->{currency} ne $form->{defaultcurrency})
               ? $form->format_amount(\%myconfig,
-                                     $form->{invtotal}
-                                     * $form->{exchangerate}, 2) : '') . qq|</td>
+                                     $form->parse_amount( \%myconfig, $form->{invtotal} )
```
ehuelsmann, line 795: Only a few lines up, `invtotal` gets formatted. Maybe we can save the unformatted value for this calculation instead of parsing it?

old/bin/aa.pl
```diff
@@ -792,8 +792,8 @@ sub form_header {
...
+                                     * $form->parse_amount( \%myconfig, $form->{exchangerate} ), LedgerSMB::Setting->new(%$form)->get('decimal_places')) : '') . qq|</td>
```
ehuelsmann, line 796: I'm not sure we should want to do more than what you're doing (that is: parsing the exchange rate), but the same function first formats the exchangerate here: https://github.com/ledgersmb/LedgerSMB/blob/3008375e4e764ff2acc7c2a2b564fa41e6844a05/old/bin/aa.pl#L425

ehuelsmann:
@ylavoie, could you check the two suggestions I'm making? Am I going overboard?

old/bin/aa.pl
ylavoie: Assigning a formatted_amount to the `$form->{exchangerate}` will truncate precision for a value which is used in 10 other modules

old/bin/aa.pl
ylavoie: We should indeed use formatted amount for output only and parse_amount for input only

PR #5752, opened 2021-08-10
Remove undefined interval warning in PNL reports

ylavoie:
Make sure that PNL header is properly initialized


ehuelsmann (approved):
Totally makes sense.

PR #5759, opened 2021-08-10
Use `make devtest` on CI to deduplicate test infrastructure

ehuelsmann:
Use `make devtest` everywhere, so it will work on local test infrastructure when I need it...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -221,7 +221,7 @@ jobs:
 
       - name: Run Perl tests
         run: |
-          make devtest TESTS='--no-progress t xt'
+          make devtest TESTS='--no-progress --job-count 2 t xt'
         env:
           LSMB_TEST_DB: 1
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -6,13 +6,11 @@ name: CI
 on:
   # Triggers the workflow on push or pull request events but only for the master branch
   push:
-    paths-ignore:
-      - '.circleci'
+    branches:
+      - '*'
```
ylavoie, line 10: This conflicts with #5751 and will bring back test runs on push & PR for `renovate`

ylavoie (dismissed):

.github/workflows/main.yml
ehuelsmann: This was never intended for merging: the branch didn't want to build, I just went into trial-and-error, but GitHub Actions turned out to be down.

PR #5761, opened 2021-08-11
Fix LedgerSMB version [skip ci]

ylavoie:
Use 1.10.0-dev in package.json too


ehuelsmann:
@ylavoie, I think this needs to be added to [the checklist on the wiki](https://github.com/ledgersmb/LedgerSMB/wiki/Release-check-list) then, to prevent this from happening again?

PR #5764, opened 2021-08-11
 Use `make devtest` on CI to deduplicate test infrastructure (1.8 branch)

ehuelsmann:


ehuelsmann:
@ylavoie when running this branch locally, I get an error `[08:32:52] xt/65-browser.t ................................................ Selenium server did not return proper status at /usr/local/share/perl/5.32.1/Selenium/Remote/Driver.pm line 552.` after which all BDD/browser tests fail.

Do you have any experience with this error?

ehuelsmann:
@ylavoie never mind; found the cause: the port number on which the selenium server runs, has changed in the docker image, but not in the 1.8 branch configuration.

ehuelsmann:
@ylavoie this PR shows another problem though: because it modifies `.circleci/`, it doesn't get run on GitHub Actions. I would expect commits that *only* modify `.circleci` to be skipped, but not commits that *include* `.circleci/` mods (but also others).

PR #5775, opened 2021-08-14
Enable renovate on master and 1.9 [skip ci]

ylavoie:
Enable renovate processing of master and 1.9.
Contrary to expected usage, renovate only reads config from the default branch, `master`, in our case. It wouldn't process a config in 1.9

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,6 +2,7 @@
     "extends": [
       "config:base"
     ],
+    "baseBranches": [ "master", "1.9" ],
     "packageRules": [
         {
             "matchPackagePatterns": [


renovate.json
```diff
@@ -2,6 +2,7 @@
     "extends": [
       "config:base"
     ],
+    "baseBranches": [ "master", "1.9" ],
```
ehuelsmann, line 5: @ylavoie thanks! We probably need to add this location to the release wiki page as I to add a new branch for a minor/major release, as well as to the list of locations to update when a release goes EOL.

ylavoie: Wiki updated

PR #5914, opened 2021-08-28
Stabilize BDD inventory adjustment test

ehuelsmann:
Too often, the test fails due to timing issues.


ehuelsmann:
Apparently we have more/additional weaknesses. Those are for later investigation.

PR #5972, opened 2021-09-12
Reduce per-request number of executed queries

ehuelsmann:
Reduce the number of queries executed to "start up the
request environment" by 20 (21 -> 1).


ehuelsmann:
@ylavoie this should resolve some of your annoyance about the number of queries run per request. It also seems to be a good way to speed up request handling.

PR #5973, opened 2021-09-12
Complete introducing Log::Any

ehuelsmann:
This commit reduces the number of errors thrown by the test suite
about Log4perl not being initialized.


ehuelsmann:
This commit actually does *not* reduce the number of "Log4perl not initialized" warnings, because those are generated by Workflow, as it turns out...

PR #5982, opened 2021-09-14
Prevent approving draft transaction by user who don't have permission

### LINKED ISSUE #5957: User role and permission are not working correctly
Although I have uncheck the `draft modify` and `draft post` at the account permission, I am able to approve at the Transaction Approval > Drafts. And I am able to edit/modify the transaction. Please see screenshot.

After I change the permission, I log out and log in. Also test with different browser.

Screen Shot 2021-09-08 at 12 15 20 PM

IMO, `Approve` button should disappear or disabled or check at backend when submit.
Screen Shot 2021-09-08 at 12 25 04 PM

IMO, `Save Draft` and `Post` buttons should disappear or disabled or check at backend when submit.
Screen Shot 2021-09-08 at 12 29 08 PM

aungzwin10:
Closes #5957

### DIFF
--- a/old/bin/aa.pl
+++ b/old/bin/aa.pl
@@ -949,7 +949,8 @@ sub form_footer {
             $hold_text = $locale->text('On Hold');
         }
 
-
+# REMARK: According to the check below, post_as_new and delete button never render
+# Why post_as_new and delete are exist?
         %button = (
             'update' => { ndx => 1, key => 'U', value => $locale->text('Update') },
             'copy_to_new' => { ndx => 2, key => 'C', value => $locale->text('Copy to New') },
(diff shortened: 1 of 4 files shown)


old/bin/aa.pl
```diff
@@ -949,7 +949,8 @@ sub form_footer {
             $hold_text = $locale->text('On Hold');
         }
 
-
+# REMARK: According to the check below, post_as_new and delete button never render
+# Why post_as_new and delete are exist?
```
ehuelsmann, line 953: Good point (asking why `delete` and `post_as_new` exist when they can't be triggered). However, I think that belongs either in a discussion item in GitHub, or in an issue, or even in the text that goes with the pull request. The *answer* to the question however, *does* belong in the code...

The answer to the question is as simple as disappointing: this code has had too many iterations by people who did not fully comprehend their changes, but did know just enough to be able to test that their edits worked (for their use-case). Transactions used to be deletable, but then "draft" transactions came and not all transactions could be deleted anymore. This required the Delete button to be (temporarily?) disabled. I think the same applies to "Post as New".

PR #6057, opened 2021-09-29
US Hierarchical COA

neilt:
This pull request adds 2 files

The first is `validate_coa_xml.sh`, which when executed will recursively scan all files in `local/coa`, check the xml against the psd, and check the format matches expectation.  If either fails it will stop, return non-zero error, and show a diff of the lines that do not match.  More instructions are in the file itself, but there is no help.

The second is `GeneralHierarchical.xml` which add a new hierarchical coa for US.  I've not verified that this file works correctly other than loading and appearing correct in the chart of accounts listing.

I have an issue in that `GeneralHierarchical.xml` does not correctly set the `earn_id`.  The xml file specifies 3999, but when loaded it defaults to 1000.  I believe this to be a bug in the default handling of `earn_id`.


ehuelsmann:
Thanks for this PR! Over all looking very well.

> I have an issue in that GeneralHierarchical.xml does not correctly set the earn_id. The xml file specifies 3999, but when loaded it defaults to 1000. I believe this to be a bug in the default handling of `earn_id`

You may actually have identified an issue in the determination of the "earn_id". Does it get set correctly in the "defaults" table in the database? (That's where all these settings are stored.) The number in the "value" column associated with the `earn_id` "setting_key" column, is a reference to an ID in the `account` or `account_heading` table.


One request: to make sure all changes we merge are also part of all future releases, we prefer to PR against "master" and then backport to release branches (in this case only 1.9 is applicable for backport). I'll merge this PR, but could you either *also* PR against "master* or *just* PR against master?

aungzwin10:
Please allow me to add my previous finding about this issue here.
```
<settings>
     <setting name="inventory_accno_id" accno="1510"/>
     <setting name="income_accno_id" accno="4010"/>
     <setting name="earn_id" accno="3999"/>
     <setting name="expense_accno_id" accno="5010"/>
     <setting name="fxgain_accno_id" accno="4450"/>
     <setting name="fxloss_accno_id" accno="5810"/>
     <setting name="weightunit" value="lbs"/>
 </settings>
```
Among the above settings, `earn_id` is the one that accept the Account Heading number not Account number. So `accno` attribute cannot apply to `earn_id` because `accno` only select the account from account table not from account_heading.(Correct me if I'm wrong)
@ehuelsmann also note about this in the code, see line 416 below.
https://github.com/ledgersmb/LedgerSMB/blob/e20fd9e5717600fb89ad1e4afb59bacf8fcecbcb/lib/LedgerSMB/Company/Configuration.pm#L401-L430

According to the above code, another way to set `earn_id` is using the `value` attribute.@ehuelsmann use `value` and sql statement to select the heading id. **Note:** It should be account_heading. See below
https://github.com/ledgersmb/LedgerSMB/blob/e20fd9e5717600fb89ad1e4afb59bacf8fcecbcb/locale/coa/nz/General.xml#L185
So @neilt need to use something like this 
`<setting name="earn_id" value="(select id from account_heading where accno = '3999')"/>`

But the problem didn't solve. The statement is not treated as SQL and it stored as a text in the db because `value` column in `defaults` table is type of `text`. https://github.com/ledgersmb/LedgerSMB/blob/e20fd9e5717600fb89ad1e4afb59bacf8fcecbcb/lib/LedgerSMB/Company/Configuration.pm#L512-L518
See the screenshot.
Screen Shot 2021-09-30 at 9 29 38 AM



There is another issue, for account_heading, parent heading is not set at all.
Example, using @neilt coa,
```
<account-heading id="h-1" code="1000" description="ALL ASSETS">
       <account-heading id="h-1a" code="1050" description="CURRENT ASSETS">
         <account code="1060" description="Checking Account" category="Asset" recon="true">
```
Parent of `CURRENT ASSETS` is `ALL ASSETS`, right? But it is not set. @neilt Please check your `CURRENT ASSETS` whether or not heading dropdown is correctly selected to `ALL ASSETS`.

neilt:
@aungzwin10 
> Parent of CURRENT ASSETS is ALL ASSETS, right? But it is not set. @neilt Please check your CURRENT ASSETS whether or not heading dropdown is correctly selected to ALL ASSETS.

After loading the xml file none of the detail asset headers `CURRENT ASSETS`, `INVENTORY ASSETS`, or `CAPITAL ASSETS` are set with the parent as `ALL ASSETS`.  This appears to be a bug. 

Also the same problem exists for `ALL LIABILITIES`. None of the sub-accounts are loaded with the heading set.

I am going to file this as a bug.

PR #6091, opened 2021-10-03
Fix undefined defaultcurrency in recurring transactions

ylavoie:
Cmplete #5975 and use the default currency from the cache


old/lib/LedgerSMB/AM.pm
```diff
@@ -680,7 +680,7 @@ sub recurring_transactions {
                   s.*, se.formname AS recurringemail,
                   sp.formname AS recurringprint,
                   s.nextdate - current_date AS overdue, '' AS vc,
-                  $defaultcurrency AS curr,
+                  '$form->{defaultcurrency}' AS curr,
```
ehuelsmann, line 683: I would feel more comfortable directly inserting values into queries if the variable content has been run through `$dbh->quote()`. That eliminates SQL injection risks.

PR #6096, opened 2021-10-05
Fix US General COA

neilt:
Add the 'Current earnings' default setting.
Add the correct link code so that 4010 Sales can be selected as the default 'Income' account.

Before:
Screen Shot 2021-10-05 at 9 08 57 AM

After:
Screen Shot 2021-10-05 at 10 07 14 AM

### DIFF
--- a/locale/coa/us/General.xml
+++ b/locale/coa/us/General.xml
@@ -73,6 +73,7 @@
       <account code="4010" description="Sales" category="Income">
         <link code="AR_amount"/>
         <link code="IC_sale"/>
+        <link code="IC_income"/>
       </account>
     </account-heading>
     <account-heading id="h-9" code="4400" description="OTHER REVENUE">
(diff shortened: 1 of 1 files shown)


locale/coa/us/General.xml
```diff
@@ -159,6 +160,7 @@
   </currencies>
   <settings>
     <setting name="inventory_accno_id" accno="1510"/>
+    <setting name="earn_id" accno="3500"/>
```
ehuelsmann, line 163: This is the reason the tests fail: setting the `earn_id` indicates to the system that it has been set up for hierarchy reporting and changes the default setup of the balance sheet and PNL from "account category basis" to "hierarchy". The two result in different roll-up of PNL results into "current earnings".

neilt: Ok, so I'll remove earn_id, but it still does not seem right that Income goes to Freight. So Should I make a new pull request with just that?

ehuelsmann: Removing this setting from the PR (we can always squash the commits) and then merging the PR does make sense, I'd say, yes.

Issue #2512, opened 2017-01-07
Without overpayment accounts, don't show overpayments section in payments


ehuelsmann:
Most default charts of accounts don't have overpayment accounts configured. Which makes the lower half of the single payments screen unusable (and incomprehensible) and thereby the screen less than usable.

Maybe the solution here could be to keep the table in place, but replace its content with the notification that there's no OP account configured?

PR #6097, opened 2021-10-05
Don't show the overpayment section without overpayment accounts configured

### LINKED ISSUE #6059: Payment form does not allow expense account entry
Step to reproduce:

1. Create a new vendor in Contacts -> Add Contact.  Verify that class is marked 'Vendor'
2. Go to 'Credit Accounts' tab.
3. Fill in 'Pay To', 'Description', 'Payment Terms', and 'Starting Date'.
4. Save
5. Go to Cash -> Payment.
6. Enter vendor name from above.
7. Click 'Continue'.
8. In 'Item 1' the 'Account' column has a pull down, but does not have any entries.

I would have expected a list of G/L expense accounts.

Branch: 1.9
Commit: c7e99971412ec470b247d5e5d0f71665a8bb41ac

The problem is present with both:
G/L: US GeneralHierarchical.xml
G/L: US General.xml

ehuelsmann:
As per the comments on the chat and in the issue, the current input screen
gives the wrong impression of what the screen actually means. The first step
to improvement is to remove the part that is irrelevant due to missing
configuration.

Closes #6059

### DIFF
--- a/UI/payments/payment2.html
+++ b/UI/payments/payment2.html
@@ -334,6 +334,8 @@ onLoad="maximize_minimize_on_load(event, 'div_topay_state', 'payments/img/down.g
         <th colspan="2">[% topay_subtotal -%]&nbsp;[% curr.value -%]</th>
       </tr>
     </table>
+
+[% IF overpayment_account.size > 0 %]
     <table width="100%" id="overpayments">
       <tr>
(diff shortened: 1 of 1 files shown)


ehuelsmann:
Closes #2512 too

PR #6119, opened 2021-10-10
Disable experimentalUseImportModule option for MiniCssExtractPlugin

dram:
Newer version of mini-css-extract-plugin enables `experimentalUseImportModule` by default, which will be conflict with dojo-webpack-plugin.

Trying to disable this option explicitly, so behavior is consistent with older versions.

This PR will fix the upgrade problem in #6105.


ehuelsmann:
@dram thank you for your contribution!

PR #6157, opened 2021-10-22
Ensure the order parameters end up in the URL's query string

### LINKED ISSUE #6151: Cannot sort asset class list table
Fixed Assets > Asset Classes > List Classes > Search(without filtering). Click column header to sort. 

Screen Shot 2021-10-22 at 9 28 25 AM

Screen Shot 2021-10-22 at 9 32 15 AM

ehuelsmann:
When the URL doesn't have a query string already, add and empty one
at the end.

Fixes #6151

### DIFF
--- a/UI/asset/begin_approval.html
+++ b/UI/asset/begin_approval.html
@@ -1,7 +1,7 @@
 \[% PROCESS "elements.html" -%]
 <body class="lsmb [% dojo_theme %]">
 <div id="assets">
-<form data-dojo-type="lsmb/Form" action="[% request.script %]" method="post">
+<form data-dojo-type="lsmb/Form" action="[% request.script %]" method="GET">
 <div class="listtop">[% text('New Asset Report Search') %]</div>
 <div class="inputrow" id="classrow"><div class="inputgroup" id="classgrp">
 [% PROCESS select element_data = {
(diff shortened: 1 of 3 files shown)


aungzwin10:
@ehuelsmann Any update regarding this? Please feel free to ask if you need more information.

ehuelsmann:
@aungzwin10 I'm not sure I can fix this problem in the current code base: it seems that the query string is missing in the original request, which makes the code fail to return the script plus the query string; this would normally be caused by the fact that the request is a POST request -- i.e. a request without a query string.

aungzwin10:
@ehuelsmann Thanks for explaining details, I think the request should be changed to GET since we fetch data.

ehuelsmann:
> @ehuelsmann Thanks for explaining details, I think the request should be changed to GET since we fetch data.

Agreed. Do you have a point where you think this needs to be changed?

aungzwin10:
@ehuelsmann 
Yes. If it is because of POST method, we can change the form method to GET in this line.
https://github.com/ledgersmb/LedgerSMB/blob/e6bec51b6d854af7c25ec843b910a5b1344015c0/UI/asset/search_class.html#L5

I tested without current PR changes, it works well.

ehuelsmann:
@aungzwin10 this new PR should fix the sorting/ordering problems you've reported.

aungzwin10:
Yes, it fix ordering issue. Thank you

ehuelsmann:
Thanks for confirming!

Issue #816, opened 2015-09-11
Installation in a schema other than 'public'


ehuelsmann:
Current experience is described by Nicolas Couchoud [his mail to the users list](http://archive.ledgersmb.org/ledger-smb-users/msg08579.html).

However, the fact that we support a configuration parameter called 'schema' in the config file suggests we actually _do_ support an out-of-'public'-installation.

PR #6186, opened 2021-11-08
Support running LedgerSMB in a schema other than "public"

ehuelsmann:
And prove it by running all tests against schema "xyz".


sql/changes/1.8/reconciliation-links-cascade.sql@1
```diff
@@ -0,0 +1,10 @@
+
+
+alter table cr_report_line_links
+   drop constraint cr_report_line_links_report_line_id_fkey,
+   add CONSTRAINT cr_report_line_links_report_line_id_fkey
+        FOREIGN KEY (report_line_id)
+           REFERENCES public.cr_report_line (id) MATCH SIMPLE
```
ylavoie, line 7: We're adding back a fixed reference to public?

ylavoie (changes requested):

sql/changes/1.8/reconciliation-links-cascade.sql@1
ehuelsmann: This is a copy of the original file for the purpose of the SHA calculation: that is used to prevent it from being applied again. The `@1` part means that it's an older version of the file that should be applied if none of the SHA sums match.

The file without the public prefix in it, is the one that's going to be run on the database going forward.

ylavoie (commented):
Nice work!
Except for documentation, we still have 12 modules refering to `public`:
- lib\LedgerSMB\Sysconfig.pm
- sql\modules\admin.sql
- sql\modules\Assets.sql
- sql\modules\Budgetting.sql
- sql\modules\Company.sql
- sql\modules\Employee.sql
- sql\modules\Journal.sql
- sql\modules\Menu.sql
- sql\modules\Reconciliation.sql
- sql\modules\Session.sql
- sql\modules\Voucher.sql
- sql\upgrade\sl3.0.sql
- t\16-dbupgrade.t

ehuelsmann:
>     * lib\LedgerSMB\Sysconfig.pm
> 
>     * sql\modules\admin.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Assets.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Budgetting.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Company.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Employee.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Journal.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Menu.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Reconciliation.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\modules\Session.sql
Only contains the word "public".

> 
>     * sql\modules\Voucher.sql
This one refers to the PUBLIC role, not the schema.

> 
>     * sql\upgrade\sl3.0.sql
Good point. We'll need to have a look at how to deal with this one.


> 
>     * t\16-dbupgrade.t
"public" is the standard schema name of "LedgerSMB::Database". This file references "public", the default value of the "LedgerSMB::Database->schema" attribute. It doesn't have to do with the actual schema in a the database.

ylavoie (approved):

ehuelsmann:
Closes #816

PR #6215, opened 2021-11-22
Reconciliation fix 1.9

aungzwin10:
Summarizing:

* Use named arguments instead of positional
Clean up for better readability

* Fix missing FROM clause in reconciliation
Fix using our_value as column name, our_value doesn't exist in cr_report_line

* Fix payment source doesn't correctly save to scn of cr_report_line

* Fix description always empty in reconciliation
entry_id from acc_trans is not saved in the cr_report_line table. It is saved in cr_report_line_links.

* Fix ID duplication on the payee view because of the cr_report_line_links

* Fix missing FROM caluse in sl3.0

* Undo change adding payments with the wrong granularity to the recon

And add a lot of documentation to (hopefully) explain why this code
looks the way it does.

* Document reconciliation test cases and re-order for clarity

Group the journal lines and add comments for clarity on the actual tests
that are intended to be executed.

* Add and adapt tests

Add test to check adjustment GL lines being added to payments given the same source and date
Adjust tests to incorporate Source as a splitting criterion *within* a payment
Adjust number of lines in test against account -11112 better distinguish test cases

Co-authored-by: Erik Huelsmann <ehuels@gmail.com>

### DIFF
--- a/sql/modules/Reconciliation.sql
+++ b/sql/modules/Reconciliation.sql
@@ -232,11 +232,15 @@ $$Marks the report approved and marks all cleared transactions in it cleared.$$;
 
 
 -- XXX Badly named, rename for 1.4.  --CT
-CREATE OR REPLACE FUNCTION reconciliation__new_report_id
-(in_chart_id int, in_total numeric, in_end_date date, in_recon_fx bool) returns INT as $$
+CREATE OR REPLACE FUNCTION reconciliation__new_report_id(
+    in_chart_id int,
+    in_total numeric,
+    in_end_date date,
+    in_recon_fx bool
+) returns INT as $$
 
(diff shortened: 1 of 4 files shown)


aungzwin10:
@ehuelsmann, Do you want to group by `entry_id` when source is null?
https://github.com/ledgersmb/LedgerSMB/blob/105ea00fa7e0bfcc48d896c7187270c7607c44a5/sql/modules/Reconciliation.sql#L553-L562

Sorry, I just notice the changes.

ehuelsmann:
Yes, that change is correct: i removed the column as it was unnamed and unused, but the aggregation is correct. When there is a `NULL` Source, we want the individual rows, not all the NULLs aggregated into one row.

ehuelsmann (approved):

PR #6230, opened 2021-11-26
Washington State tax webservice module

ehuelsmann:
Sponsored by @freelock

### DIFF
--- a/old/bin/aa.pl
+++ b/old/bin/aa.pl
@@ -508,7 +508,8 @@ $form->open_status_div($status_div_id) . qq|
     $form->hide_form(
         qw(batch_id approved id printed emailed sort
            oldtransdate audittrail recurring checktax reverse batch_id subtype
-           entity_control_code tax_id meta_number default_reportable address city)
+           entity_control_code tax_id meta_number default_reportable
+           address city zipcode state country)
     );
 
     if ( $form->{vc} eq 'customer' ) {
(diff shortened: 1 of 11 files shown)


old/bin/aa.pl
```diff
@@ -508,7 +508,8 @@ sub form_header {
     $form->hide_form(
         qw(batch_id approved id printed emailed sort
            oldtransdate audittrail recurring checktax reverse batch_id subtype
-           entity_control_code tax_id meta_number default_reportable address city)
+           entity_control_code tax_id meta_number default_reportable
+           address city zipcode state country)
```
ylavoie, line 512: We're tying AA to USA based address, which may not be appropriate for other countries.

ylavoie: Same for other old code modules

ehuelsmann: Are you referring to the names of the variables? `zipcode` and `state` ? You're aware that I'm not creating these values here, but just persisting them across page reloads?

ylavoie: I was refering to the variable names, yes. They are only names but the need to add them show increasing ties with US/CA/UK and similar.

Canada has counties, which are never used is postal purposes. Same for USA. Templating address would surely be a good idea, fields vary per country. So does their order of presentation.

Address localization will have to be addresed somewhen but you're right, it's out of scope for this issue.

PR #6234, opened 2021-12-01
Fix end date doesn't bind correctly

aungzwin10:
Might be because of the typo, UI enddate field is bind with "credit.enddate" where "credit" object doesn't exist. The correct object name is "credit_act".


ehuelsmann (approved):
Looks good! Approved.

PR #6249, opened 2021-12-06
Support multi-currency for AR/AP batch import

aungzwin10:
Since LSMB support for multi-currency, the import should also support too. Previously the AR/AP import batch can import only with default currency. This allows user to import AR/AP batch with different currency defined in the system.
Add the example of the AR/AP batch import csv format


ehuelsmann:
@aungzwin10 Could you add a Changelog entry listing the added multi currency support in AR/AP batch uploads?

aungzwin10:
@ehuelsmann I think `Changed functionalities` section is more suitable than `Added functionalities` section to list this change since this change the already existing function? How do you think?

ehuelsmann:
> @ehuelsmann I think `Changed functionalities` section is more suitable than `Added functionalities` section to list this change since this change the already existing function? How do you think?

I think you're right. This is more changed than added.

PR #6250, opened 2021-12-06
Support multi-currency for AR/AP batch import

aungzwin10:
Since LSMB support for multi-currency, the import should also support too. Previously the AR/AP import batch can import only with default currency. This allows user to import AR/AP batch with different currency defined in the system.
Add the example of the AR/AP batch import csv format

### DIFF
--- a/lib/LedgerSMB/Scripts/import_csv.pm
+++ b/lib/LedgerSMB/Scripts/import_csv.pm
@@ -43,10 +43,10 @@ our $cols = {
     gl_multi        =>  ['debit_accno', 'credit_accno', 'amount', 'curr',
                          'amount_fx', 'reference', 'transdate', 'description',
                          'source_debit', 'source_credit', 'memo'],
-    ap_multi        =>  ['vendor', 'amount', 'account', 'ap', 'description',
-                        'invnumber', 'transdate'],
-    ar_multi        =>  ['customer', 'amount', 'account', 'ar', 'description',
-                        'invnumber', 'transdate'],
+    ap_multi        =>  ['vendor', 'amount', 'curr', 'fx_rate', 'account', 'ap',
+                        'description', 'invnumber', 'transdate'],
+    ar_multi        =>  ['customer', 'amount', 'curr', 'fx_rate', 'account', 'ar',
+                        'description', 'invnumber', 'transdate'],
     timecard        =>  ['employee', 'business_unit_id', 'transdate',
                         'partnumber', 'description', 'qty', 'non_billable',
                         'sellprice', 'allocated', 'notes', 'jctype', 'curr'],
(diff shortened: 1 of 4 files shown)


ehuelsmann:
@aungzwin10 hmm, this changes the API on 1.9. we try not to do that... Maybe it's possible to support the old *and* the new style? In that case it's an addition, which is not a problem for a released series (although patch releases generally don't add functionality).

aungzwin10:
@ehuelsmann Hmm, two new fields `curr and fx_rate` was added to the existing functionality. Actually it is more like a small improvement/addition.
Others changes are basically for readability which don't change the app behavior.

However changes in `sql/modules/Voucher.sql` is mandatory even without multi-currency because approving the batch doesn't approve the acc_trans line(kind of a bug).

sql/modules/Voucher.sql
```diff
@@ -365,6 +365,13 @@ $$
         WHERE id IN (select trans_id FROM voucher
                 WHERE batch_id = in_batch_id);
 
+        -- When approving the AR/AP batch import,
+        -- we need to approve the acc_trans line also.
+        UPDATE acc_trans SET approved = true
+        WHERE trans_id IN (select trans_id FROM voucher
+                WHERE batch_id = in_batch_id
+                AND batch_class IN (1, 2));
```
ehuelsmann, line 373: Can't the query below be expanded by batch classes 1 and 2? (so we have shorter, more readable, code?)

ehuelsmann: (At the same time, I'm wondering why the original author would not have included batch types 1 and 2 from the start...)

aungzwin10: > Can't the query below be expanded by batch classes 1 and 2? (so we have shorter, more readable, code?)

The original author write the below query is for payment/receipt batch. which join acc_trans and voucher on voucher_id.
However, the AR/AP Batch import doesn't set the voucher_id, it join voucher with trans_id. That's why I add separate query for that.

ehuelsmann:
> Hmm, two new fields `curr and fx_rate` was added to the existing functionality. Actually it is more like a small improvement/addition. Others changes are basically for readability which don't change the app behavior.

Yes, for the `master` branch, all those are a good idea. For the 1.9 branch, I think we should consider the upload file an "API": something people should be able to depend on. From that perspective, it would be good if we were able to support both the upload with the `curr` and `fx_rate` fields as well as the upload without those.That way, for anyone who already started using 1.9, there won't be a surprise that the upload suddenly stops working.
 
> However changes in `sql/modules/Voucher.sql` is mandatory even without multi-currency because approving the batch doesn't approve the acc_trans line(kind of a bug).

Agreed.

sql/modules/Voucher.sql
aungzwin10: > (At the same time, I'm wondering why the original author would not have included batch types 1 and 2 from the start...)

If I remember correctly, the lines in acc_trans are inserted as approved when save draft. But in 1.8, lines are inserted as unapproved when save draft and then update approved when Post or approve  from Transaction Approval.

aungzwin10:
> Yes, for the `master` branch, all those are a good idea. For the 1.9 branch, I think we should consider the upload file an "API": something people should be able to depend on. From that perspective, it would be good if we were able to support both the upload with the `curr` and `fx_rate` fields as well as the upload without those.That way, for anyone who already started using 1.9, there won't be a surprise that the upload suddenly stops working.

I see. So what if I add the check for existence of `curr` and `fx_rate` in 1.9? So, the upload use the default currency if these are not present in the file.

sql/modules/Voucher.sql
aungzwin10: > The original author write the below query is for payment/receipt batch. which join acc_trans and voucher on voucher_id.
> However, the AR/AP Batch import doesn't set the voucher_id, it join voucher with trans_id. That's why I add separate query for that.

I found a comment about this here.
https://github.com/ledgersmb/LedgerSMB/blob/009cdf6c6ec2b7ac0ebc3123047d67207cf2e6fc/sql/modules/Voucher.sql#L439-L440

ehuelsmann: > If I remember correctly, before 1.8, the lines in acc_trans are inserted as approved when save draft. But in 1.8, lines are inserted as unapproved when save draft and then update approved when Post or approve from Transaction Approval

Correct. I think in 1.9 that is one of the things we changed.

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -132,56 +132,92 @@ sub _aa_multi {
     my $vcst = $request->{dbh}->prepare(
         'select count(*) from entity_credit_account where meta_number = ?'
         );
+    my $currst = $request->{dbh}->prepare(
+        'select count(*) from currency where curr = ?'
+    );
     for my $ref (@$entries){
+        my %entry;
+        # Reference with name instead of index for better readability
+        if((scalar @$ref) > 7) {
```
ehuelsmann, line 141: I like this solution for 1.9

sql/modules/Voucher.sql
aungzwin10: Do you need more clarification? Or Shall we mark as resolve?

lib/LedgerSMB/Scripts/import_csv.pm
aungzwin10: Thanks

sql/modules/Voucher.sql
ehuelsmann: Marking resolved

PR #6265, opened 2021-12-08
Import screen UX improvement

### LINKED ISSUE #6264: UX improvement for CSV import screen
CSV Import/Import Batch screen need UX improvement.
Although screen contains 4 visible fields and 1 button, it will be better for the user if we provide more information.
Screen Shot 2021-12-08 at 1 34 20 PM
This screen use to import the following process.
```
gl             
gl_multi       
ar_multi       
ap_multi       
chart          
gifi           
sic            
timecard       
inventory      
inventory_ulti
parts          
services       
overhead
```

As you can see the same screen is used to import a lot of different process, and  it doesn't show the `title` and information such as how user should import or prepare file. The user cannot easily understand what he should do next.
Also all 4 visible fields are mandatory in some case and the code doesn't check for existence of these value. So the error message is not easily understandable by the user.

That's why I want to propose a solution which can improve UX. 
Screen Shot 2021-12-08 at 1 45 18 PM

Above is AP > Import Batch. I add `title` as `AP Transaction Batch Import` and some help information.
Although we provide example csv upload file in the code under `/doc` folder, in real life, normal user cannot access server and file system except sys-admin. User have no idea what is the format or what type of file he should use to import. So under the `title`, I add some sample information for normal user.

Then I add `required` for the field so that form cannot submit till the `required` fields are not fill.

Screen Shot 2021-12-08 at 1 48 39 PM

aungzwin10:
This PR is the proposed solution for UX improvement of import screen.

Closes #6264

### DIFF
--- a/UI/import_csv/import_csv.html
+++ b/UI/import_csv/import_csv.html
@@ -13,9 +13,10 @@
              name = "type"
              value = request.type
              } %]
-      [% IF request.multi %]
+      [% IF request.show_batch_title %]
       <div class="listtop">[% text('Batch Information') %]</div>
       [% END %]
+      [% IF request.show_input %]
       <div data-dojo-type="lsmb/layout/TableContainer"
            data-dojo-props="cols:1">
         <div class="inputrow" id = "referencerow">
@@ -54,6 +55,7 @@
                  } %]
         </div>
       </div>
+      [% END %]
       <div class="inputrow" id="filerow">
         [% PROCESS input element_data = {
                name = "import_file"
--- a/UI/import_csv/import_csv.html
+++ b/UI/import_csv/import_csv.html
@@ -13,7 +13,7 @@
              name = "type"
              value = request.type
              } %]
-      [% IF multi %]
+      [% IF request.multi %]
       <div class="listtop">[% text('Batch Information') %]</div>
       [% END %]
       <div data-dojo-type="lsmb/layout/TableContainer"
@@ -27,6 +27,7 @@
                  label = "_none_"
                  value = request.reference
                  size  = 15
+                 required = "true"
                  } %]
         </div>
         <div class="inputrow" id = "descriptionrow">
@@ -37,6 +38,7 @@
                  title = text('Description')
                  label = "_none_"
                  value = request.description
+                 required = "true"
                  } %]
         </div>
         <div class="inputrow" id="daterow">
@@ -48,6 +50,7 @@
                  name = "transdate"
                  value = request.transdate
                  size = 12
+                 required = "true"
                  } %]
         </div>
       </div>
@@ -58,6 +61,7 @@
                label = text('From File') #'
                class = "file"
                accept = ".csv"
+               required = "true"
                } %]
       </div>
       <div class="inputrow" id="buttonrow">
--- a/lib/LedgerSMB/Scripts/import_csv.pm
+++ b/lib/LedgerSMB/Scripts/import_csv.pm
@@ -106,7 +106,7 @@ my $header_info = {
         title => 'Inventory Batch Import',
         info => ''
     },
-    goods => {
+    parts => {
         title => 'Goods Import',
         info => ''
     },
@@ -648,7 +648,8 @@ sub begin_import {
 
     $request->{title} = $header_info->{$request->{type}}->{title};
     $request->{info} = $header_info->{$request->{type}}->{info};
-    $request->{multi} = 1 if $request->{type} =~ s/multi$//;
+    $request->{show_batch_title} = 1 if $request->{type} =~ m/multi$/;
+    $request->{show_input} = 1 if ($request->{type} =~ m/multi$/ or $request->{type} eq 'gl');
 
     # $request->{page_id} = $request->{type};
     # $request->{page_id} =~ s/_/-/;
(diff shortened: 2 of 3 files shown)


UI/import_csv/import_csv.html
```diff
@@ -27,6 +27,7 @@
                  label = "_none_"
                  value = request.reference
                  size  = 15
+                 required = "true"
```
ehuelsmann, line 30: These fields were not marked "Required" because they don't matter for GIFI/parts/overhead/services uploads. Maybe you can add an IF block around these inputs and a template variable to indicate whether or not they apply to the upload (like `request.multi`?)

aungzwin10: Please check the new commit

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -591,6 +646,11 @@ sub begin_import {
         $template_setup->{$request->{type}}($request);
     }
 
+    $request->{title} = $header_info->{$request->{type}}->{title};
+    $request->{info} = $header_info->{$request->{type}}->{info};
+    $request->{show_batch_title} = 1 if $request->{type} =~ m/multi$/;
+    $request->{show_input} = 1 if ($request->{type} =~ m/multi$/ or $request->{type} eq 'gl');
```
ehuelsmann, line 652: This looks like a good criterion, except that `inventory` (but not `inventory_multi`) uses `$request->{transdate}` as well, so that too would need the input to be shown.

lib/LedgerSMB/Scripts/import_csv.pm
```diff
@@ -65,6 +65,61 @@ our $cols = {
                          bom partsgroup / ],
 };
 
+my $header_info = {
+    gl => {
+        title => 'Journal Import',
+        info => ''
```
ehuelsmann, line 71:
```perl
info => "Imports a CSV file with the following columns (in exactly the order specified): @{$cols->{gl}}"
```

There's one problem with this approach: we can't mark these strings for translation when we use the interpolation as shown.

Or we'd need:
```perl
use LedgerSMB::I18N;

...

info => LedgerSMB::I18N::text('Imports a CSV file with the following columns (in exactly the order specified): [_1]', join(' ', @{$cols->{gl}}))
```

lib/LedgerSMB/Scripts/import_csv.pm
ehuelsmann: Maybe a more flexible solution is to add the "show_input" boolean to the "$header_info" hashes and make them 0/1 or ''/1 explicitly in those. That makes the process of showing them even more transparent.

aungzwin10: > This looks like a good criterion, except that `inventory` (but not `inventory_multi`) uses `$request->{transdate}` as well, so that too would need the input to be shown.

Inventory(multi) import use different html template file. So we dont need to consider for inventory in this case.

Previously, I add show_input in header_info hash. But when I realize only multi... and gl use these fields, I move to the current approach.

lib/LedgerSMB/Scripts/import_csv.pm
aungzwin10: I tried to use LedgerSMB:I18N as you said in import_csv.pm. I am not able to start the container with following error.
```
Uncaught exception from user code:

Error while loading /srv/ledgersmb/bin/ledgersmb-server.psgi: Can't locate object method "Locale" via package "LedgerSMB::App_State" at /srv/ledgersmb/lib/LedgerSMB/I18N.pm line 100.

Compilation failed in require at /usr/share/perl5/Module/Runtime.pm line 314.

Plack::Util::load_psgi("/srv/ledgersmb/bin/ledgersmb-server.psgi") called at /usr/share/perl5/Plack/Runner.pm line 181

Plack::Runner::__ANON__() called at /usr/share/perl5/Plack/Runner.pm line 24

Plack::Runner::__ANON__() called at /usr/share/perl5/Plack/Runner.pm line 24

Plack::Runner::__ANON__() called at /usr/share/perl5/Plack/Runner.pm line 24

Plack::Runner::__ANON__() called at /usr/share/perl5/Plack/Runner.pm line 24

Plack::Runner::__ANON__() called at /usr/share/perl5/Plack/Loader.pm line 55

Plack::Loader::preload_app(Plack::Loader=HASH(0x55bd1acfd530), CODE(0x55bd1af301e0)) called at /usr/share/perl5/Plack/Runner.pm line 276

Plack::Runner::run(Plack::Runner=HASH(0x55bd1aa6b8e0)) called at /usr/bin/plackup line 7
```

Do you have any clue why it happens?

ehuelsmann: oh. ok. I know what the problem is: this code needs to run inside a `sub` (the code calling `*::text()`).

aungzwin10: Thank you, I got it. So shall I commit the changes or shall we stop the PR here and move to vue?

Do you want this changes to back port to 1.9?

ehuelsmann: I don't think we need to backport to 1.9: it's not fixing an immediate problem although it does address a major usability problem. Normally, I'd expect that to go into 1.10, whereas for bugfixes, I'd expect us to be able to create a change which allows backporting.

The good part of moving this to Vue is that it raises a number of important questions which we need to solve when we want to continue with Vue: E.g. where/how do we get the (translated!) content (which we call 'info variable' in this PR).

PR #6277, opened 2021-12-12
Introduce a directory structure for the Vue front-end app

ehuelsmann:


ehuelsmann:
The immediate next step that I'd like to take here is adding a `locales` subdirectory with translations, so the `Home` component can be translated.

aungzwin10 (approved):

ylavoie (approved):

PR #6279, opened 2021-12-13
Move away from TT to custom web elements

ehuelsmann:
This PR introduces the following elements:

* Custom elements wrapping Dojo's Dijit elements
* The main login page as a Vue component
* A `grid` based layout instead of Dojo's `dijit/layout/TableContainer`
* Vue components (with documentation) for CSV imports
* Expand the use of `window.lsmbConfig` as a way to marshal information to the web client UI on page load
* Management of the `data-lsmb-done` attribute by the router (instead of each component), unless claimed by the component

and in the process removes:

* the loginForm widget
* the TableContainer widget
* the Template::Toolkit based UI templates for CSV imports
* XPath expressions which depend on the exact HTML generated by Dojo

In order to be able to do away with the TableContainer, it restructures the HTML on the table container's pages so as to fit the new grid patterns.

Closes #6265.


UI/src/elements/lsmb-text.js
```diff
@@ -0,0 +1,78 @@
...
+    adoptedCallback() {
+        if (this.dojoWidget
+            && this.dojoWidget.resize) {
+            this.dojoWidget.resize();
+        }
+    }
+    connectedCallback() {
+        if (this.dojoWidget) {
+            if (registry.byId(this.dojoWidget.id) !== this.dojoWidget) {
+                registry.add(this.dojoWidget);
```
ehuelsmann, line 20: This code exists to compensate for the behaviour of `lsmb/layout/TableLayout` which moves widgets around in the dom and (re)registers them. However, this workaround to prevent double registration actually does not solve the problem. We'll need to do away with the TableLayout's behaviour as it is defined today.

UI/src/elements/lsmb-text.js
```diff
@@ -0,0 +1,78 @@
...
+            }
+            return;
+        }
+        let root = this;
```
ehuelsmann, line 24: We're *not* creating a shadowRoot here, because that separates the styling from the rest of the page, but we want the body's list of classes to influence the theme used to render the widgets. Instead, we simply modify the DOM below the tag directly.

PR #6289, opened 2021-12-18
Display languages in print options.

### LINKED ISSUE #6267: Printable language options show a list of country instead of actual language
Transaction Approval > Batches > Batch Details

Screen Shot 2021-12-09 at 2 39 52 PM

ylavoie:
Use localized languages to set printing options. Fixes #6267


ylavoie:
UTF8 problem
image

lib/LedgerSMB/I18N.pm
```diff
@@ -135,6 +137,29 @@ sub get_country_list {
...
+=cut
+
+sub get_language_list {
+    my ($self,$language) = @_;
+    my $locale = Locales->new($language);
+    my @rows = $self->call_dbmethod(funcname => 'person__list_languages');
+    my @languages;
+    for (@rows) {
+        my $text = ucfirst($locale->get_language_from_code(lc($_->{code})) // $_->{description});
+        utf8::decode($text);
```
ehuelsmann, line 153: shouldn't we decode the text *before* doing the `ucfirst` transformation?

ehuelsmann: (and maybe we should make sure not to decode `$_->{description}` ?

ehuelsmann: or instead, maybe we need to use this hint: https://metacpan.org/pod/Locales#To-byte-or-not-to-byte,-that-is-the-question.-Ok,-I%E2%80%99ll-byte-%E2%80%A6

ehuelsmann: just one quick question (finding this only now; would have asked earlier if I had found it before): is https://metacpan.org/pod/Locale::CLDR#all_languages() helpful in reducing the extra dependency?

ylavoie: Much better with unicode. Fixed.
Using all_languages would bring back tons of languages, including sign, with only english descriptions. Thus we cannot display them in user language.

ehuelsmann:
@aungzwin10 could you test that this functionality correctly triggers the foreign language reports?

lib/LedgerSMB/I18N.pm
```diff
@@ -135,6 +136,28 @@ sub get_country_list {
...
 }
 
+=item get_language_list($language)
+
+Get a language localized list to allow user selection
+
+=cut
+
+sub get_language_list {
+    my ($self,$language) = @_;
```
ehuelsmann, line 146: In:

 https://github.com/ledgersmb/LedgerSMB/pull/6289/files#diff-535e82bbb947ea4d6c0b06744fcd4b564650c37c2adec048c531743dd014baa4R240

you're calling this routine, but not as `$i18n->get_language_list()`, but as `LedgerSMB::I18N::get_language_list( $report, $language );` -- if that's not an error then calling the report instance `$self` here is confusing.

ehuelsmann (approved):

ehuelsmann:
@ylavoie I applied changes in line with my earlier comments and squashed a number of changes into more targetted changes (eliminating back-and-forth on some of the topics).

PR #6304, opened 2021-12-27
Added support for linux distribution uos

wxp19831104:
Signed-off-by: xinpeng wang wangxinpeng@uniontech.com


ehuelsmann:
Thanks for your contribution!

PR #6306, opened 2021-12-28
Vue i18n

ylavoie:
This PR adds internationalization to the Vue components.
- The current locale data is converted to JSON format and duplicated in Vue;
- Language data is generated and loaded on demand, defaulting to english
- A small language selected is provided to demonstrate functionality,


ehuelsmann (changes requested):
There's at least one component missing: I tried to compile the sources (originally thought it would compile using `devdojo`), but can't test the branch.

UI/src/locales/de.json
```diff
@@ -0,0 +1 @@
...nicht in der Kartei!"],"[_1]: Vendor not on file!":[null,"[_1]: Lieferant nicht in der Kartei!"],"action not defined!":[null,"Aktion nicht definiert!"],"balance":[null,"Balance"],"city":[null,"Stadt"],"csv":[null,""],"days":[null,"Tage"],"debits":[null,"Belastungen"],"default":[null,""],"delete":[null,"lÃ¶schen"],"depreciation":[null,""],"disposal":[null,""],"done":[null,"Erledigt"],"e":[null,""],"ea":[null,"ea"],"failed":[null,"fehlgeschlagen"],"from_date":[null,""],"html":[null,""],"import":[null,"importieren"],"in months":[null,"in Monaten"],"in use":[null,""],"in years":[null,"in Jahren"],"is bigger than the amount available":[null,"ist grÃ¶sser als der verfÃ¼gbare Betrag"],"is lesser than 0":[null,"ist kleiner als 0"],"is lesser than the amount to be paid":[null,"ist weniger als der zu zahlende Betrag"],"is null":[null,"ist null"],"label":[null,""],"partial disposal":[null,"teilweisige Entsorgung"],"permalink":[null,"Permalink"],"production":[null,"Produktion"],"sent":[null,"gesendet"],"tex":[null,""],"time":[null,"Zeit"],"time(s)":[null,"Zeit(en)"],"to_date":[null,""],"unexpected error!":[null,"unerwarteter Fehler!"],"use of an overpayment":[null,"Nutzen einer Ãœberzahlung"]}
```
ehuelsmann, line 1: in `{"30":[null,"30"],...}` what is the `null` about?

ylavoie: Leftover from full json format. Converted to standard.

webpack.config.js
```diff
@@ -175,7 +189,55 @@ if (TARGET !== 'readme') {
         loaderConfig: require("./UI/js-src/lsmb/webpack.loaderConfig.js"),
         environment: { dojoRoot: "UI/js" }, // used at run time for non-packed resources (e.g. blank.gif)
         buildEnvironment: { dojoRoot: "node_modules" }, // used at build time
-        locales: ["en"],
+        locales: [
```
ehuelsmann: This is the current list of supported languages (should that be called locales?); shouldn't we make that dynamic?

UI/src/i18n.js
```diff
@@ -0,0 +1,91 @@
...
+    "ru",
+    "sv",
+    "tr",
+    "uk",
+    "zh_CN",
+    "zh_TW"
+];
+
+export function setI18nLanguage(i18n, locale) {
+    if (i18n.mode === "legacy") {
```
ehuelsmann, line 58: As we only run on Vue3, does it make sense to support "legacy" mode?

UI/src/i18n.js
```diff
@@ -0,0 +1,91 @@
...
+        i18n.global.locale.value = locale;
+    }
+    /**
+     * NOTE:
+     * If you need to specify the language setting for headers, such as the `fetch` API, set it here.
+     * The following is an example for axios.
+     *
+     * axios.defaults.headers.common['Accept-Language'] = locale
+     */
+    document.querySelector("html").setAttribute("lang", locale);
```
ehuelsmann, line 72: Our documents don't have the "lang" tag on the HTML attribute, which makes this line moot, unless we intend to start using the "lang" attribute?

UI/src/i18n.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export function setupI18n(options = { locale: "en" }) {
+    const i18n = createI18n(options);
+    setI18nLanguage(i18n, options.locale);
```
ehuelsmann, line 77: Why do you need this line, if the i18n object was just created using *exactly* those settings?

UI/src/i18n.js
```diff
@@ -0,0 +1,91 @@
...
+export async function loadLocaleMessages(i18n, locale) {
+    // load locale messages with dynamic import
+    const messages = await import(
+        /* webpackChunkName: "locale-[request]" */ `./locales/${locale}.json`
+    );
+
+    // set locale and locale message
+    i18n.global.setLocaleMessage(locale, messages.default);
+
+    return nextTick();
```
ehuelsmann, line 90: Why are you returning `nextTick()`? should `loadLocaleMessages()` return anything at all?

UI/src/locales/ar_EG.json
```diff
@@ -0,0 +1,1817 @@
+{
+    "30": "30",
```
ehuelsmann, line 2: You converted the entire translation lexicon, but we have only a very small amount of translation on the client side as of yet. Genuine question: should we only send the bits for which we have actual translations? (Obviously, we can only present the english version for anything we didn't translate yet, so maybe there's no reason to even *send* it? also, for any keys that aren't being used on the client for translation today, maybe we need not send those either? That'd be more efficient, but I have no idea about the downsides...)

webpack.config.js
ylavoie: Fixed

UI/src/i18n.js
ylavoie: No, it doesn't. Fixed

ylavoie:
This PR badly needs cleaning and be splitted to address different concerns. Closing.

PR #6308, opened 2021-12-28
Default language to configured value for login & setup

ylavoie:
Non logged in screens should use the configured default in ledgersmb.conf

### DIFF
--- a/lib/LedgerSMB/Template/UI.pm
+++ b/lib/LedgerSMB/Template/UI.pm
@@ -32,7 +32,7 @@ our $singleton;
 our @pre_render_cbs = (
     sub {
         my ($request, $template, $vars, $cvars) = @_;
-        $vars->{USER} = $request->{_user};
+        $vars->{USER} = $request->{_user} // { language => LedgerSMB::Sysconfig::language() };
         $vars->{DBNAME} = $request->{company};
         $vars->{locale} = $vars->{language} // $vars->{locale};
         $cvars->{locale} = $cvars->{language} // $cvars->{locale};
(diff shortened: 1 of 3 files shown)


lib/LedgerSMB/Template/UI.pm
```diff
@@ -32,7 +32,7 @@ our $singleton;
 our @pre_render_cbs = (
     sub {
         my ($request, $template, $vars, $cvars) = @_;
-        $vars->{USER} = $request->{_user};
+        $vars->{USER} = $request->{_user} // { language => LedgerSMB::Sysconfig::language() };
```
ehuelsmann, line 35: Why do we need this one? You handled setup and login in the other two changes...

UI/setup/edit_user.html
```diff
@@ -50,9 +50,12 @@
                 <td><label for="password">[% text('Password') %]</label>
                 </td>
                 <td>
-                    <input type="password"
-                           id="password" name="password" value=""
-                           autocomplete="off" />
+                  [% PROCESS input element_data = {
```
ehuelsmann, line 53: Hmm. why did you include this commit in the PR? It's already on master, no?

ylavoie: Rebase to remove it

lib/LedgerSMB/Template/UI.pm
ylavoie: _user is only set for a logged in user

ehuelsmann: True, but in which situations does that happen, now that you override the content of `_user` in login and setup handling?

ylavoie: Every time a user is not logged. I only override `{user}` default language in login and setup, not `{_user}`.

ehuelsmann: Are you telling me we have both `user` *and* `_user` keys in `$request`? can we eliminate `user` in favor of having only `_user`?

ylavoie: Fixed

PR #6317, opened 2022-01-01
Vue standardize home

ylavoie:
- Simplify Home.vue
- Extract the router


ehuelsmann:
Please don't do ec3b894. I'll explain more on live chat, but the summary is: all these checks make me feel really unhappy.

Other than that: there are 15 commits here, but I think there should only have been 2.

PR #6320, opened 2022-01-02
Vue i18n

ylavoie:
This PR adds translation capability to the Vue pages.
It also adds the current localized data in json format so that they can be loaded on demand


UI/src/i18n.js
```diff
@@ -0,0 +1,46 @@
...
+                /* webpackChunkName: "locale-[request]" */ `./locales/${locale}.json`
+            );
+
+            // set locale and locale message
+            i18n.global.setLocaleMessage(locale, messages);
+            i18n.global.locale = locale;
+        }
+        setI18nLanguage(locale);
+    }
+    return nextTick();
```
ehuelsmann, line 40: What does this `nextTick()` call do? Can't we just `return;`? I mean, I thought `nextTick()` takes a function argument which will be called on the next tick, but without an argument, it seems pointless?

UI/src/i18n.js
```diff
@@ -0,0 +1,46 @@
...
+    missingWarn: false, // warning off
+    locale: window.lsmbConfig.language,
+    fallbackLocale: "en",
+    messages: {}
+});
+
+function setI18nLanguage(locale) {
+    // Update document
+    document.querySelector("html").setAttribute("lang", locale);
+    document.querySelector("html").setAttribute("xml:lang", locale);
```
ehuelsmann, line 21: Since we stopped declaring our documents as `xhtml`, I think the `xml:lang` attribute is not appropriate/not required anymore?

UI/src/i18n.js
```diff
@@ -0,0 +1,46 @@
...
+            i18n.global.setLocaleMessage(locale, messages);
+            i18n.global.locale = locale;
+        }
+        setI18nLanguage(locale);
+    }
+    return nextTick();
+}
+
+// Load the user desired language if not default
+loadLocaleMessages(window.lsmbConfig.language);
```
ehuelsmann, line 42: Do we want to do this at module load time, or is it better to have modules load without side-effects and effectuate this setting before "app mounting"?

webpack.config.js
```diff
@@ -118,6 +129,20 @@ if (TARGET !== 'readme') {
         }
     };
 
+    const json = {
+        test: /\.(json5?|ya?ml)$/, // target json, json5, yaml and yml files
```
ehuelsmann, line 133: it looks rather weird that in the "json" block, YAML will be supported. Maybe you can add a bit more documentation on why this is a good idea anyhow?

webpack.config.js
```diff
@@ -68,6 +68,17 @@ if (TARGET !== 'readme') {
         ).filter((x, i, a) => a.indexOf(x) === i);
     }
 
+    function getPOFilenames(_path, extension) {
+        return fs
+            .readdirSync(_path)
+            .filter(
+                item =>
+                    fs.statSync(path.join(_path, item)).isFile() &&
+                    (extension === undefined || path.extname(item) === extension)
```
ehuelsmann, line 77: I see that `extension` is defined in all invocations. Is there a reason to allow it to be unspecified, still?

UI/src/i18n.js
ylavoie: Left over. Removed.

webpack.config.js
```diff
@@ -308,6 +333,17 @@ if (TARGET !== 'readme') {
             minRatio: 0.8
         }),
 
+        // Statics from build.
+        // The first one should come from user runtime environment instead.
```
ehuelsmann, line 337: What do you mean by "come from the user runtime environment"? Is that a build we expect the user to run? or is it the webbrowser environment? (do I have any other choices?)

ehuelsmann (commented):
I love the work so far. A few questions added here and there.

One overall remark: I don't see scripts to extract the translation strings nor do I see scripts to regenerate the JSON translation lexicons from the PO files. Are those to come?

One last question: if we have a requirement to build our JS assets, does it make sense to store the generated JSON files in our tree, or should we just generate those dynamically off of the PO files and use them as build artefacts?

UI/src/i18n.js
ylavoie: Have we? https://github.com/ledgersmb/LedgerSMB/blob/master/UI/lib/ui-header.html#L15 still uses xhtml

ylavoie:
> One overall remark: I don't see scripts to extract the translation strings nor do I see scripts to regenerate the JSON translation lexicons from the PO files. Are those to come?

They need to come in a subsequent PR. PO should become stale and JSON be pulled from the translation utility. Only the default language, `en`, should be directly updateable

webpack.config.js
ylavoie: Fallback should be `en`. Imposing `en` as default forces unlogged screens to follow that, even if a local admin has set his configuration to use something else.

UI/src/i18n.js
ehuelsmann: Hmm. well, one or the other is incorrect, because `<!DOCTYPE html>` indicates HTML5 which definitely isn't xhtml...

ehuelsmann: Thinking about it, we probably need to remove the XML namespace declaration for XHTML.

ylavoie: Agreed. Doing so

webpack.config.js
ehuelsmann: Sorry if I'm dense, but I still partially follow: as far as I understand the situation, you want the default to be `en`, except when the local admin has set a different language. I suppose that setting will be through `ledgersmb.conf`? Then your remark is referring to the language set in 'ledgersmb.conf' which needs to provide the default. And that `en` should be the fallback if no default was provided in `ledgersmb.conf`?

Maybe we can simplify things, because the default when no language is set in `ledgersmb.conf` already *is* `en`: see https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Sysconfig.pm#L293

webpack.config.js
ylavoie: I wanted to make sure that it could be reusable and would load only what I asked

webpack.config.js
ylavoie: Vue language files can be specified in JSON or YAML.

ehuelsmann: is it then better to call them `vueTranslations` instead?

webpack.config.js
ylavoie: Webpack has no access to the Perl code, nor the user ledgersmb.conf during compile, so I need to provide a sane default and then load the proper language at runtime.
If no language was set as default, then everything would be handled by the fallback entry

ehuelsmann: Ah! It's mostly a dummy setting! Now I get it. But should it say "we're providing a default here, but the true default will come at runtime from the browser context"?

webpack.config.js
ylavoie: FIxed

webpack.config.js
ylavoie: Not yet from the browser context. Default and fallback are set to `en`. When logging, the user preferences will provide the input to pull the proper locale.

ehuelsmann (approved):

UI/src/i18n.js
ylavoie: Works but still flicker

PR #6350, opened 2022-01-08
Vue i18n extract

ylavoie:
This PR add scripts to extract the Vue strings to allow internationalization.
- Vue text is extracted in JSON form and integrated to our current set, to allow translation with Transifex. Care is taken to set references to ease translator job.
- Translated text is exported to JSON format to allow compilation for Vue UI. Untranslated text is removed from JSON to allow defaulting to english


UI/src/locales/ar_EG.json
```diff
@@ -2,166 +2,72 @@
...
-    "AIM": "",
     "AP": "موردين",
     "AP Aging": "كشف حساب موردين",
-    "AP Invoice": "",
-    "AP Invoices": "",
     "AP Outstanding": "ارصدة موردين متبقية",
     "AP Transaction": "حركة حساب مورد",
     "AP Transactions": "حركات حساب الموردين",
-    "AP Voucher": "",
-    "AP account": "حساب مورد",
```
ehuelsmann, line 25: Hmm. If I understand correctly, the change is supposed to remove all *untranslated* strings. But this one (and several others) *is* translated.

locale/po/ar_EG.po
```diff
@@ -6719,42 +6666,43 @@ msgstr ""
...
 #: UI/Reports/filters/search_pricegroups.html:17
-#: UI/Reports/filters/timecards.html:48 UI/Reports/filters/unapproved.html:78
-#: UI/asset/search_asset.html:188 UI/asset/search_class.html:66
-#: UI/business_units/filter.html:41 sql/Pg-database.sql:2562
-#: sql/Pg-database.sql:2570 sql/Pg-database.sql:2583 sql/Pg-database.sql:2593
-#: sql/Pg-database.sql:2651 sql/Pg-database.sql:2698 sql/Pg-database.sql:2761
+#: UI/Reports/filters/timecards.html:48 UI/Reports/filters/unapproved.html:69
+#: sql/Pg-database.sql:2561 sql/Pg-database.sql:2569 sql/Pg-database.sql:2582
+#: sql/Pg-database.sql:2592 sql/Pg-database.sql:2650 sql/Pg-database.sql:2697
+#: sql/Pg-database.sql:2760
```
ehuelsmann: This change looks incorrect, because `UI/business_units/filter.html` *definitely* still has the "Search" string. (Check https://github.com/ledgersmb/LedgerSMB/blob/a5f4bdaee7dcff9b469a2ae50492aa41e96f5590/UI/business_units/filter.html#L41 )

cpanfile
```diff
@@ -20,6 +20,7 @@ requires 'Email::Sender::Transport::SMTP';
 requires 'Email::Stuffer';
 requires 'Feature::Compat::Try';
 requires 'File::Find::Rule';
+requires 'File::Grep';
```
ehuelsmann: If this dep is used in `utils/devel/` then shouldn't it be in `on devel =>` as a dependency? That might save the exclusion in `xt/01.2-deps.t` as well.

xt/01.2-deps.t
```diff
@@ -22,6 +22,7 @@ push @on_disk, 'bin/ledgersmb-server.psgi';
 ok_dependencies($file, \@on_disk,
                 phases => 'runtime',
                 ignores => [
+                    'File::Grep', # used to make PO from JSON
```
ehuelsmann: If you specify `File::Grep` as a `devel` dependency, this command won't go search for it (that's what the `'runtime'` is for).

package.json
```diff
@@ -132,7 +133,8 @@
     "lint:js": "eslint --ext .js ./",
     "lint:markdown": "markdownlint --config .markdownlint.json --ignore ./node_modules .",
     "profile": "webpack --progress=profile --json > logs/stats.json",
-    "readme": "webpack --mode=production --stats errors-warnings"
+    "readme": "webpack --mode=production --stats errors-warnings",
+    "vue-i18n-extract-translations": "vue-i18n-extract-translations -v './UI/src/**/*.?(js|vue)' -l './UI/src/locales/' --def-locale 'en' --key 'i18n' --keep-unused"
```
ehuelsmann: maybe `i18n-extract` would be good? the command that's there now is almost as hard to remember as the command it replaces...

ehuelsmann (changes requested):
There's a bit more checking and changing to be done re my prior comments.

UI/src/locales/ar_EG.json
ylavoie: "AP Voucher" is not translated in `ar_EG` so it is removed in JSON to default to the english version.
"AP account" is not used anymore in our code, so it is removed complelety

cpanfile
ylavoie: Indeed. Fixed.

xt/01.2-deps.t
ylavoie: Fixed

package.json
ylavoie: Removed in favor of 'utils/devel/extract-vue-template-translations.sh' which keeps JSON & PO synced.

ehuelsmann (approved):
Took me a while to figure out the `utils/devel/` part. Got it now.

Big +1 and thanks for the change!

PR #6353, opened 2022-01-09
Add back (and support) Language drop-down in aging statement generation

ehuelsmann:
Because templates can be created in language specific variants, we can't look up a single template and evaluate that for multiple languages, hoping that meets the user's expectations (although generally users will make use of a generic template with in-template translations, where this *does* meet their expectations).

Since different templates need to be run through different template evaluations (e.g. because the template generates a stand-alone PDF), the "Generate" button may now create multiple outputs. For Download (Screen sink), this is a problem because HTTP doesn't allow downloading multiple files in a single response. Hence the response gets packaged in a ZIP file.

For E-mail, this PR changes things so that the E-mail target will stop ignoring all but the first selected customer. Instead, all outgoing e-mails will be created and a list of available mails presented in the result window.

~~There's one problem with the status quo: if the user leaves the result window, there's no way to get back to the list of e-mails ready for sending... :-(~~


ylavoie (approved):

ehuelsmann:
> There's one problem with the status quo: if the user leaves the result window, there's no way to get back to the list of e-mails ready for sending...

There are several ways around this, all revolving around the idea of creating a workflow to handle the processing after the "Generate" button was clicked:
1. By using the `workflow_context` table to store workflow state, which in this case would be an array of all e-mail workflows that have been generated
2. By creating a separate table which stores the links between the e-mails and the workflow
3. By using the "spawned_workflow" concept which links several new workflows to a "parent" workflow (where the parent is supposed to have generated the spawned ones)

From the perspective of the reminder batch workflow, the following functionalities are important:
1. Sending the mails in batch, using a template text with variable replacement
2. Sending mails individually, being able to return to the index page which lists all generated e-mails
3. Tracking of dates of reminders (per invoice) in order to be able to send out reminders for "all invoices which didn't have one in the last 14 days"

Although the last point isn't part of the current solution, its design should take that requirement into account to prevent complete refactoring in the future.

Testing the current state of the branch against each of the requirements above, I'm coming to the conclusion that the first and second *can* be satisfied by each of the designs for handling the "Generate" button. However, none of them can satisfy the requirement for tracking reminder dates per invoice/customer, as far as I can see (and neither does the state of the branch at the moment).

ehuelsmann:
Going over the status quo in the chat channel, @freelock suggested that there be a search function to find both complete and incomplete workflows to address the issue of there being no way back into it.

This does not solve the cross reference desire against the invoices listed.

ehuelsmann:
Going over the scope of this PR:
* Adds content to the "language" drop-down in the Aging query/report, with  the counterparty's language selected by default
* Fixes downloading and printing of multiple aging reports (i.e. for different counterparties) with different language selections
* Changes e-mail output target from ignoring all but the first selected counterparty by generating a separate e-mail workflow for each counterparty with the aging statement as an attachment

By creating a workflow for this behaviour, a medium is created to store the links to the e-mails that were generated and a "callback" link for the e-mail workflow to return to the batch workflow in order to be able to proceed to the next e-mail.
With that setup, this functionality will allow batch-sending of mails at some point as well.

The question raised by @freelock as to filtering rules ("exclude all customers who received a statement in the last 6 days"), can't be supported by the functionality as described above. It does allow to close this PR *and* it allows to extend the functionality at a later point.

ehuelsmann:
This PR opens the option to sending aging statements in bulk.

PR #6354, opened 2022-01-09
Freshen locale/LedgerSMB.pot and locale/po for 1.10

ylavoie:
Make sure that the current locale data is in line with current 1.10


ehuelsmann:
The Opera failure is just intermittent.

PR #6356, opened 2022-01-09
Vue localize home

ylavoie:
Add localization calls to `Home.vue`
Add the fr_CA text as example.

### DIFF
--- a/UI/src/components/Home.vue
+++ b/UI/src/components/Home.vue
@@ -4,8 +4,7 @@
             <h1 style="margin-bottom: 3em">{{ $t(`Welcome to LedgerSMB`) }}</h1>
             <div class="welcomed w1">
                 <h2>{{ $t(`What is LedgerSMB`) }}</h2>
-                <p>{{ $t(`LedgerSMB is a double entry accounting system offering a wide
-                  variety of related functionalities, such as`) }}</p>
+                <p>{{ $t(`i18n.home.ledgersmbIntro`) }}</p>
                 <ul>
                     <li>{{ $t(`invoicing`) }}</li>
                     <li>{{ $t(`bank reconciliation`) }}</li>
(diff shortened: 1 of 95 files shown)


UI/src/components/Home.vue
```diff
@@ -4,8 +4,7 @@
             <h1 style="margin-bottom: 3em">{{ $t(`Welcome to LedgerSMB`) }}</h1>
             <div class="welcomed w1">
                 <h2>{{ $t(`What is LedgerSMB`) }}</h2>
-                <p>{{ $t(`LedgerSMB is a double entry accounting system offering a wide
-                  variety of related functionalities, such as`) }}</p>
+                <p>{{ $t(`i18n.home.ledgersmbIntro`) }}</p>
```
ehuelsmann, line 7: Why did you switch to using keys as indexes to the translation instead of the original text?

UI/src/components/Home.vue
```diff
@@ -18,44 +17,59 @@
             </div>
             <div class="welcomed w2">
                 <h2>{{ $t(`Community resources`) }}</h2>
-                <p>{{ $t(`Community support for LedgerSMB is a available through various
-                  channels, ranging from static all the way to near-realtime (depending
-                  on availability of people):`) }}</p>
+                <p>{{ $t(`i18n.home.communitySupport`) }}</p>
```
ehuelsmann, line 20: Why are you using backticks in some cases and double quotes in others?

UI/src/components/Home.vue
```diff
@@ -18,44 +17,59 @@
...
-                  <li>Fix a <a href="https://github.com/ledgersmb/LedgerSMB/issues?q=is%3aopen%20is%3aissue%20label%3abite-sized%20-label%3aenhancement">{{ $t(`bug`) }}</a> and
-                    submit a &quot;Pull Request&quot;<br />
-                    When you do, please start with one marked 'bite-sized' and contact
-                    the development team through IRC and/or matrix for help getting
-                    started.</li>
+                  <i18n-t keypath="i18n.home.reportBugs.text" tag="li" scope="global">
+                    <template #url1><a :href="reportBugs_url1" target="_blank" rel="noopener noreferrer">
+                        {{ $t("i18n.home.reportBugs.link1") }}
+                    </a></template>
+                    <template #url2><a :href="reportBugs_url2" target="_blank" rel="noopener noreferrer">
```
ehuelsmann, line 59: 1. why are you using an attribute bound to an attribute in the object?
2. what does the `#url1` part in `<template #url1>` mean?

UI/src/components/Home.vue
ylavoie: To remove white spaces imposed by line wrap in the translated strings.

UI/src/components/Home.vue
ylavoie: reportBugs is defined in `en.json` as
```
            "reportBugs": {
                "link1": "the project's GitHub bug tracker",
                "link2": "mailing lists",
                "text": "Report your bugs and problems to {url1} or {url2}"
            },
```
so that the texts can be translated properly and that the `{url}` variables can be substituted for the 2 url required

UI/src/components/Home.vue
ylavoie: Fixed

UI/src/components/Home.vue
ehuelsmann: Ok. So, the Home page isn't a good example for in-page translation of HTML? I mean, it's probably better to have a  page like "Login" which has only a few fractions of text, instead of longer sections of explanatory text with external references. Maybe our "welcome" or Home page is best translated by storing HTML blobs on the server?

UI/src/components/Home.vue
ehuelsmann: I still don't understand. I'm likely behind on the capabilities of the library we're using. Do you have a reference for me to read?

ylavoie: https://kazupon.github.io/vue-i18n/guide/formatting.html#named-formatting

ehuelsmann:
I think we should close this one: full page translations are probably better supported by translating the entire page and serving the translated version instead of the english version. This translation got ugly quickly as this PR proves.

PR #6369, opened 2022-01-18
Fix deletion of *all* payment links

### LINKED ISSUE #6368: Deleting the batch payment/receipt wipe all lines from `payment_links` table
Deleting the batch payment/receipt or payment/receipt reversal using Transaction Approval > Batches, delete all lines from `payment_links` table.

Step to reproduce

1. Create Batch payment
2. Go to Transaction Approval > Batches
3. Select the batch and click `Delete`
4. Check the payment_links

ehuelsmann:
When deleting a batch, the procedure would delete *all* payment links;
after this change, there's now a restiction on the subselect relating
the DELETE to the subquery, restricting the results in the subquery.

(Before, *any* existance would delete *all* payment links, not just the
ones relating to the batch being deleted.)

While at it, remove code duplication.

Fixes #6368.

### DIFF
--- a/sql/modules/Voucher.sql
+++ b/sql/modules/Voucher.sql
@@ -458,17 +458,17 @@ BEGIN
                        (select id from voucher where batch_id = in_batch_id)
                );
 
+        WITH deleted_payment_ids AS (
+           DELETE FROM payment_links p
+            WHERE EXISTS (select 1 from acc_trans a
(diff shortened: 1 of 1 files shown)


aungzwin10 (approved):
@ehuelsmann Thank you for the fix. I have tested the PR and it works well.

PR #6375, opened 2022-01-23
Cldr to locales

ylavoie:
Remove Locale::CLDR in favor of Locales to significantly lower the amount of dependencies.

### DIFF
--- a/lib/LedgerSMB/I18N.pm
+++ b/lib/LedgerSMB/I18N.pm
@@ -19,7 +19,6 @@ we look only to the current locale.
 
 #use v5.24.0;
 
-use Locale::CLDR;
 use Locales unicode => 1;
 use Moose::Role;
 use namespace::autoclean;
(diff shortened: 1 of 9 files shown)


ylavoie:
Locales 0,34 is using CLDR data from 2016 and misses some variants like ar_EG, es_MX and a few others. Locale::CLDR was using 2014 data, so we've improved but bringing data up to CLDR v41 would be desirable.

lib/LedgerSMB/I18N.pm
```diff
@@ -127,11 +126,12 @@ Get a country localized list to allow user selection
...
 sub get_country_list {
     my $language = shift;
-    my %regions = Locale::CLDR->new($language)->all_regions->%*;
+    my $locale = Locales->new($language);
+    my %regions = $locale->get_territory_lookup();
     return [
         sort { $a->{text} cmp $b->{text} }
         map { +{ value => uc($_),
-                 text  => $regions{$_} }
+                 text  => $regions{lc($_)} }
```
ehuelsmann, line 134: When iterating over the keys of `%regions`, I'd expect you shouldn't be transforming the key before doing a lookup (i.e. no `lc()`)?

ehuelsmann (commented):
Should we wait for `Locales` to be updated? From what I understood, the version it uses was 29 with 40 the latest release and there's an outstanding request to the author to update it?

lib/LedgerSMB/I18N.pm
ylavoie: The keys are already lowercased but I fixed it.

ylavoie:
> Should we wait for `Locales` to be updated? From what I understood, the version it uses was 29 with 40 the latest release and there's an outstanding request to the author to update it?

I'm pushing on the maintainer to update `Locales` to a more recent `CLDR` version. I'd say let's wait a few weeks, to give him time to answer

PR #6388, opened 2022-02-01
Fix broken CSV report output

ehuelsmann:
In 1.9, CSV report output was broken while trying to fix earlier
problems caused by attempts to address the problem that PGNumbers
need to be run through `.to_output()` before becoming 'text that
can be quoted'.


ehuelsmann:
@freelock This fixes the CSV report problem you pointed out in the chat channel.

PR #6423, opened 2022-02-22
Restrict tests for translatable texts to project sources

ylavoie:
Prevent searches in local directories (.git, .github, .circleci, .local, ...)

### DIFF
--- a/xt/06.1-Perl-PO.t
+++ b/xt/06.1-Perl-PO.t
@@ -22,7 +22,8 @@ if ($ENV{COVERAGE} && $ENV{CI}) {
 
 my @on_disk;
 sub collect {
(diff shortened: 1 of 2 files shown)


xt/06.1-Perl-PO.t
```diff
@@ -22,7 +22,7 @@ if ($ENV{COVERAGE} && $ENV{CI}) {
 
 my @on_disk;
 sub collect {
-    return if $File::Find::dir  =~ m(^blib|xt/lib/|xt/66-cucumber/|LaTex|.local/);
+    return if $File::Find::dir  =~ m(^blib|xt/lib/|xt/66-cucumber/|LaTex|^.*/);
```
ehuelsmann, line 25: the old and the new regex are both wrong.

```suggestion
    return if $File::Find::dir  =~ m!^(blib|xt/lib/|xt/66-cucumber/|LaTex|\.)!;
```

Issue #6428, opened 2022-02-28
Search Goods and Services - some options do not work


pasman1:
I have installed LedgerSMB 1.9.10 from tarball (on Raspbian Bullseye aarch64)
Other components are: perl 5.31.1 , postgresql 13.5, starman 0.4015.

When searching for the parts on the page **Search Goods and Services**,
changing option **Part criteria - > Type** to Services or Assemblies, always returns all parts.
(other search criteria are default)

Additionally option **Include in Report - > Warehouse** do not include a Warehouse column

Best Regards.

ehuelsmann:
Thanks for the report! This looks like it should be rather easy to fix. I'll have a look soonish.

pasman1:
This sql looks good. Probably there is problem with sending the value of **in_parttype** from a form to this function.

ehuelsmann:
The above referenced PR fixes the problem with the parts selection. I'm looking at how to add the warehouse to the output (without being totally confusing).

PR #6429, opened 2022-03-01
Fix selection of part types in Goods search

ehuelsmann:
Re #6428.


ehuelsmann:
Only the first of the two commits can be backported to 1.9: the second commit contains changes to the database schema which we really try to avoid very hard in patch releases.

PR #6444, opened 2022-03-05
Remove explicit static Dijit references in webpack config

ylavoie:
Make sure to pull Dijit references from old code to remove static references from webpack config.


ehuelsmann:
Nice!

PR #6515, opened 2022-04-15
Restore GIFI Balance Sheet report

ylavoie:
Make sure that the GIFI acounts are used in Income Statement report when desired.


ehuelsmann (commented):
@ylavoie the title is "Restore ... *Income Statement*". However, you're editing the **Balance sheet** source code. How should I understand that???

PR #6519, opened 2022-04-15
Restore GIFI Balance Sheet report

ylavoie:
Make sure that the GIFI acounts are used in Balance Sheet report when desired.


ehuelsmann:
Although the 1.6 branch isn't maintained anymore:

I verified the 1.6 branch with these steps:

* git clone "ledgersmb-docker"
* git checkout 1.6
* docker-compose pull
* docker-compose up
* create a database through setup.pl
* load the standard US coa
* in the CoA screen, add a gifi code to "current account" and "petty cash"
* post a transaction, moving some amount between the two accounts mentioned above
* run the balance sheet report with "GIFI" checked

This branch doesn't exhibit the problem mentioned.

PR #6594, opened 2022-07-02
Configuration through APIs and Vue components

ehuelsmann:
Please use SQUASH MERGE to merge this PR!

This PR implements management of **warehouses** through an API, using Vue components; the first to use this new infrastructure. The reason to pick this one is that it allows to phase out "old code" (i.e. a part of "am.pl" and "AM.pm").

The code has been set up to allow relatively quick on-boarding of other APIs, by providing re-usable components, both in the UI as well as setting a pattern for coding APIs.

The following steps have to be completed before this PR is to be considered complete:

* [x] Addressing string translation in the Vue components
* [x] Hooking up the new UI to the menu
* [x] Generation and verification of real ETag values (to replace the current faked ones)
* [x] Factoring out the style rules which have been incorporated on the component level for ease of experimentation
* [x] Addressing the issue that the page is still presented as configuration page when the user has read-only access
* ~~Restore consistency of Toast-popups during navigation by popping up a toast when navigating to existing Vue components (the CSV imports)~~  (no loading from the server takes place; no need to have a Toast here)

A more lengthy task will be separated out into its own issue (#6597), which is about abstracting out the request and response validation to reduce boiler plate code.


ehuelsmann:
@ylavoie there are 3 failures here. One is in `xt/01-html-checks.t` and that one is bogus, as far as I'm concerned: There's a `<Toaster />` tag there. The complaint is that it's not closed, but any decent browser considers that a "self-closing" tag. Can we accept self-closing tags in the tests? I don't see much value to change the code to `<Toaster></Toaster>`...

ylavoie:
> @ylavoie there are 3 failures here. One is in `xt/01-html-checks.t` and that one is bogus, as far as I'm concerned: There's a `<Toaster />` tag there. The complaint is that it's not closed, but any decent browser considers that a "self-closing" tag. Can we accept self-closing tags in the tests? I don't see much value to change the code to `<Toaster></Toaster>`...

Fixed in #6595

ylavoie (approved):

PR #6604, opened 2022-07-06
Migrate price groups administration to Vue ConfigTable

ehuelsmann:


ehuelsmann:
I think this PR is complete: the last items in "am.pl" and "AM.pm" are Taxes and Recurring Transactions. The former doesn't map well to the ConfigTable view, because it's a table with multiple sections, each with its own expansion dynamics. The latter is simply too complex to move into an API at this point.

The "Parts groups" could have been a nice functionality to move over, but this essentially models a tree structure (with parent groups), which doesn't work well with the current ConfigTable view.

Other candidates (which don't eliminate "old code", but are linear dimensions and therefor good candidates) are Sequences and Currencies > Edit currencies. I consider neither of high importance at the moment, although I'd welcome contributions!

ehuelsmann:
In other news: I think there's a *lot* of code duplication on the Perl side of things. However, I'm thinking this was mostly a write-once effort. If we end up editing the resulting code reguarly, I think we need to further abstract it out. For now, I'm ready to accept the result.

ylavoie (approved):

PR #6608, opened 2022-07-08
Exclude dirs with dependencies from i18n extraction

ehuelsmann:


ehuelsmann:
I don't want to put energy into this at the moment.

PR #6615, opened 2022-07-09
Fix rebuild_pot file selection

ylavoie:
- Remove dot-files from source selection for Perl translation files
- Fix file selection pattern to exclude compiled JS files

### DIFF
--- a/utils/devel/rebuild_pot.sh
+++ b/utils/devel/rebuild_pot.sh
@@ -29,12 +29,12 @@ EOF
 
 # EXTRACT STRINGS AND CREATE POT
 find . -name '*.pl' -o -name '*.pm' | \
-  grep -v '\(b\|xt/\)lib/\|UI/pod/' | grep -v LaTeX | sort | \
+  grep -v '\(b\|xt/\)lib/\|UI/pod/\|\./\.[a-z]*\+' | grep -v LaTeX | sort | \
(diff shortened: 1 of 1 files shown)


utils/devel/rebuild_pot.sh
```diff
@@ -29,12 +29,12 @@ EOF
...
 # EXTRACT STRINGS AND CREATE POT
 find . -name '*.pl' -o -name '*.pm' | \
-  grep -v '\(b\|xt/\)lib/\|UI/pod/' | grep -v LaTeX | sort | \
+  grep -v '\(b\|xt/\)lib/\|UI/pod/\|\./\.[a-z]*\+' | grep -v LaTeX | sort | \
   utils/devel/extract-perl >> locale/LedgerSMB.pot
 
 find UI/ templates/ t/data/ \
      -name '*.html' -o -name '*.tex' -o -name '*.csv' | \
-   grep -v 'UI/\(js\|pod\)/)' | sort | \
+   grep -v 'UI/\(js\|pod\)' | sort | \
```
ehuelsmann, line 37: I think the closing paren was indeed a mistake, but don't you think the forward slash was there for the reason that it wants to exclude files in those directories, but not the ones in `UI/js-src/`?

ylavoie: Absolutely. Misread it for a backslash

PR #6625, opened 2022-07-14
Pr/6611/vue router

### LINKED ISSUE #6635: Update all non-major dependencies (master) (`browserslist`, `caniuse-lite`, `vue-router`)

ylavoie:
Make sure that the URI fed to the router are relative.
Fixes #6635

### DIFF
--- a/UI/src/main-vue.js
+++ b/UI/src/main-vue.js
@@ -45,7 +45,11 @@ if (document.getElementById("main")) {
                     document.body.setAttribute("data-lsmb-done", "true");
                 });
             });
-            window.__lsmbLoadLink = (url) => this.$router.push(url);
+            window.__lsmbLoadLink = (url) =>
+                this.$router.push(
+                    // eslint-disable-next-line no-useless-escape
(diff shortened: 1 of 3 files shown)


UI/src/main-vue.js
```diff
@@ -45,7 +45,11 @@ if (document.getElementById("main")) {
                     document.body.setAttribute("data-lsmb-done", "true");
                 });
             });
-            window.__lsmbLoadLink = (url) => this.$router.push(url);
+            window.__lsmbLoadLink = (url) =>
+                this.$router.push(
+                    // eslint-disable-next-line no-useless-escape
+                    url.replace(/^https?:\/\/(?:[^@\/]+)/, "")
```
ehuelsmann, line 51: Why do you exclude the @-sign? It can be used to separate username and password from the hostname, but if it does, the regex doesn't match. (No idea if this makes sense in the context ofLedgerSMB, actually)

ylavoie: I didn't think it made sense for LedgerSMB.

ehuelsmann (approved):

PR #6643, opened 2022-07-18
Vue localize login page

ylavoie:
- Integrates Vue translation files
- Add translations to LoginPage and ServerUI
- Make LoginPage display adjust for locale texts variations
- Adjust for Arabic languages to present a right to left presentation


UI/src/i18n.js
```diff
@@ -15,42 +16,53 @@ function _mapLocale(locale) {
...
+            i18n.global.setLocaleMessage(_locale, messages);
         }
+    } else {
+        _locale = "en";
     }
+    setI18nLanguage(_locale);
     return nextTick();
 }
 
+loadLocaleMessages(window.lsmbConfig.language);
```
ehuelsmann, line 66: Since this function is `async`, shouldn't this be prefixed with "await"?

Also: this duplicates the `beforeCreate` handler in `main-vue.js`. I'm thinking that, since the `beforeCreate` handler can't be `async`, we probably want to remove that invocation in favor of this one?

webpack.config.js
```diff
@@ -343,12 +340,21 @@ if (TARGET !== "readme") {
...
     ];
 
-    var pluginsList = prodMode ? pluginsProd : pluginsDev;
+    var pluginsList = prodMode
+        ? [
+              // Clean UI/js before building (must be first)
+              new CleanWebpackPlugin(CleanWebpackPluginOptions),
+              ...pluginsProd
+          ]
+        : pluginsDev;
```
ehuelsmann, line 357: What's the relationship to this PR?

ehuelsmann (changes requested):
Very nice PR overall. Love the fact that the error strings from LoginPage now appear in the translation dictionary!

I've marked one nit that I think warrants further discussion or work: we are now calling `loadLocaleMessages` twice during startup, with the same arguments.

webpack.config.js
ylavoie: None. Removed.

UI/src/i18n.js
ylavoie: Duplicated code removed. `beforeCreate` could have been made async though.

ehuelsmann: I see that's how it's been resolved now. Thanks! No further remarks or nits. Only praise! :+1:

PR #6644, opened 2022-07-19
Webpack dev server

ylavoie:
Adding `webpack devServer` to our configuration to ease UI development


webpack.config.js
```diff
@@ -243,10 +250,21 @@ if (TARGET !== "readme") {
...
-        new CleanWebpackPlugin(CleanWebpackPluginOptions),
+    // Compile bootstrap module as a virtual one
+    const VirtualModulesPluginOptions = {
+        "./bootstrap.js":
+            `/* eslint-disable */\n` +
+            `define(["dojo/parser","dojo/ready","` +
+            includedRequires.join('","') +
+            `"], function(parser, ready) {\n` +
+            `    ready(function() {\n` +
+            `        console.log('Bootstrapped');\n` +
```
ehuelsmann, line 261: this maybe only for dev builds? (the `console.log` I mean)

package.json
```diff
@@ -277,6 +279,11 @@
       ]
     }
   },
+  "postcss": {
+    "plugins": [
+      "autoprefixer"
```
ehuelsmann, line 284: Didn't we decide that we didn't need autoprefixer?

webpack.config.js
```diff
@@ -453,7 +476,46 @@ if (TARGET !== "readme") {
...
+            },
+            compress: true,
+            devMiddleware: {
+                index: false,
+                serverSideRender: true,
+                writeToDisk: true // Required for Perl TT
+            },
+            hot: true,
+            port: 9000,
+            proxy: {
```
ehuelsmann, line 499: Very complete solution that it's also able to proxy! In this scenario we really don't need the nginx proxy, but I guess we'll just leave it in there for production simulation, right?

webpack.config.js
```diff
@@ -453,7 +476,46 @@ if (TARGET !== "readme") {
...
+                progress: true
+            },
+            compress: true,
+            devMiddleware: {
+                index: false,
+                serverSideRender: true,
+                writeToDisk: true // Required for Perl TT
+            },
+            hot: true,
+            port: 9000,
```
ehuelsmann, line 498: Locally, I added a row `host: "0.0.0.0",` before the port in order to be able to access the dev server from outside the container I'm running this in.

ehuelsmann (commented):
Looks like this is almost done and with my testing, a tremendous improvement when doing JS development!

package.json
ylavoie: Leftover. Removed

webpack.config.js
ylavoie: Exact

webpack.config.js
ylavoie: Debugging leftover. Removed.

ehuelsmann (approved):

PR #6649, opened 2022-07-22
Merge templates targetting {xe,pdf}latex

ehuelsmann:
By using the 'iftex' package, a single set of templates can
target both, with a few conditionalized commands.

This commit introduces an option to select a template
processor (PDF generator) for all templates at once by making
an option available to the templates through dependency injection.

Before this commit, template sets had a tendency to diverge
and explicit effort needed to be spent to keep them aligned; this
was aggrevated by the fact that the template set holds *all*
templates, not only the TeX ones.


ylavoie (approved):
Nice simplification!

Issue #6606, opened 2022-07-07
Add functionalities lost when moving configuration items to ConfigTable


ehuelsmann:
The following things were lost while moving our configuration code to Vue's ConfigTable code:

* [x] Reports listing the items in the configuration table, per table:
  * [x] Warehouses
  * [x] SIC
  * [x] Languages
  * [x] Pricegroups
  * [x] Type of Business
  * [x] GIFI
* [ ] Import:
  * [ ] SIC codes
  * [ ] GIFI codes

ehuelsmann:
Documenting the plan so I don't forget:

* Move the items returned on a collection from top-level to the `items` key
* Add `_links` containing "available next actions" (HATEOAS-like)
* Add elements to the`_links` array with `rel="download"` and an `href=""` target ; one for each downloadable format

ehuelsmann:
The items above are now sitting in a branch, with an implementation to download alternative representations for Languages.

PR #6672, opened 2022-07-25
Refactor reports to be useable for services

ehuelsmann:
In order to be able to use our reports from the web services implementations - which we want in order to generate other-than-JSON representations - we need to be able to evaluate our reports in a context which does *not* provide a `$request` instance.

This PR also adds download links to `ConfigTable` based configuration components and fixes (and simplifies) the PDF/PS as well as the HTML `display_report` and `dynatable` "document templates" (i.e. the ones which don't go in the UI). It finishes the first of the two main bullets in #6606.

### DIFF
--- a/UI/Reports/PNL.html
+++ b/UI/Reports/PNL.html
@@ -2,8 +2,6 @@
 account_data = report.account_data;
 
 FORMATS = LIST_FORMATS();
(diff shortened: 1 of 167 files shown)


ehuelsmann:
@ylavoie this PR is ready for review!

ylavoie (approved):

PR #6684, opened 2022-07-30
Ledgersmb admin user

ylavoie:
This add user management to `ledgersmb-admin`, allowing to list, create or change users.

### DIFF
--- a/lib/LedgerSMB/Admin/Command/user.pm
+++ b/lib/LedgerSMB/Admin/Command/user.pm
@@ -279,7 +279,8 @@ sub create {
         return 1;
     }
     #TODO: Fix validity
-    $dbh->do(qq(ALTER USER "$username" VALID UNTIL 'infinity'));
+    my $ident_username=$dbh->quote_identifier($username);
+    $dbh->do(qq(ALTER USER $ident_username VALID UNTIL 'infinity'));
 
     # Add permissions
     return 1 if $permission && !$self->_add_permissions($dbh, $user);
@@ -351,7 +352,8 @@ sub _create_user {
         _dbh => $dbh,
         );
     $user->create($args{password});
-    $dbh->do(qq(ALTER USER "$_username" VALID UNTIL 'infinity'));
+    my $ident_username=$dbh->quote_identifier($username);
+    $dbh->do(qq(ALTER USER $ident_username VALID UNTIL 'infinity'));
 
     return $user;
 }
--- a/lib/LedgerSMB/Admin/Command/user.pm
+++ b/lib/LedgerSMB/Admin/Command/user.pm
@@ -1,0 +1,582 @@
+package LedgerSMB::Admin::Command::user;
+
+=head1 NAME
+
+LedgerSMB::Admin::Command::user - ledgersmb-admin 'user' command
+
+=cut
+
+use strict;
+use warnings;
+
+use Getopt::Long qw(GetOptionsFromArray);
+use LedgerSMB::Admin::Command;
+use LedgerSMB::App_State;
+use LedgerSMB::Company;
+use LedgerSMB::Database;
+use LedgerSMB::DBObject::User;
+use LedgerSMB::Entity::Person::Employee;
+use LedgerSMB::Entity::User;
+use LedgerSMB::PGDate;
+
+use Moose;
+extends 'LedgerSMB::Admin::Command';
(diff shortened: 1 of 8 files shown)


lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -0,0 +1,582 @@
...
+
+use Moose;
+extends 'LedgerSMB::Admin::Command';
+use namespace::autoclean;
+
+use Feature::Compat::Try;
+
+# LedgerSMB database interface
+has db => (is => 'rw');
+has super_dbh => (is => 'rw',
```
ehuelsmann, line 30: Not used?

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -0,0 +1,582 @@
...
+                  predicate => '_has_super_dbh');
+has admin_dbh => (is => 'rw', lazy => 1,
+                  builder => '_build_admin_dbh',
+                  clearer => '_clear_admin_dbh',
+                  predicate => '_has_admin_dbh',
```
ehuelsmann, line 35: The builder isn't defined.Do you really eed the admin_dbh, or isn't a remnant of an early version?

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -0,0 +1,582 @@
...
+    my $_username = shift // $username;
+
+    # The code below doesn't provide the entity_id
+    #my $user_obj = LedgerSMB::DBObject::User->new();
+    #$user_obj->set_dbh($dbh);
+    #my @users = @{$user_obj->get_all_users};
+    #my ($user) = grep { $_username eq $_->{username} } @users;
+
+    #This ugly hack does
+    my $sth = $dbh->prepare("SELECT id, entity_id FROM users WHERE username='$username'");
```
ehuelsmann, line 89: Please use placeholders and not string interpolation with SQL commands.

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -0,0 +1,582 @@
...
+    if (!$user) {
+        $self->logger->error('Invalid user');
+        $dbh->rollback;
+        return 1;
+    }
+    #TODO: Fix validity
+    $dbh->do(qq(ALTER USER "$username" VALID UNTIL 'infinity'));
+
+    # Add permissions
+    return 1 if $permission && !$self->_add_permissions($dbh, $user);
```
ehuelsmann, line 295: This prevents adding and removing permissions in a single invocation. Is that intentional?

ehuelsmann (changes requested):
Yep. This one is long over due.... Thanks for taking the time to do it!

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Remant. Removed

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: No, this fails early when permissions were given but aren't valid.
Same for no-permissions

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Remnant. Removed

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -0,0 +1,575 @@
...
+use LedgerSMB::PGDate;
+
+use Moose;
+extends 'LedgerSMB::Admin::Command';
+use namespace::autoclean;
+
+use Feature::Compat::Try;
+
+# LedgerSMB database interface
+has db => (is => 'rw');
```
ehuelsmann, line 29: Are you sure this one is correct? Shouldn't it inherit it from `LedgerSMB::Admin::Command`?

ylavoie: Indeed. Fixed.

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Done wherever possible.

ehuelsmann (approved):

PR #6697, opened 2022-08-06
Add tool to do ini -> yaml config translation

ehuelsmann:
Based on the Sysconf functionality which generates the config hash,
serialize the configuration using YAML::PP for use as a yaml config
file.


ehuelsmann:
Already merged??

PR #6698, opened 2022-08-06
Proxy via `nginx` instead of directly to `plackup`

ehuelsmann:
Proxying via nginx results in a stable proxy setup, whereas
via plackup, the proxy engine gets stuck -- presumably on the
fact that plackup only handles a single connection.


ehuelsmann:
Already in webpack.config.js??

PR #6700, opened 2022-08-07
Clean up use of tables

ehuelsmann:
* The `projects` table hasn't existed for a long time
* The `translation` table is an abstract table not to be used directly.


ehuelsmann:
Wrong branch...

PR #6714, opened 2022-08-09
Show all applicable tax accounts in invoices, transactions and orders

### LINKED ISSUE #2322: Sales tax on posted invoice shows incorrectly if taxable status for customer is changed
Related to #2321, if you have a customer with a sales tax checkbox unchecked, create and post an invoice for that customer, the invoice does not show any tax. If you then check the sales tax box for that customer and go back and view the invoice, you see tax listed, even though the tax is not actually posted in the invoice. It does appear correctly (as posted) on reports.

ehuelsmann:
Applicable is defined as the union between the taxes currently marked
applicable for the customer and the taxes used in the transaction.
Before this commit, the latter category was silently dropped.

Closes #2322.

### DIFF
--- a/old/bin/aa.pl
+++ b/old/bin/aa.pl
@@ -53,6 +53,8 @@ use LedgerSMB::Tax;
 use LedgerSMB::DBObject::Draft;
 use LedgerSMB::DBObject::TransTemplate;
 
+use List::Util qw(uniq);
+
 # any custom scripts for this one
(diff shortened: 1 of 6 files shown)


ehuelsmann:
From @freelock's comments, on the chat, I found out that the original issue was a one-off situation, meaning that review isn't *that* important or high priority.

PR #6715, opened 2022-08-09
Transaction reversal tracking

### LINKED ISSUE #2321: VOID transaction does not reverse actual invoice
When using the VOID button in 1.4, I expect to have the original transaction reversed. However, tax lines are not reversed but calculated fresh. For an invoice that did not have sales tax applied (because the customer did not have the tax box checked) I set the tax, then used Copy To New to get the sales tax, then voided the original transaction. The VOID transaction subtracted the tax amount in addition to the base amount, even though the original invoice did not have sales tax.

ehuelsmann:
The reversal of the reversed transaction is to be *exact* (which is what ~~#2321~~ says is currently *not* happening). Part of this has been fixed earlier, but taxes are still being redone on reversal (for invoices), which means that the transactions cannot be *exact* before this PR.

Additionally, in order to be able to track what happened to transactions, it's important to see relations between transactions as they are created. To that purpose, this PR creates infrastructure to track which transactions or invoices have been reversed (voided) and which transactions are the reversing(voiding) ones. It also prevents a transaction from being voided twice.

The following functionalities are planned for this PR:

| Item                                        | AR                 | AP                 | GL                      | IS                 | IR                 |
|---------------------------------------------|--------------------|--------------------|-------------------------|--------------------|--------------------|
| Tracking reversals  [^2]                    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:      |  :heavy_check_mark:   |  :heavy_check_mark:  |
| Prevent duplicate reversals (shared)        | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:      | :heavy_check_mark: | :heavy_check_mark: |
| Read-only UI on reversals                   | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:      | :heavy_check_mark: | :heavy_check_mark: |
| Editable headers only (reference/invnumber) |:heavy_check_mark: | :heavy_check_mark:| :heavy_check_mark:      |  :heavy_check_mark: | :heavy_check_mark:   |
| Editable headers only (sequence)            | :heavy_check_mark: | :heavy_check_mark:  | :heavy_check_mark:      | :heavy_check_mark: | :heavy_check_mark:  |
| Editable headers only (date[s])             | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:      |  :heavy_check_mark: | :heavy_check_mark:  |
| Editable headers only (description)         | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:      |  :heavy_check_mark: | :heavy_check_mark:   |
| Editable headers only (notes)               | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:      |   :heavy_check_mark: | :heavy_check_mark:   |
| Editable headers only (internal notes)      | :heavy_check_mark: | :heavy_check_mark: | :heavy_minus_sign: [^1] | :heavy_check_mark:      | :heavy_check_mark:     |

Additional functionality implemented in this PR:
* Searching for reversed and voided GL/AR/AP transactions and AR/AP invoices

This change adds tracking to transactions like this:

[^1]: Not applicable
[^2]: A problem is that the triggering workflow advances to the next state (reversed/voided) even when the replacement transaction hasn't been posted yet. We need a mechanism to advance the transaction to VOIDED / REVERSED at the same time the reversal is posted. Until that happens, it'd be nice to help the user by listing the transaction as "in the process of being reversed.


ehuelsmann:
Since the fix for #2321 has been released for 1.7, 1.8 and 1.9, the 'exact reversal' goal of this PR has been resolved. This PR was mostly initiated to address that problem (and improve on the situation of reversal tracking while at it).

old/bin/aa.pl
```diff
@@ -156,12 +156,81 @@ sub edit {
...
+            local $form->{notes};
+            local $form->{description};
+            local $form->{transdate};
+            local $form->{reference};
+            local $form->{approved};
+
+            &create_links; # create_links overwrites 'reversing'
+        };
+
+        AA->post_transaction( \%myconfig, \%$form );
```
ehuelsmann, line 213: The unfortunate side-effect of this choice is that the transaction goes through all the logic of loading and posting a transaction instead of copying the original transaction and negating its effect. I'm looking for ideas on how to bypass the loading and saving to prevent "damage" to the exact (negated) copy.

old/bin/aa.pl
```diff
@@ -157,12 +157,81 @@ sub edit {
...
+    }
+    else {
+        my $query = <<~'QUERY';
+        UPDATE gl
+           SET reference = ?,
+               description = ?,
+               transdate = ?,
+               notes = ?
+         WHERE id = ?
+        QUERY
```
ehuelsmann, line 230: What's with this query? It's not being used?!

ehuelsmann:
@christian-eriksson, @neilt, I can't add you to the list of reviewers, but your comments regarding this PR are highly appreciated!

ehuelsmann:
@neilt notes:

ehuelsmann: I tested GL and the only anomaly I found was this:

1. Enter Journal Entry
2. Do Transaction Approval
3. Search for Transaction
4. Reverse Transaction
5. Done

Why when I have "Separate Duties" turned on do I have to approve the original GL Journal Entry, but the reversal happens immediately and does not have to be Approved?

ehuelsmann:
> @neilt notes:
> 
> ehuelsmann: I tested GL and the only anomaly I found was this:
> 
>     1. Enter Journal Entry
> 
>     2. Do Transaction Approval
> 
>     3. Search for Transaction
> 
>     4. Reverse Transaction
> 
>     5. Done
> 
> 
> Why when I have "Separate Duties" turned on do I have to approve the original GL Journal Entry, but the reversal happens immediately and does not have to be Approved?

I'm unable to reproduce. Hoping that means it's no longer a problem... Can't fix it without reproduction recipe, is my main point though.

ehuelsmann:
> Not sure which lines it goes on (status or include in report). But I think being able to pull out recent voided invoices would be useful.

"Include in" is an output column with no influence on the selection. Your description suggests it's about filtering, which means it should go in the status column.

ehuelsmann:
> When I try to reverse a AR Invoice with both a service and a part I get the following error.

Hmm. I'm unable to reproduce. Neither on master nor on the branch... I guess I'll need help or we need to declare it a heisenbug.

neilt:
I'm fine with closing this PR.  I'll open an issue if I can recreate the issue.  I don't time to test it again right now.

PR #6716, opened 2022-08-10
Jest openapi

ylavoie:
With the advent of the API, the need to validate its proper behavior became urgent.
This PR introduces the infrastructure to test compliancy of the API endpoints and responses against the API definition.
Currently, the following are tested for creation, updates, retrieval and delete of:
- Business Types
- SIC
- GIFI
- Languages
- Product Pricegroups
- Warehouses


Makefile
```diff
@@ -169,6 +171,31 @@ else
             $(TESTS)
 endif
 
+jstest: TESTS ?= UI/tests
```
ehuelsmann, line 174: I understand why you would do this, but how then do we prevent the tests from ending up on production deployments? Also' everything in UI is accessible from outside the server. I'd prefer to put the tests elsewhere so it's easier for packagers and the like to exclude tests from installation.

.github/workflows/main.yml
```diff
@@ -271,6 +288,18 @@ jobs:
...
+        with:
+          name: node_modules
+
+      - name: 'Untar node modules'
+        run: tar -xf node_modules.tar
+
+      - name: Run API tests
+        run: |
+          PWD=$HOME
+          make jstest
```
ehuelsmann, line 301: I'm not too keen on having this as a separate step' we integrated everything into a single invocation of prove. You explained how jest can produce tap too, so i was expecting you to integrate with prove, at least for CI purposes.

Makefile
```diff
@@ -158,7 +159,8 @@ ifneq ($(origin DOCKER_CMD),undefined)
 	$(DOCKER_CMD) make devtest TESTS="$(TESTS)"
 else
 #        the 'dropdb' command may fail, hence the prefix minus-sign
-	-dropdb lsmb_test
+	-perl -Ilib bin/ledgersmb-admin destroy \
```
ehuelsmann, line 162: But dropdb works everywhere except on coverage testing CI. (And it never worked not was even required on CI because the database doesn't exist yet). I like the simplicity of dropdb.

Makefile
```diff
@@ -169,6 +171,31 @@ else
             $(TESTS)
 endif
 
+jstest: TESTS ?= UI/tests
+jstest:
```
ehuelsmann, line 175: If the tests get integrated into prove, you can trigger the required subset with arguments to the price invocation.

ehuelsmann (changes requested):
Very nice PR. Highly needed and good finds regarding the API issues!

Makefile
ylavoie: Jest Tests sources can be anywhere, so we can remove them from UI

Makefile
ylavoie: Fixed, the API tests will use another sample database

.github/workflows/main.yml
ylavoie: `yath` & `prove` do not have a clue on how to handle `Jest` test files.
We would need to develop another plugin for them to do so.

ehuelsmann: I thought to remember that at least prove would be able to run tests in executables. However, i cannot find any info on that now.

ehuelsmann (approved):

PR #6721, opened 2022-08-11
Restore translation

ylavoie:
Restore translation silently broken by #6711

### DIFF
--- a/UI/src/i18n.js
+++ b/UI/src/i18n.js
@@ -5,7 +5,6 @@
 export const SUPPORT_LOCALES = __SUPPORTED_LOCALES;
 const rtlDetect = require("rtl-detect");
 
-import { nextTick } from "vue";
 import { createI18n } from "vue-i18n";
 
 function _mapLocale(locale) {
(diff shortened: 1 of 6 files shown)


UI/src/i18n.js
```diff
@@ -16,10 +16,56 @@ function _mapLocale(locale) {
...
+import nl from "@/locales/nl.json";
+import pl from "@/locales/pl.json";
+import pt_BR from "@/locales/pt_BR.json";
+import pt from "@/locales/pt.json";
+import ru from "@/locales/ru.json";
+import sv from "@/locales/sv.json";
+import tr from "@/locales/tr.json";
+import uk from "@/locales/uk.json";
+import zh_CN from "@/locales/zh_CN.json";
+import zh_TW from "@/locales/zh_TW.json";
```
ehuelsmann, line 68: this bit is rather ugly indeed. However, since it works and using the dynamic variant doesn´t... I don´t see how we can solve this any better way...

ehuelsmann (approved):

PR #6690, opened 2022-08-03
Separate save/submit concerns for recon reports

ehuelsmann:
This PR also fixes a warning caused by operator precedence mismatches (and actually, that's what this was about initially).

PR #6745, opened 2022-08-29
Fix Reconciliation submit signature

ylavoie:
Completes #6690 changes.
Fixes 1.10 upgradeabily issue.


sql/changes/LOADORDER
```diff
@@ -166,4 +166,5 @@ mc/delete-migration-validation-data.sql
 1.10/gifi-config.sql
 1.10/reconciliation-columns-cleanup.sql
 1.10/unique-transaction-reconciliation.sql
+1.10/reconciliation-submit.sql
```
ehuelsmann, line 169: I think this fix belongs in `sql/modules/Reconcilation.sql`, similar to https://github.com/ledgersmb/LedgerSMB/blob/master/sql/modules/Payment.sql#L98-L102

ylavoie: Fixed.

ehuelsmann (approved):

PR #6767, opened 2022-09-15
Revert "Change missing 'security' in API to be a warning"

ylavoie:
Reverts ledgersmb/LedgerSMB#6763


ylavoie:
Reverting through #6765

PR #6771, opened 2022-09-16
Fix If-Match warning

ylavoie:
Fix If-Match warning


ehuelsmann (approved):
Looks like what I *intended* (but apparently did not write up exactly)

PR #6815, opened 2022-10-03
Fix ledgersmb-admin user

ylavoie:
- Normalize options
- Reports all errors when possible
- Allow specifying country using name or abbreviation
- Complete documentation

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -3,6 +3,7 @@
 
 requires 'perl', '5.32.0';
 
+requires 'Array::PrintCols';
 requires 'Archive::Zip';
 recommends 'Authen::SASL';
 requires 'Beam::Wire';
--- a/lib/LedgerSMB/Admin/Command/user.pm
+++ b/lib/LedgerSMB/Admin/Command/user.pm
@@ -295,8 +295,10 @@ sub _create_employee {
     my $dbh = $args{dbh};
 
     # Validate
-    my $salutation_id = $self->_get_valid_salutation($dbh);
-    return undef if !$salutation_id;
+    my $salutation_id;
+    $salutation_id = $self->_get_valid_salutation($dbh)
+        if $self->options->{salutation};
+    return undef if $self->options->{salutation} && !$salutation_id;
 
     $self->logger->error('Missing country')
         if !$self->options->{country};
--- a/lib/LedgerSMB/Admin/Command/user.pm
+++ b/lib/LedgerSMB/Admin/Command/user.pm
@@ -19,37 +19,44 @@ use LedgerSMB::Entity::User;
 use LedgerSMB::PGDate;
 use LedgerSMB::User;
 
+use Array::PrintCols;
+
 use Moose;
 extends 'LedgerSMB::Admin::Command';
 use namespace::autoclean;
 
 use Feature::Compat::Try;
 
-
 has options => (is => 'ro', default => sub { {} });
 
 sub _get_valid_salutation {
     my ($self, $dbh) = @_;
-    my $sth = $dbh->prepare('SELECT * FROM salutation');
-    $sth->execute
+    my $sth = $dbh->prepare('SELECT id FROM salutation WHERE salutation ilike ?');
+    $sth->execute($self->options->{salutation})
         or die $dbh->errstr;
-    my ($values) = $sth->fetchall_arrayref({});
-    foreach (@$values) {
-        return $_->{id}
-            if ( $_->{salutation} eq $self->options->{salutation})
-    }
-    $self->logger->error('Invalid salutation');
-    return 0;
+    my $value = $sth->fetchrow_hashref;
+    $self->logger->error('Invalid salutation')
+        if !$value;
+    return $value->{id};
+}
+
+sub _get_salutation_by_id {
+    my ($self, $dbh, $id) = @_;
+    my $sth = $dbh->prepare('SELECT salutation FROM salutation WHERE id=?');
+    $sth->execute($id)
+        or die $dbh->errstr;
+    my $values = $sth->fetchrow_hashref;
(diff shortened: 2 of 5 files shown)


lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -444,10 +450,10 @@ __END__
 
 =head1 SYNOPSIS
 
-   ledgersmb-admin user list <db-uri>
-   ledgersmb-admin user create <db-uri> [options] 
-   ledgersmb-admin user delete <db-uri>
-   ledgersmb-admin user change <db-uri>
+   ledgersmb-admin user list <db-uri> <user>
```
ehuelsmann, line 453: Where does the `<user>` get parsed from the arguments? Is it stored in a variable somewhere?

ylavoie: `user list` doesn't use a user tag, it lists all users. Fixed

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -445,9 +451,9 @@ __END__
 =head1 SYNOPSIS
 
    ledgersmb-admin user list <db-uri>
-   ledgersmb-admin user create <db-uri> [options] 
-   ledgersmb-admin user delete <db-uri>
-   ledgersmb-admin user change <db-uri>
+   ledgersmb-admin user create <db-uri> <user> [options]
```
ehuelsmann, line 454: Where is the `<user>` parsed from the command line arguments (and presumably stored in a variable)?

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -475,11 +481,11 @@ User name of the created or changed user
 
 Password
 
-=item start_date
+=item start-date
```
ehuelsmann, line 484: Takes a date as an argument. But which format?

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -507,23 +513,23 @@ Employee manager
 
 Employee number
 
-=item is_manager
+=item is-manager
```
ehuelsmann, line 516: Doesn't take an argument, but how do I know?

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -112,20 +115,20 @@ sub _option_spec {
             # employee
             'dob=s' => \$dob,
             'employeenumber=s' => \$employeenumber,
-            'end_date=s' => \$end_date,
-            'is_manager!' => \$is_manager,
+            'end-date=s' => \$end_date,
+            'is-manager!' => \$is_manager,
             'manager=s' => \$manager,
             'role=s' => \$role,
```
ehuelsmann, line 121: How does the user know which values can be specified for the "role" argument?

ehuelsmann: Ok. Based on the code, I figured it out. Given that you allow individual roles to be specified, don't you think this option is useless unless we have an option to query available roles?

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -112,20 +115,20 @@ sub _option_spec {
...
             'sales!' => \$is_sales,
```
ehuelsmann, line 122: This is rather inconsistent: declaring a sales works with `--sales` but declaring a manager is `--is-manager` ?!

lib/LedgerSMB/Admin/Command/user.pm
```diff
@@ -112,20 +115,20 @@ sub _option_spec {
             # employee
```
ehuelsmann, line 115: Where is `permission` applied? Or if it's applied, where is `role` applied?!

ehuelsmann (changes requested):
There's more work to be done here: we need a list of available permissions if we're going to ask the names of the permissions as arguments of the `--permission` argument. Also, I've submitted a PR against this branch for further textual and consistency improvements.

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Fixed in documentation

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Fixed in documentation

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Added `user list a_user` to query current permissions,

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Fixed

cpanfile
```diff
@@ -3,6 +3,7 @@
 
 requires 'perl', '5.32.0';
 
+requires 'Array::PrintCols';
```
ehuelsmann, line 6: Hmm. Why do we need this dependency? I know they are considered archaic, but Peel has formats which are useable to print table data. It's what i used to print the data about the available templates. They still work, although they lack flexibility with respect to the number of printed columns and their width.

ylavoie: We have tons of roles. Printing them vertically sorted is more usual than horizontally.

lib/LedgerSMB/Admin/Command/user.pm
ylavoie: Cleaned confusion between user role (Job title) and postgres roles (permissions)

ehuelsmann (approved):

PR #6819, opened 2022-10-05
Webservices

ehuelsmann:


freelock (approved):
I see a github action calling out an extra "my" declaration, but aside from that, I don't see anything jumping out at me on this... if it's the same basic code I've been running, I'd love to see this go in!

ehuelsmann:
> I see a github action calling out an extra "my" declaration, but aside from that, I don't see anything jumping out at me on this... if it's the same basic code I've been running, I'd love to see this go in!

Yep. It's the same thing, except that this version is cleaned up for some boiler plate that has been abstracted out for the other APIs already. Additionally, this version comes with OpenAPI documentation, which is used to test request *and* replies against.

ylavoie (approved):
We'll need to implement the OpenAPI tests to complete validation of this PR

Issue #6787, opened 2022-09-25
Can't delete draft transactions with attachments


ehuelsmann:
### Version

1.9

### What browsers are you seeing the problem on?

This problem isn't browser related

### What happened?

Error popup:

![image](https://user-images.githubusercontent.com/2326559/192156653-a764e80a-c513-425a-a32d-72c280f8cf11.png)

### What should have happened?

Draft should have been deleted...

PR #6824, opened 2022-10-07
Can't save draft transactions with attachments (#6787)

datajunkie-eng:
Adds cascading deletes to the file_transaction@ref_key => transaction foreign key

### DIFF
--- a/sql/changes/1.10/file-tranaction-cascade-delete.sql
+++ b/sql/changes/1.10/file-tranaction-cascade-delete.sql
@@ -1,9 +1,9 @@
-
 -- adding cascade deletes
-alter table file_transaction 
-	drop constraint file_transaction_ref_key_fkey;
+alter table file_transaction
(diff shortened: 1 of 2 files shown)


ehuelsmann:
Thanks for this PR! Yes, it's exactly what's required.

The tests fail, though: there are some checks for correct formatting for these files:
* No trailing whitespace
* No TAB characters (use spaces instead)

The last point mainly exists so the file is rendered exactly the same way on everybody's editor -- irrespective of TAB width settings.

ehuelsmann (changes requested):
Could you please replace the TAB characters with spaces and remove the trailing spaces?

ehuelsmann (approved):

ehuelsmann:
@datajunkie-eng thanks for your contribution! Could you cherry pick the commit into a branch of `master` so it will be in 1.11+ too?

PR #6871, opened 2022-10-15
Add STABLE keyword to functions containing only SELECT command to improve query performance

aungzwin10:


einhverfr:
I think it is good practice to add the STABLE keyword in most cases. Were you able to benchmark a difference in performance? I am asking because my mental model of what this keyword does indicates it should not have any impact on the plan generated in the cases where the application is running the function from the front-end (say, via PGObject), but would likely improve things in some other cases (for example, some LATERAL joins against the function output).

So for an example if why this is a niche improvement that shouldn't provide gains for LedgerSMB:

```
stability_test=# create function stable_random() returns float language sql stable as 'select random()'
;
CREATE FUNCTION
stability_test=# create function immutable_random() returns float language sql immutable as 'select random()'
;
CREATE FUNCTION
stability_test=# select g, random(), stable_random(), (select stable_random()) as stable_random_in_subselect, immutable_random() from generate_series(1, 10) g;
 g  |       random        |    stable_random    | stable_random_in_subselect |  immutable_random  
----+---------------------+---------------------+----------------------------+--------------------
  1 | 0.17980929012250613 |  0.8229691481019579 |       0.046210618367986456 | 0.6518413450347715
  2 | 0.09393382982394272 |  0.5665446119477195 |       0.046210618367986456 | 0.6518413450347715
  3 |  0.9557985605362909 |   0.929482135396448 |       0.046210618367986456 | 0.6518413450347715
  4 |   0.589420333530402 |  0.6951802972772612 |       0.046210618367986456 | 0.6518413450347715
  5 |  0.8376368853156855 |  0.5701574767561866 |       0.046210618367986456 | 0.6518413450347715
  6 |  0.9664047997838345 | 0.47701014098142025 |       0.046210618367986456 | 0.6518413450347715
  7 |  0.2811356802188776 | 0.35943575057671495 |       0.046210618367986456 | 0.6518413450347715
  8 | 0.12664484905207374 | 0.43100783311402324 |       0.046210618367986456 | 0.6518413450347715
  9 | 0.44446695637754985 | 0.34134981505156503 |       0.046210618367986456 | 0.6518413450347715
 10 |  0.6109326545936433 |  0.3364650077773348 |       0.046210618367986456 | 0.6518413450347715
(10 rows)
```
you can see here that giving the random function an immutable status makes it return the same for each row.  For the stable version, it still gets run on each row unless it is in a correlated subquery, in which case it gets optimized away.

So there are probably some cases where manual reporting queries might be sped up by this improvement and it is a good practice to make the guaranteed read-only functions stable. But I doubt this will improve performance given that these are usually single function calls per snapshot and therefore cannot be optimized away.

einhverfr:
Also for the record and for any further reviews:  there are three "stability" levels to PostgreSQL functions: immutable, stable, and volatile (the default).  With volatile functions, the function either has a side effect or changes output on every run, and therefore needs to be run every time it is called.  Stable functions have no side effects and return the same data for a given database snapshot, so there are certain cases where these calls can be optimized away and only called once in a query.  Immutable functions guarantee their output doesn't change even across database snapshots, and so you can index their output etc.

ehuelsmann (approved):

ehuelsmann:
Merging, based on https://github.com/ledgersmb/LedgerSMB/pull/6873#issuecomment-1279722947

PR #6873, opened 2022-10-15
1.9 - Add STABLE keyword to functions containing only SELECT command to improve query performance

aungzwin10:


einhverfr:
While I am in favor of applying this to master, I am opposed to applying to stable branches.  Because of how we call the functions, they are always going to run once per call.  In other words we are never likely to break things or suffer performance problems due to having these functions at the wrong stability level.    On the other hand, if we change them in a stable release and get them wrong, there is a slim (but significantly greater) chance of breaking someone's external queries.  In other words, the chance of this fixing something in LedgerSMB is very nearly zero, and the chance of breaking something external without a proper beta and feedback period is somewhat higher.

einhverfr:
After further discussions something weird is happening here performance-wise, and this appears to fix it, so I remove my objection.

ehuelsmann (approved):

PR #6912, opened 2022-10-28
Clean linter warnings about @import taking a URL

ehuelsmann:


ylavoie (commented):
I was planning to do it tonight. ;-)

PR #6920, opened 2022-10-29
Render approved transactions as read-only forms

ehuelsmann:
This commit renders all invoice and transaction inputs except the 'Save Info'-saveable fields as read-only in order to remove the impression that the transaction could be editable.

To further remove that impression, it also prevents rendering of the empty 'new input' line on the invoice and payments sections.

### DIFF
--- a/UI/journal/journal_entry.html
+++ b/UI/journal/journal_entry.html
@@ -153,14 +153,15 @@
                         [% IF displayrow.accnoset == 0 %]
                                 [% PROCESS input element_data = {
                                       name = "accno_$INDEX"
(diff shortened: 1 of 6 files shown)


neilt:
The idea to not show non-editable fields as editable is conceptually correct and hopefully will be applied, where possible, across other LedgerSMB views.

Did not a review the code.

neilt (approved):
The idea to not show non-editable fields as editable is conceptually correct and hopefully will be applied, where possible, across other LedgerSMB views.

Did not a review the code.

ylavoie (approved):

PR #6927, opened 2022-10-29
Fix action triggered by delete button

ehuelsmann:


ylavoie (dismissed):
Using an old config for jest tests with leak detection.
https://github.com/ehuelsmann/LedgerSMB/blob/fix/master/delete-button/jest.config.js#L52

ehuelsmann:
Ok, but so does master?! https://github.com/ledgersmb/LedgerSMB/blob/master/jest.config.js#L52

ehuelsmann:
Review comments fixed in a separate PR (by @ylavoie).

PR #6928, opened 2022-10-30
Sort CSS files

ylavoie:
Sort CSS files for easier selection


ehuelsmann:
I'd rather see a case insensitive sort here. That easy, the user experience will simply be "alphabetical order"

PR #7017, opened 2022-12-20
Add the tables to Statistics Report after a database upgrade in setup.pl

neilt:
Add the tables parts, orderitems, and invoice to the tables stats reported after a database upgrade in setup.pl. Put the tables in alpha order.


ehuelsmann:
Looking good! Thanks for the contribution!

PR #7055, opened 2023-01-02
Add stringification to PGDate instead of the one inherited

ylavoie:
Make sure that PGDate handles the output to handle undefined, date or time only and uses user desired format.

### DIFF
--- a/lib/LedgerSMB/PGDate.pm
+++ b/lib/LedgerSMB/PGDate.pm
@@ -252,6 +252,11 @@ sub _formatter {
 # table (being GL>Search-ed without restrictions)
 memoize('_formatter');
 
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/PGDate.pm
```diff
@@ -252,6 +252,11 @@ sub _formatter {
 # table (being GL>Search-ed without restrictions)
 memoize('_formatter');
 
+# Provide a stringification routine instead of the one inherited
+use overload (
+    q{""}    => 'to_output',
```
ehuelsmann, line 257: I think you need `fallback => ` there:

https://perldoc.perl.org/overload#fallback

ehuelsmann:
Thanks! Looks good now!

PR #7072, opened 2023-01-09
Fix SQL function slow compared to plain query

aungzwin10:
On a large database, the functions with language SQL take much longer than plain query because of the query planner take too long to prepare statement.

Previously, the performance issue was solved by adding STABLE keyword to the function but after some amount of times, the issue happen again. This commit fix the performance issue by changing the SQL language to dynamic SQL with PLPGSQL language.


ehuelsmann:
@aungzwin10, there are lots of test failures. You can find the screenshots of the failing steps in the zip downloads here: https://github.com/ledgersmb/LedgerSMB/actions/runs/3873742878

ehuelsmann (approved):
Wow. This is a major effort! Thanks for going through it.

PR #7101, opened 2023-01-21
Fix rounding crash in reports

ehuelsmann:
This fix solves the rounding crash, caused by "money" data in the row hashes that isn't formatted by the report, because the columns aren't included in the report.

Next to that, it eliminates double traversal of the row data while encoding the data for HTML: once via the 'report' variable and once via the 'rows' variable. The latter traversal meant full double encoding of the data (because the data is copied during traversal).

Reported by @aungzwin10 on the Matrix chat channel.


ehuelsmann:
Too risky to merge with these failures.

PR #7105, opened 2023-01-22
Asset sale should take cash into account

### LINKED ISSUE #5180: Unbalance disposal transaction
According to our chat, we have unbalance transaction when we do disposal for the asset.
There should be another account transaction with the amount 350,000, which is the actual sale amount of asset.

image

ehuelsmann:
This commit adds a Cash Account to asset disposal reports in order to account for the cash component of the sale. Please do note that if an invoice was created for the sale, there's some additional accounting to be done to move the income out of the PNL, into the "cash" account.

There's a bigger plan for rewriting fixed assets support. This commit is part of the "old world", not the start of development of the new, shiny and highly desired code base.

Fixes #5180.

### DIFF
--- a/sql/modules/Assets.sql
+++ b/sql/modules/Assets.sql
@@ -888,7 +888,7 @@ $$ language sql;
 CREATE OR REPLACE FUNCTION asset_report__get_cash_accts()
 RETURNS SETOF account
 AS $$
-  SELECT * FROM account;
+  SELECT * FROM account where not obsolete;
 $$ language sql;
 
 COMMENT ON FUNCTION asset_report__get_loss_accts() IS
--- a/sql/modules/Assets.sql
+++ b/sql/modules/Assets.sql
@@ -1,6 +1,31 @@
 
 set client_min_messages = 'warning';
 
+/*
+
+
+  asset_report__generate_gl   - creates GL transaction upon approval of depreciation report
+  asset_report__disposal_gl   - creates GL transaction upon approval of a disposal (sale/abandonment)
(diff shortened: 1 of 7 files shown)


ylavoie (approved):

sql/modules/Assets.sql
```diff
@@ -843,6 +885,12 @@ AS $$
     SELECT * FROM account__get_by_link_desc('asset_loss');
 $$ language sql;
 
+CREATE OR REPLACE FUNCTION asset_report__get_cash_accts()
```
aungzwin10, line 888: 1. Should exclude obsolete account.
2. Instead of getting all accounts, better to filter only `AR_paid`, `AP_paid`?
Is there any other use case which need all accounts?
(Because, we used only those account when we manually add cash account in Journal line)

aungzwin10 (changes requested):

sql/modules/Assets.sql
ehuelsmann: Thanks! The obsolete account fix was added.

The lack of selection of accounts based on "AR_paid" or "AP_paid" was intentional: these are intended to select the drop-down values on payments from vendors and to customers. However, for fixed assets you may want to use other accounts (e.g. suspense accounts) which don't apply to payments/receipts. When I created this,  I estimated that the added complexity of yet another "knob" (checkbox on the account configuration screen) is not warranted given the fact that fixed assets disposal will be far less common than processing payments -- meaning that added flexibility of the selection of the right account in the drop-down outweighs the productivity lost by needing to scroll to the right account. Agreed?

aungzwin10 (approved):

PR #7116, opened 2023-01-25
Fix Contact > Addresses mismatch

aungzwin10:
- Fix Contact > Addresses doesn't render correctly after moving from TableContainer to Grid system
- Add gap and margin to the grid system to render more elegantly
- Remove inline grid-gap style from importCSV-Base for two reasons:
  1. gap property is defined in CSS, no inline style need
  2. grid-gap is obsolete and replaced by 'gap' property

### DIFF
--- a/UI/Contact/divs/address.html
+++ b/UI/Contact/divs/address.html
@@ -91,9 +91,9 @@
                 value= DISPLAY.id
         };
 %]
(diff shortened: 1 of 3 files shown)


ehuelsmann:
Nice! Please backport to 1.10!

PR #7125, opened 2023-01-26
Reverting #7116 changes for contact > addresses

aungzwin10:
Reverting #7116 changes for contact > addresses because fields are no longer aligned

The fields are mismatched only when entity doesn't have credit account. It just need to add a new element(in this case <span>) on the condition where the entity doesn't have credit account to retain the two column grid.

The current changes will look like this.
Screenshot 2023-01-26 at 10 02 13 AM

### DIFF
--- a/UI/Contact/divs/address.html
+++ b/UI/Contact/divs/address.html
@@ -91,9 +91,9 @@
                 value= DISPLAY.id
         };
 %]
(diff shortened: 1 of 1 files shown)


ehuelsmann (approved):
Yep. Looking good now. Thanks!

PR #7129, opened 2023-01-27
Follow up 7072 - Fix type incompatibility between query and return type

aungzwin10:


ehuelsmann:
I think this is a duplicate of #7124.

aungzwin10:
@ehuelsmann #7124 is for outstanding function
This PR is for aging related.

PR #7130, opened 2023-01-27
Add a payments subtotal and remaining balance on invoices

ehuelsmann:
After this change, the invoice screens (both sales and purchase) will look like this:

(Note that the CSS to set the font characteristics is not included.)

Added the same for transaction screens, which now look like:


aungzwin10 (approved):

neilt:
Looks like a good improvement.  But 59 files changed? Wow.

PR #7152, opened 2023-01-29
Mention user name and timestamp in invoice/order history

### LINKED ISSUE #5023: Post new invoice by API
Jotting down from chat:

Need a better way to post an invoice from another system. Current integrations use the web interface, but have to do multiple posts to build up an invoice before it can be posted. Instead, have a way to post an invoice document -- json or xml -- in a single go.

Opening this issue to discuss the shape of the payload, the API endpoint to use, requirements, etc.

In my use case, I do need this to be able to support some edge cases:

- Payload document can specify prices that vary from the product price in the LSMB database
- Payload document can specify a manual tax rate to apply to taxable products (because we have destination-based tax rules that result in thousands of possible rates for a particular tax, and our external system already calculates this correctly, while LSMB does not).

For the first cut, I personally do not need the API to populate many of the necessary references -- the external system can supply inside the payload:
- eca id (or meta number)
- product id
- tax id, rate, total amount

Thoughts?

ehuelsmann:
Closes #5023

### DIFF
--- a/old/bin/ir.pl
+++ b/old/bin/ir.pl
@@ -779,43 +779,11 @@ qq|<textarea data-dojo-type="dijit/form/Textarea" id=intnotes name=intnotes rows
   </tr>
   <tr>
     <td>
-      <table width=100%>
-         <caption>History</caption>
 |;
-    # insert history items
-    my ($wf_id) =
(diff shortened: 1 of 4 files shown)


ehuelsmann:
After this change, the history looks like:

PR #7176, opened 2023-02-02
Adjust build-dev trigger

ylavoie:
Fix build-dev trigger to build dev images when needed

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -374,8 +374,7 @@ jobs:
         shell: bash
         run: |
           URL="https://api.github.com/repos/${{ github.repository }}/commits/${{ github.sha }}"
-          FILES=$(curl -s -X GET -G $URL | jq -r '.files[].filename')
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -374,8 +374,7 @@ jobs:
         shell: bash
         run: |
           URL="https://api.github.com/repos/${{ github.repository }}/commits/${{ github.sha }}"
-          FILES=$(curl -s -X GET -G $URL | jq -r '.files[].filename')
-          if echo $FILES | grep -q "cpanfile"; then
+          if $(curl -s -X GET -G $URL | jq -r '.files[].filename' | grep -q "cpanfile"); then
```
ehuelsmann, line 377: is this real shell code? If so, shouldn't that be something like:

```bash
if [ -n $(...) ] ; then
```

?

ylavoie: Indeed

ehuelsmann:
Please rebase: there are too many commits in the list for the PR

PR #7199, opened 2023-02-07
Handle form state with a state machine

ehuelsmann:
Remaining work:

* [x] Add errors returned by the server upon CSV upload to the popup-Toast
* [x] Determine the next-state upon CSV processing (server-side); choices so far are
   a. Reset the form to empty and start from scratch
   b. Show a greyed-out button with nowhere to go
   c. Clean the upload file field, but leave others in place for re-use; otherwise like (a)
  d. Show a greyed-out button which becomes active on the first form activity (e.g. changing the file or tabbing to another input field)

Went with (d) for now.

### DIFF
--- a/UI/src/views/LoginPage.machines.js
+++ b/UI/src/views/LoginPage.machines.js
@@ -11,21 +11,13 @@ import {
     state,
     transition
 } from "@/robot-vue";
-const registry = require("dijit/registry");
-
-const not = (fn) => (
-    (...args) => !fn(...args)
-);
-const formValid =
-      (ctx) => registry.findWidgets(ctx.form.value).reduce(
-            (acc, w) => (acc && (w.disabled || !w.validate || w.validate())),
-            true
-        );
-const okResponse = (ctx) => ctx.response.ok;
-const responseStatus = (status) => (
-    (ctx) => (ctx.response.status === status)
-);
-
+import {
+    testNot,
+    testResponseOkFn,
+    testResponseStatusFn,
+    transitionFormValid,
+    transitionFormInvalid
+} from "@/machine-helpers.js";
 
 function submitLogin(ctx) {
     return fetch("login.pl?action=authenticate&company=" + encodeURI(ctx.company.value), {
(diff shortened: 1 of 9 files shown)


ylavoie (commented):
I like it very much, no comments on the code.
We need to adjust the tests though

ehuelsmann:
@ylavoie since your review, I've addressed my gripe with  the fact that I was declaring the form constraints in the code, instead of *in the html*. The latter feels much more natural with Dojo as our current background.

@ylavoie there's one thing you could help me with: I'm in doubt the string translations still get correctly extracted: I wouldn't be surprised if "ctx.t" isn't being detected as the translation function. Question will be what we need to do to restore correct detection.

ylavoie:
> @ylavoie since your review, I've addressed my gripe with the fact that I was declaring the form constraints in the code, instead of _in the html_. The latter feels much more natural with Dojo as our current background.
> 
> @ylavoie there's one thing you could help me with: I'm in doubt the string translations still get correctly extracted: I wouldn't be surprised if "ctx.t" isn't being detected as the translation function. Question will be what we need to do to restore correct detection.

Your code is perfectly set but there is no translation for the error messages, so they default to the LedgerSMB configured language, usually the english version.

ehuelsmann:
Next step for this PR is too move the import form to this paradigm as well and factor out the commonality.

UI/src/views/LoginPage.machines.js
```diff
@@ -0,0 +1,119 @@
...
+    createMachine,
+    immediate,
+    interpret,
+    invoke,
+    guard,
+    reduce,
+    state,
+    transition
+} from "@/robot-vue";
+const registry = require("dijit/registry");
```
ylavoie, line 14: The LoginPage is Vue and WebElements, yet we're bringing back Dojo as a direct dependency?

ehuelsmann: Yes. I dislike it a lot. Here's the reason: each form has a concept of "being valid", but that checks the standard conditions, not the ones implemented by dojo as part of its widgets. So we can't ask the form whether it's valid or not, because it won't return the correct value.

That's where Ihad to rever to using dojo for now...

ehuelsmann: the registry requirement has now been centralized. Hope that's less upsetting.

ehuelsmann:
for the import error to be rendered readably in the toast, it needs to be returned as a JSON structure instead of an HTML page.

UI/src/views/LoginPage.machines.js
ylavoie: It only upsets Jest. I'll mock it away.

ylavoie (approved):

PR #7214, opened 2023-02-11
Wrap Dijit widgets required to move the UI away from directly depending on Dojo

ehuelsmann:
The purpose of this PR is to move parsing of the DOM away from Dojo: the browser can do it a lot more efficiently using Custom Web Components. These web components can instantiate Dojo widgets (for now) to create a consistent UI experience (which we can later swap for another type of custom element for a different look).

We make Dojo widgets out of standard elements in some places, such as [the `TABLE` element which is converted to a `lsmb/InvoiceLines` widget](https://github.com/ledgersmb/LedgerSMB/blob/29763d643e75935a71d59bd25d8500650dc018cf/old/bin/io.pl#L253). Support for this usage exists in custom web components [using the `is` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is).

There's a problem with Custom Built-in Web Components: [Safari does not support Custom Web Elements for built-in elements](https://www.sobyte.net/post/2021-08/safari-buildin-custom-element-polyfill/). There is a solution to that problem: using a [polyfill](https://github.com/WebReflection/custom-elements-builtin).

### DIFF
--- a/UI/Configuration/settings.html
+++ b/UI/Configuration/settings.html
@@ -1,6 +1,6 @@
 [% PROCESS elements.html %]
 <body class="lsmb [% dojo_theme %]">
-  <form data-dojo-type="lsmb/Form"
+  <form is="lsmb-form"
         id="system-settings"
         method="post"
         action="[% form.script %]">
@@ -102,7 +102,7 @@
               };
            END;
            PROCESS button element_data={
-                   name => 'action'
+                   name => '__action'
                    value => 'save_defaults'
                    text => text('Save')} %]
   </form>
--- a/UI/js-src/lsmb/Form.js
+++ b/UI/js-src/lsmb/Form.js
@@ -13,19 +13,9 @@ define([
     var c = 0;
     return declare("lsmb/Form", [Form], {
         clickedAction: null,
-        startup: function () {
-            var self = this;
-            this.inherited(arguments);
-
-            // <button> tags get rewritten to <input type="submit" tags...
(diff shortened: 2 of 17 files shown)


ehuelsmann:
Testing showed that it's impossible to use the dojo parser in combination with dojo-based custom components: the browser renders the widgets, but the AMD parser does it again, leading to havoc. This means we can't partially move to the use of these widgets. We either need to go "all in" or forget about it.

I'm all for moving: it means that each element maintains its own state and registration with the Dojo widget registry, eliminating the need to sync destruction of Dojo widgets with Vue state. It also opens the possibility to start moving to a heavier (and smarter) web-UI as described above.

ehuelsmann:
@ylavoie do you have any comments as to my reasoning?

Note: my idea is that eventually, we use Vue to control the UI and we use web components for the layout. Thereby I mean: checkboxes, textboxes, date selectors, drop-downs etc will be custom components. The component adding/removing lines on the invoice will eventually be a Vue component.

ehuelsmann:
Additional findings while working on this PR showed that Dojo creates new nodes in the DOM tree, even if that would not be necessary (e.g. because the tag is only being "enhanced" with additional methods). This means it's impossible for custom tags to be a dojo widget at the same time: when passing the DOM node (`this`) to Dojo, it removes the node from the DOM and puts its own tag (a standard tag) in place.

ehuelsmann:
Based on the discussion and findings so far, is clear were not there yet... Closing for now.

PR #7246, opened 2023-02-28
Fix tests and date comparison

ehuelsmann:


ehuelsmann:
This can't be backported due to missing schema support.

PR #7247, opened 2023-02-28
Revert "Fix cleared balances with overlapping reconciliations"

ehuelsmann:
Reverts ledgersmb/LedgerSMB#7245


ehuelsmann:
Removing commit which fails tests and was accidentally merged before test completion. The schema misses some fields to support this PR.

PR #7257, opened 2023-03-04
Update PG versions

ylavoie:
Explicitly fix CI PG versions to the versions we support


ehuelsmann:
CircleCI won't run for PRs which only change GitHub Actions. Merging without CCI run.

PR #7270, opened 2023-03-16
Upgrade OpenAPI

ylavoie:
Remove deprecated `redoc-cli` package in favour of `@redocly/cli` 
Deliver a very small viewer and the API as a JSON file in the doc directory instead of the previous HTML file.


utils/devel/rebuild_api.sh
```diff
@@ -29,7 +29,8 @@ cp $gitDirName/openapi/.redocly.yaml .
 npx @redocly/cli lint API.yaml || exit
 
 # Build the documentation
-redoc-cli build API.yaml -o $gitDirName/doc/openapi/LedgerSMB-api.html
```
ehuelsmann, line 32: I had expected to see an update to package.json. why isn't that the case when it says "upgrade"?

ehuelsmann (approved):

PR #7271, opened 2023-03-17
Jest views

ylavoie:
Add Jest tests for UI/src/views


ehuelsmann (approved):
Very nice work! Good to see this effort caught some actual problems and that the coverage goes back up nicely too!

PR #7310, opened 2023-04-06
Add print functionality in the Journal Entry

aungzwin10:
Add the print button in the Journal Entry after transaction is saved.
Screenshot 2023-04-06 at 5 04 57 PM

Generate the HTML template as sample below
Screenshot 2023-04-06 at 5 08 53 PM

### DIFF
--- a/UI/templates/widget.html
+++ b/UI/templates/widget.html
@@ -26,7 +26,7 @@
                data-dojo-attach-point="templateName">
             <span value=""><span>[% text("--TEMPLATE--") %]</span></span>
             <span value="print_journal_entry">
-              <span>[% text("JE transaction") %]</span></span>
+              <span>[% text("GL transaction") %]</span></span>
             <span value="print_cash_transfer">
               <span>[% text("Cash Transfer transaction") %]</span></span>
             <span value="ap_transaction">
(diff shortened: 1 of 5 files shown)


UI/templates/widget.html
```diff
@@ -25,6 +25,10 @@
                data-dojo-props="required:true,value:'[% template_name %]'"
                data-dojo-attach-point="templateName">
             <span value=""><span>[% text("--TEMPLATE--") %]</span></span>
+            <span value="print_journal_entry">
+              <span>[% text("JE transaction") %]</span></span>
```
ehuelsmann, line 29: I'm in doubt: the menu under which the print button resides is called "Journal Entry". That menu item is located within the "General Journal" menu. I'm inclined to call it a "General Journal transaction"... But elsewhere I *think* we may be calling the same thing a G/L transaction.

ehuelsmann: Ok. I figured it out. I'd like it to be called "G/L transaction".

ehuelsmann (changes requested):
One change requested.

UI/templates/widget.html
ehuelsmann: Done.

ehuelsmann (approved):

PR #7334, opened 2023-04-15
Move login.css and setup.css to system

ylavoie:
Make sure that all CSS are included by webpack

### DIFF
--- a/UI/webpack.config.js
+++ b/UI/webpack.config.js
@@ -65,6 +65,20 @@ if (TARGET !== "readme") {
             .sort();
     }
 
+    /* eslint-disable-next-line no-inner-declarations */
+    function globCssEntries(globPath) {
+        var files = glob.sync(globPath);
+        var entries = {};
+
+        for (var i = 0; i < files.length; i++) {
+            var entry = files[i];
+            const dirName = path.dirname(entry).replace(/\.\/css\/?/,"");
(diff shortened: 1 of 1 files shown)


UI/webpack.config.js
```diff
@@ -65,6 +65,20 @@ if (TARGET !== "readme") {
             .sort();
     }
 
+    /* eslint-disable-next-line no-inner-declarations */
+    function globCssEntries(globPath) {
+        var files = glob.sync(globPath);
+        var entries = {};
```
ehuelsmann, line 71: Using `let` should eliminate inner declaration warnings or errors.

ehuelsmann (approved):
Looks good. Can we backport to 1.10 and 1.9 ?

UI/webpack.config.js
```diff
@@ -67,11 +67,11 @@ if (TARGET !== "readme") {
 
     /* eslint-disable-next-line no-inner-declarations */
     function globCssEntries(globPath) {
-        var files = glob.sync(globPath);
-        var entries = {};
```
ehuelsmann: With this change, we can remove the "no-inner-declaratios" eslint directive, no?

ylavoie: No, the function is not defined at root but in the  application section, because README has no use for that.

PR #7341, opened 2023-04-15
Fix/1.9/unselectable css

ylavoie:
Make sure that all CSS are included by webpack


ehuelsmann:
This one seems to have one too many commits. the "Fix typo" commit doesn't look like it should be part of this PR.

ehuelsmann (changes requested):
As per the earlier comment.

ehuelsmann (approved):

PR #7379, opened 2023-04-30
Show reversing reference when reversal is posted

christian-eriksson:
I noticed an issue when playing around with the new reverse transaction. When a reversing transaction is created the GUI shows the reference and ID of the reversed transaction. If the user presses "save" only the ID is shown. When the user presses "save" the only reference LedgerSMB has to the reversed transaction is the transaction ID from the database. To show the reference it needs to be collected.

so instead of:
> this transaction will reverse transaction X with ID Y
it shows:
> this transaction will reverse transaction with ID Y

I haven't added any test, I'm not sure how tests are setup or if there are any test for these things. At least I don't think I broke any existing tests as they still passed when running them.

The reason why the reference show in the first step, but not in the second, seems to be two-fold:

1. The reference data is not passed to the next view when pressing "save".
2. The form data is reset once in the post subroutine (which is called on pressing "save") so the data doesn't reach the display subroutine.

When the user press "reverse" the form data needed in the next view is passed as hidden form fields in the HTML. To preserve other form data it is scoped to be local to a do{...} block. Without this it would be overwritten with data from the database and since the references to the reversing has not been saved the unscoped fields will be reset or undefined. As this happens before rendering the form the displayed form will not have the expected data.

### DIFF
--- a/old/bin/aa.pl
+++ b/old/bin/aa.pl
@@ -174,6 +174,7 @@ sub edit {
 
 sub reverse {
     $form->{title}     = $locale->text('Add');
+    $form->{reversing_reference} = $form->{invnumber};
     $form->{invnumber} .= '-VOID';
 
     &create_links;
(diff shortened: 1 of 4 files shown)


christian-eriksson:
I'm also wondering if this:
```pl
(...  $form->{reversing}) : $locale->text('This transaction will reverse transaction [_1]', $form->{reversing})) .q|</a><br />| : '') .
```
in [old/bin/aa.pl](https://github.com/ledgersmb/LedgerSMB/blob/459bebc153fcfbc84380ff7db35d515feea3305c/old/bin/aa.pl#L741) is a similar bug that I just haven't seen yet. I seems to me that that line will only show the ID of the transaction rather than the reference+transaction.

Wasn't sure how to provoke those lines though, so I couldn't test it :/

christian-eriksson:
> I'm also wondering if this:
> 
> ```perl
> (... : $locale->text('This transaction will reverse transaction [_1]', $form->{reversing})) .q|</a><br />| : '') .
> ```
> 
> in [old/bin/aa.pl](https://github.com/ledgersmb/LedgerSMB/blob/459bebc153fcfbc84380ff7db35d515feea3305c/old/bin/aa.pl#L741) is a similar bug that I just haven't seen yet. I seems to me that that line will only show the ID of the transaction rather than the reference+transaction.
> 
> Wasn't sure how to provoke those lines though, so I couldn't test it :/

If that's an issue, then its probably the same in [old/bin/ir.pl](https://github.com/ledgersmb/LedgerSMB/blob/459bebc153fcfbc84380ff7db35d515feea3305c/old/bin/ir.pl#L464) and [old/bin/is.pl](https://github.com/ledgersmb/LedgerSMB/blob/459bebc153fcfbc84380ff7db35d515feea3305c/old/bin/is.pl#L521)

ehuelsmann:
Thank you for this PR! It'll be merged once tests finish.

PR #7380, opened 2023-04-30
Add infrastructure to document roles

### LINKED ISSUE #3305: Doc base user
What info can be acquired by-- and what damage can be done by a base user? 
How can a prospective user assess LSMB security absent some hints on this?

### LINKED ISSUE #7376: It seems that one of role ap_transaction_all or ap_all_transactions is wrong
These should be consolidated into one role.

ehuelsmann:
Closes #3305
Closes #7376

### DIFF
--- a/sql/modules/Roles.sql
+++ b/sql/modules/Roles.sql
@@ -1434,12 +1434,22 @@ SELECT lsmb__grant_role('batch_list', 'gl_reports');
 SELECT lsmb__grant_perms('batch_list', obj, 'SELECT')
   FROM unnest(array['batch'::text, 'batch_class', 'voucher']) obj;
 
-SELECT lsmb__create_role('gl_all');
+SELECT lsmb__create_role('gl_all',
+                         $DOC$
+                         This role combines GL transaction and batch creation
+                         with GL reporting and year-end processing.
+                         $DOC$
+);
 SELECT lsmb__grant_role('gl_all', rname)
   FROM unnest(array['gl_transaction_create'::text, 'gl_voucher_create',
                     'yearend_run', 'gl_reports']) rname;
 
-SELECT lsmb__create_role('timecard_add');
+SELECT lsmb__create_role('timecard_add',
+                         $DOC$
+                         This role allows adding time cards for which it needs
+                         read access to customers.
+                         $DOC$
+);
 SELECT lsmb__grant_role('timecard_add', 'contact_read');
 SELECT lsmb__grant_menu('timecard_add', node_id, 'allow')
   FROM unnest(array[100, 106, 8]) node_id;
@@ -1448,13 +1458,24 @@ SELECT lsmb__grant_perms('timecard_add', 'jcitems_id_seq', 'ALL');
 SELECT lsmb__grant_perms('timecard_add', 'jcitems', 'INSERT');
 SELECT lsmb__grant_perms('timecard_add', 'jcitems', 'UPDATE');
 
-SELECT lsmb__create_role('timecard_list');
+SELECT lsmb__create_role('timecard_list',
+                         $DOC$
+                         This role allows viewing the list of time cards;
+                         for which it needs read access to customers.
+                         $DOC$
+);
 SELECT lsmb__grant_role('timecard_list', 'contact_read');
 SELECT lsmb__grant_menu('timecard_list', 106, 'allow');
 SELECT lsmb__grant_perms('timecard_list', 'jcitems', 'SELECT');
 
 \echo ORDER GENERATION
-SELECT lsmb__create_role('orders_generate');
+SELECT lsmb__create_role('orders_generate',
+                         $DOC$
+                         This role combines the rights to generate orders from
+                         time cards, purchase orders from sales orders
+                         and consolidate (purchase and sales) orders.
+                         $DOC$
+);
 SELECT lsmb__grant_role('orders_generate', 'contact_read');
 SELECT lsmb__grant_perms('orders_generate', obj, ptype)
   FROM unnest(array['oe'::text, 'orderitems', 'business_unit_oitem']) obj,
@@ -1481,6 +1502,7 @@ SELECT lsmb__create_role('orders_sales_consolidate');
 SELECT lsmb__grant_role('orders_sales_consolidate', 'orders_generate');
 SELECT lsmb__grant_menu('orders_sales_consolidate', 61, 'allow');
 
+--###BUG: Duplicate with 'orders_generate'
 SELECT lsmb__create_role('orders_manage');
 SELECT lsmb__grant_role('orders_manage', rname)
   FROM unnest(array['timecard_order_generate'::text, 'orders_sales_to_purchase',
@@ -1488,7 +1510,13 @@ SELECT lsmb__grant_role('orders_manage', rname)
        ) rname;
 
 \echo FINANCIAL REPORTS
-SELECT lsmb__create_role('financial_reports');
+SELECT lsmb__create_role('financial_reports',
+                         $DOC$
+                         This role allows running of financial reports:
+                         Income Statement, Balance Sheet, Trial Balance and
+                         Inventory & COGS.
+                         $DOC$
+);
 SELECT lsmb__grant_role('financial_reports', 'gl_reports');
 SELECT lsmb__grant_menu('financial_reports', node_id, 'allow')
   FROM unnest(array[75,110,111,112,113,114]) node_id;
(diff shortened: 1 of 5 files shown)


ehuelsmann:
@neilt I think this is one of the items you're missing for the book: the ability to extract role documentation from our application. Here's at least a start.

neilt:
Excellent, I was going about this the other way around.  Getting the definitions in the book first, then moving them to the role comments.  But I think this is the best way.

sql/modules/Roles.sql
```diff
@@ -195,16 +197,37 @@ GRANT SELECT ON location_class_to_entity_class TO PUBLIC;
 
 
 \echo BASE ROLES
-SELECT lsmb__create_role('base_user');
+SELECT lsmb__create_role(
+  'base_user',
+  $DOC$
+  Users need to be given this role in order to be granted access to the database schema which holds all LedgerSMB objects.
```
neilt, line 203: Please add to base_user.

This role only allows access to menu items Preferences, Logout, and New Window.  The user basically cannot do anything without added additional roles.

sql/modules/Roles.sql
```diff
@@ -195,16 +197,37 @@ GRANT SELECT ON location_class_to_entity_class TO PUBLIC;
...
+SELECT lsmb__create_role('budget_approve',
+                         $DOC$
+                         This role allows searching, viewing and approving of budgets.
+                         $DOC$
+);
 SELECT lsmb__grant_role('budget_approve', 'budget_view');
-SELECT lsmb__create_role('budget_obsolete');
+SELECT lsmb__create_role('budget_obsolete',
+                         $DOC$
+                         This role allows searching, viewing and marking as obsolete of budgets.
```
neilt, line 228: How about

This role allows searching and viewing of budgets as well as marking them as obsolete.

neilt (changes requested):
Looking good, well very good.  I'll get started on a perl file to retrieve these descriptions from the database for the book.

sql/modules/Roles.sql
```diff
@@ -326,7 +382,14 @@ SELECT lsmb__grant_perms(role, 'file_incoming', 'DELETE'),
                     'file_attach_part', 'file_attach_eca']) role;
 
 \echo Contact Management
-SELECT lsmb__create_role('contact_read');
+SELECT lsmb__create_role('contact_read',
+                         $DOC$
+                         This role allows searching and viewing entities, persons and companies (contacts).
+
+                         Combine this role with one or more 'contact_class_*' roles to be able to access
+                         contacts of the specific class. By itself, this role does not provide sufficient rights.
```
christian-eriksson, line 390: Is this similar to the `base_user` that a user with `contact_read` don't really get any access from this role?

I'm not sure what the role actually permits, if I can search and view `entities, persons and companies` for what permission do I need the `contact_class_*`? Does the `contact_read` only allow the user to see a list of search results perhaps?

sql/modules/Roles.sql
```diff
@@ -356,18 +419,75 @@ SELECT lsmb__grant_perms('contact_read', 'file_entity', 'SELECT');
...
+);
+
+SELECT lsmb__create_role('contact_create',
+                         $DOC$
+                         This role allows searching and viewing existing contacts and creation of new contacts.
+
+                         Combine this role with one or more 'contact_class_*' roles to be able to access
+                         contacts of the specific class. By itself, this role does not provide sufficient rights.
+                         $DOC$
+);
```
christian-eriksson, line 490: Is it correct that this role provides the same permissions as `contact_read`, probably not?

I have the same questions here as for `contact_read` what does this role provide by itself?

sql/modules/Roles.sql
```diff
@@ -415,7 +539,14 @@ SELECT lsmb__grant_menu('employees_manage', 49, 'allow');
...
 
-SELECT lsmb__create_role('contact_edit');
+SELECT lsmb__create_role('contact_edit',
+                         $DOC$
+                         This role allows editing of existing contacts.
+
+                         Combine this role with one or more 'contact_class_*' roles to be able to access
+                         contacts of the specific class. By itself, this role does not provide sufficient rights.
+                         $DOC$
+);
```
christian-eriksson, line 549: I have the same question as for `contact_read` and `contact_create`, what does it do on its own, but I suspect it doesn't do anything?

sql/modules/Roles.sql
```diff
@@ -434,15 +565,26 @@ SELECT lsmb__grant_perms('contact_edit', 'eca_to_contact', 'ALL');
...
 
-SELECT lsmb__create_role('contact_delete');
+SELECT lsmb__create_role('contact_delete',
+                         $DOC$
+                         This role allows editing of existing contacts.
+
+                         Combine this role with one or more 'contact_class_*' roles to be able to access
+                         contacts of the specific class. By itself, this role does not provide sufficient rights.
+                         $DOC$
+);
```
christian-eriksson, line 575: does it to the same as `contact_edit` or does it allow users to delete contacts? and I'm guessing it doesn't do anything on its own?

Also, this role should probably be mentioned in the descriptions of the `contact_class_*` roles?

christian-eriksson (changes requested):

neilt:
I have the perl script written to gather the role comments from the database, so I hope this can be merged soon so I can test the script.

sql/modules/Roles.sql
ehuelsmann: The roles are complementary: you need both "contact_class_vendors" and "contact_read" in order to be able to read contact data of type "vendor" (but not be able to access "customer" data).

sql/modules/Roles.sql
ehuelsmann: it provides more than `contact_read`: contact read won't allow creation of new contacts. This role does.

sql/modules/Roles.sql
ehuelsmann: This role can change existing contacts (which inherently means being able to read their data). However, with this role you can't create new ones.

sql/modules/Roles.sql
ehuelsmann: Not the same as 'edit': with this role, you can search them, but not do anything with them but to delete them (so: no editing; no creation of new ones).

sql/modules/Roles.sql
ehuelsmann: Done.

sql/modules/Roles.sql
ehuelsmann: Done.

ehuelsmann (commented):
Resolved requests from @neilt.

sql/modules/Roles.sql
christian-eriksson: Would a better wording be something like:

> When paired with one or more `contact_class_*` role/-s, this role allows searching and viewing entities, persons and companies (contacts)
> 
> Each `contact_class_<resource>` role, when paired with `contact_read`, enables this access for the specific `<resoucre>`. On it's own, the `contact_read`-role does not provide any rights.

or similar?

sql/modules/Roles.sql
```diff
@@ -356,18 +421,75 @@ SELECT lsmb__grant_perms('contact_read', 'file_entity', 'SELECT');
...
+
+SELECT lsmb__create_role('contact_create',
+                         $DOC$
+                         This role allows creation of new contacts.
+
+                         Combine this role with one or more 'contact_class_*' roles to be able to access
+                         contacts of the specific class. By itself, this role does not provide sufficient rights.
+                         $DOC$
+);
```
christian-eriksson, line 492: I'd suggest a similar wording as suggested for `contact_read`.

sql/modules/Roles.sql
```diff
@@ -356,18 +421,75 @@ SELECT lsmb__grant_perms('contact_read', 'file_entity', 'SELECT');
...
+SELECT lsmb__create_role('contact_class_hot_lead',
+                         $DOC$
+                         This role allows access to hot sales lead contact data.
+                         Combine with 'contact_read', 'contact_create', 'contact_edit' to determine the type of access granted.
+                         $DOC$
+);
+SELECT lsmb__create_role('contact_class_cold_lead',
+                         $DOC$
+                         This role allows access to cold sales lead contact data.
+                         Combine with 'contact_read', 'contact_create', 'contact_edit' to determine the type of access granted.
```
christian-eriksson, line 469: aren't all `contact_class_*` roles missing the `contact_delete` option?

sql/modules/Roles.sql
```diff
@@ -415,7 +541,14 @@ SELECT lsmb__grant_menu('employees_manage', 49, 'allow');
...
 
-SELECT lsmb__create_role('contact_edit');
+SELECT lsmb__create_role('contact_edit',
+                         $DOC$
+                         This role allows editing of existing contacts and creation of new ones.
+
+                         Combine this role with one or more 'contact_class_*' roles to be able to access
+                         contacts of the specific class. By itself, this role does not provide sufficient rights.
+                         $DOC$
+);
```
christian-eriksson, line 551: I suggest the same wording as the one suggested for `contact_read`.

sql/modules/Roles.sql
```diff
@@ -434,15 +567,26 @@ SELECT lsmb__grant_perms('contact_edit', 'eca_to_contact', 'ALL');
...
 
-SELECT lsmb__create_role('contact_delete');
+SELECT lsmb__create_role('contact_delete',
+                         $DOC$
+                         This role allows removal of existing contacts.
+
+                         Combine this role with one or more 'contact_class_*' roles to be able to access
+                         contacts of the specific class. By itself, this role does not provide sufficient rights.
+                         $DOC$
+);
```
christian-eriksson, line 577: I suggest a similar wording as the one suggested in `contact_read`.

Also, this role `contact_delete` is not mentioned in the description of the `contact_class_*` roles.

christian-eriksson: I saw in the chat that you actually also need the `contact_read` role for this role to have effect, unless I misunderstood something. You might want to mention this here to (and in `contact_edit` and  `contact_create` if it's true for those as well. A suggestion would be:

>  When paired with one or more `contact_class_*`-role/-s, this role allows removal of existing contacts.
>
> Each `contact_class_<resource>`-role, when paired with `contact_delete`, enables this access for the specific `<resoucre>`. On it's own, the `contact_delete`-role does not provide any rights. Note that you also need `contact_read` for `contact_delete` to be applied as expected.

sql/modules/Roles.sql
ehuelsmann: yep. they are.

sql/modules/Roles.sql
ehuelsmann: Done.

sql/modules/Roles.sql
ehuelsmann: Done.

sql/modules/Roles.sql
ehuelsmann: Adjusted.

sql/modules/Roles.sql
ehuelsmann: Done.

sql/modules/Roles.sql
ehuelsmann: Used different wording, but, yes, this comment is incorporated too.

sql/modules/Roles.sql
```diff
@@ -415,7 +555,16 @@ SELECT lsmb__grant_menu('employees_manage', 49, 'allow');
...
 
 
-SELECT lsmb__create_role('contact_edit');
+SELECT lsmb__create_role('contact_edit',
+                         $DOC$
+                         When paired with one or more 'contact_class_*' role/-s, this role allows
+                         editing of existing entities, persons and companies (contacts).
+
+                         Each contact_class_<resource> role, when paired with contact_read, enables
+                         this access for the specific <resource>. On it's own, the contact_read-role
```
christian-eriksson, line 564: contact_read -> contact_edit

sql/modules/Roles.sql
```diff
@@ -356,18 +423,87 @@ SELECT lsmb__grant_perms('contact_read', 'file_entity', 'SELECT');
...
+                         $DOC$
+);
+
+SELECT lsmb__create_role('contact_create',
+                         $DOC$
+                         When paired with one or more 'contact_class_*' role/-s, this role allows
+                         creation of new entities, persons and companies (contacts).
+
+                         Each contact_class_<resource> role, when paired with contact_read, enables
+                         this access for the specific <resource>. On it's own, the contact_read-role
```
christian-eriksson, line 503: contact_read -> contact_create

sql/modules/Roles.sql
```diff
@@ -434,15 +583,31 @@ SELECT lsmb__grant_perms('contact_edit', 'eca_to_contact', 'ALL');
...
+SELECT lsmb__create_role('contact_delete',
+                         $DOC$
+                         When paired with one or more 'contact_class_*' role/-s, this role allows
+                         removal of existing entities, persons and companies (contacts).
+
+                         Note that in order to be able to search for contacts to be deleted, the user
+                         needs to be assigned the 'contact_read' role.
+
+                         Each contact_class_<resource> role, when paired with contact_read, enables
+                         this access for the specific <resource>. On it's own, the contact_read-role
```
christian-eriksson, line 595: contact_read -> contact_delete

christian-eriksson (changes requested):

sql/modules/Roles.sql
christian-eriksson: yeah I think I read too quick, and missed the `and creation of new contacts.` so it looked like it said the same for `contact_create` and `contact_read`.

sql/modules/Roles.sql
ehuelsmann: Changed. Will push in a sec.

sql/modules/Roles.sql
ehuelsmann: Changed. Will push in a sec.

sql/modules/Roles.sql
ehuelsmann: Changed. Will push in a sec.

sql/modules/Roles.sql
ehuelsmann: :+1:

sql/modules/Roles.sql
christian-eriksson: maybe it should say something like:

> allows editing of existing contacts, but not creation of new contacts.

just to be explicit?

sql/modules/Roles.sql
christian-eriksson: ah, I was referring to the `This role allows editing of existing contacts.` which should have been something like `allows for removal...`. But I think this was fixed already?

PR #7386, opened 2023-05-06
Centralization of some OpenAPI docs

ehuelsmann:
And at the same time add more documentation (instead of tripple dots)

### DIFF
--- a/lib/LedgerSMB/Routes/ERP/API/Languages.pm
+++ b/lib/LedgerSMB/Routes/ERP/API/Languages.pm
@@ -286,10 +286,10 @@ paths:
                     type: array
                     items:
                       type: object
-                items:
-                  type: array
                   items:
-                    $ref: '#/components/schemas/Language'
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/Language'
         400:
           $ref: '#/components/responses/400'
         401:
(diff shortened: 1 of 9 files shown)


lib/LedgerSMB/Routes/ERP/API/Languages.pm
```diff
@@ -274,7 +274,7 @@ paths:
       operationId: getLanguages
       responses:
         200:
-          description: ...
+          description: Returns the full set of configured languages
```
ylavoie, line 277: openapi & info sections left purposely?

ylavoie (changes requested):
openapi & info sections in Invoices should be checked too.

lib/LedgerSMB/Routes/ERP/API/Languages.pm
ehuelsmann: No; accidentally left behind. Fix pushed. Thanks for the review!

ehuelsmann:
> openapi & info sections in Invoices should be checked too.

@ylavoie this one is actually intentional: Invoices isn't using the DSL keywords from the router to declare API entrypoints (yet). Which makes it different from the others in that it uses its  own OpenAPI validator instead of the one from Router.pm. You're right, however, that this should be changed. I'll have a look at how difficult it is to change this.

ehuelsmann:
@ylavoie i think I can fix https://github.com/ledgersmb/LedgerSMB/actions/runs/4913494760/jobs/8773795061?pr=7386#step:19:330 by overriding discount to be a float instead of a bigfloat. Or should we change the schema?

PR #7399, opened 2023-05-14
Add ADRs regarding using the database for authn and authz

ehuelsmann:


doc/adr/0003-use-database-for-authn.md
```diff
@@ -0,0 +1,34 @@
...
+
+Date: Unknown
+
+## Status
+
+Accepted
+
+## Context
+
+Password storage is a weak point in many webapps: passwords are often
```
ylavoie, line 11: How about this:
The security of password storage in numerous web applications has always been concern, as passwords were frequently stored either in plain text or obfuscated forms instead of employing robust security measures like hashing.

In the authentication process of LedgerSMB, passwords are utilized and therefore need to be stored. However, we aim to avoid developing an insecure solution and strive for enhanced security measures.

ehuelsmann: I've replaced it with:

```markdown
## Context

The security of password storage has been and remains a major point of
concern in IT in general and with web applications in particular:
passwords often used to be stored in clear text or would use other
insufficiently secure storage mechanisms.

LedgerSMB uses passwords in its authentication process and therefor is
potentially affected by these problems too; however, we have a strong
desire to falling into these traps *and* leverage other (much more
heavily audited) open source solutions.
```

Is that Ok?

ylavoie: Good for me

PR #7413, opened 2023-05-20
Restore default AR/AP selection

ylavoie:
Restore the default AP & AR selection when multiple accounts are available during migration


ehuelsmann (approved):
Looks sane.

PR #7420, opened 2023-05-20
Restore default AR/AP selection

ylavoie:
Restore the default AP & AR selection when multiple accounts are available during migration


ehuelsmann:
I think we can safely skip this one: it's out of community support anyway.

PR #7422, opened 2023-05-20
Migrate sl30 mc currencies

ylavoie:
Fix SL3.0 migration with transactions using multiple currencies.

### DIFF
--- a/sql/upgrade/sl3.0.sql
+++ b/sql/upgrade/sl3.0.sql
@@ -1132,7 +1132,7 @@ INSERT INTO cr_report(chart_id, their_total, submitted, end_date, updated, enter
 -- The ID and matching post_date are entered in a temp table to pull the back into cr_report_line immediately after.
 -- Temp table will be dropped automatically at the end of the transaction.
 WITH cr_entry AS (
-SELECT cr.id::INT, cr.end_date, a.source, a.type, a.cleared::TIMESTAMP, a.amount::NUMERIC, a.transdate AS post_date, a.lsmb_entry_id
+SELECT cr.id::INT, cr.end_date, a.source, a.type, a.cleared::DATE, a.amount::NUMERIC, a.transdate AS post_date, a.lsmb_entry_id
     FROM cr_coa_to_account cta
     JOIN account c on cta.chart_id = c.id
     JOIN cr_report cr ON cr.chart_id = c.id
@@ -1167,15 +1167,14 @@ FROM (
 
 -- Patch for suspect clear dates
 -- The UI should reflect this
--- Unsubmit the suspect report to allow easy edition
-UPDATE cr_report SET submitted = false
-WHERE id IN (
-    SELECT DISTINCT report_id FROM cr_report_line
-    WHERE clear_time - post_date > 150
-);
 -- Approve valid reports.
-UPDATE cr_report SET approved = true
-WHERE submitted;
+UPDATE cr_report
+SET approved = true
+WHERE submitted
+  AND id NOT IN (
+    SELECT DISTINCT report_id FROM cr_report_line
+     WHERE NOT (clear_time - post_date BETWEEN 0 AND 150)
+);
 
 -- Log out the Migrator
 DELETE FROM users
--- a/sql/upgrade/sl3.0.sql
+++ b/sql/upgrade/sl3.0.sql
@@ -905,11 +905,22 @@ UPDATE business_unit_class
  WHERE id = 2
    AND EXISTS (select 1 from :slschema.project);
 
+-- Prefill approved status of transactions with gl entries
+INSERT INTO transactions (id, table_name, approved)
+SELECT id,'gl',approved
+FROM :slschema.gl;
+
 INSERT INTO gl(id, reference, description, transdate, person_id, notes)
     SELECT gl.id, reference, description, transdate, p.id, gl.notes
       FROM :slschema.gl
  LEFT JOIN :slschema.employee em ON gl.employee_id = em.id
  LEFT JOIN person p ON em.entity_id = p.id;
+
+-- Prefill approved status of transactions with ar entries
+INSERT INTO transactions (id, table_name, approved)
+SELECT ar.id,'ar',approved
+FROM :slschema.ar
+JOIN :slschema.customer ON (ar.customer_id = customer.id) ;
 
 --TODO: Handle amount_tc and netamount_tc
 insert into ar
@@ -932,6 +943,12 @@ SELECT
         ar.terms, description
 FROM :slschema.ar
 JOIN :slschema.customer ON (ar.customer_id = customer.id) ;
+
+-- Prefill approved status of transactions with ap entries
+INSERT INTO transactions (id, table_name, approved)
+SELECT ap.id,'ap',approved
+FROM :slschema.ap
+JOIN :slschema.vendor ON (ap.vendor_id = vendor.id) ;
 
 insert into ap
 (entity_credit_account, person_id,
--- a/sql/upgrade/sl3.0.sql
+++ b/sql/upgrade/sl3.0.sql
@@ -17,9 +17,6 @@
 
 BEGIN;
 
-ALTER TABLE :slschema.acc_trans DROP COLUMN IF EXISTS lsmb_entry_id;
-ALTER TABLE :slschema.acc_trans add column lsmb_entry_id SERIAL UNIQUE;
-
 -- Migration functions
 -- TODO: Can we do without?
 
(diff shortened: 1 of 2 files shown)


sql/upgrade/sl3.0.sql
```diff
@@ -2,7 +2,7 @@
 
 -- With help of a few conditional statements handled by the Template toolkit,
 -- this migration file handles migration from all SQL-Ledger version up to 3.0
--- to all Ledgersmb up to 1.6
+-- to all Ledgersmb up to 1.10
```
ehuelsmann, line 5: The commit message doesn't explain why we the migration target moves from 1.6 to 1.10; the idea was to have all migrations to a single history point, because that nicely fixes at least one side of the migration schema. At the time it was logical to choose 1.6.

Concluding: I think it's important to mention in the commit message why this shift is important.

sql/upgrade/sl3.0.sql
```diff
@@ -376,6 +373,22 @@ $$
         SELECT entity_id from users where username = SESSION_USER OR username = 'Migrator';
 $$ LANGUAGE SQL;
 
+-- Make sure transactions is filled. The actual trigger will be redefined
+-- in modules later.
+
+CREATE OR REPLACE FUNCTION track_global_sequence()
```
ehuelsmann, line 379: Why do you need this? We can insert the records in the transactions table based on what's in the gl/ar/ap tables right after filling the ar/ap/gl tables, right? Or we can fill the transactions table in the same query as we're using to fill the ar/ap/gl tables?

sql/upgrade/sl3.0.sql
```diff
@@ -461,12 +474,34 @@ UPDATE :slschema.vendor SET entity_id = (SELECT id FROM entity WHERE 'V-' || ven
 
 UPDATE :slschema.customer SET entity_id = coalesce((SELECT min(id) FROM entity WHERE 'C-' || customernumber = control_code), entity_id);
 
+--TODO: Do we need that for MC?
+INSERT INTO defaults(setting_key,value)
+    SELECT 'curr',array_to_string(array_agg(curr),':')
+    FROM :slschema.curr
+    WHERE rn=1;
```
ehuelsmann, line 481: I don't think this statement is correct, because it generates a string of the format "CAD:USD:EUR". Under "mc", the value of the "curr" key is the default currency code. Not all of them.

sql/upgrade/sl3.0.sql
```diff
@@ -461,12 +474,34 @@ UPDATE :slschema.vendor SET entity_id = (SELECT id FROM entity WHERE 'V-' || ven
...
+SELECT DISTINCT curr, curr
+FROM (
+          SELECT DISTINCT curr FROM :slschema.ar
+    UNION SELECT DISTINCT curr FROM :slschema.ap
+    UNION SELECT DISTINCT curr FROM :slschema.gl
+) xx
+WHERE NOT EXISTS (
+    SELECT 1 FROM :slschema.curr c
+    WHERE c.curr = xx.curr AND curr is not null
+);
```
ehuelsmann, line 497: You can simplify the WHERE clause by simply moving the first select into the FROM clause UNION too. That way, it'll simply be one of the factors.

sql/upgrade/sl3.0.sql
```diff
@@ -461,12 +474,34 @@ UPDATE :slschema.vendor SET entity_id = (SELECT id FROM entity WHERE 'V-' || ven
...
+
 --Entity Credit Account
 
 UPDATE :slschema.vendor SET business_id = NULL WHERE business_id = 0;
 INSERT INTO entity_credit_account
 (entity_id, meta_number, business_id, creditlimit, ar_ap_account_id,
-        cash_account_id, startdate, enddate, threshold, entity_class)
+        cash_account_id, startdate, enddate, threshold, entity_class, curr)
```
ehuelsmann, line 504: The "curr" here isn't new for mc, right? Just trying to understand what are fixes and what are incremental improvements.

sql/upgrade/sl3.0.sql
```diff
@@ -896,19 +920,24 @@ UPDATE business_unit_class
  WHERE id = 2
    AND EXISTS (select 1 from :slschema.project);
 
+ALTER TABLE gl DISABLE TRIGGER gl_audit_trail;
+ALTER TABLE gl DISABLE TRIGGER gl_prevent_closed;
```
ehuelsmann, line 924: Do you need to disable this trigger? I'd think you didn't set up the year-end table yet (and if you did, maybe it's an option to delay that), so that you don't need to disable the trigger... Or is this a performance measure? If the latter, could you please add a comment to that effect, substantiating the choice with your experience in your testing (e.g. by "disabling this trigger I was able to migrate a 2mio transactions company in 1min instead of 15mins")?

sql/upgrade/sl3.0.sql
```diff
@@ -896,19 +920,24 @@ UPDATE business_unit_class
...
+
 INSERT INTO gl(id, reference, description, transdate, person_id, notes)
     SELECT gl.id, reference, description, transdate, p.id, gl.notes
       FROM :slschema.gl
  LEFT JOIN :slschema.employee em ON gl.employee_id = em.id
  LEFT JOIN person p ON em.entity_id = p.id;
 
+ALTER TABLE gl ENABLE TRIGGER gl_prevent_closed;
```
ehuelsmann, line 932: That way, you can't forget to re-enable it :smile:

sql/upgrade/sl3.0.sql
```diff
@@ -896,19 +920,24 @@ UPDATE business_unit_class
...
 ALTER TABLE gl ENABLE TRIGGER gl_audit_trail;
 
 ALTER TABLE ar DISABLE TRIGGER ar_audit_trail;
+ALTER TABLE ar DISABLE TRIGGER ar_prevent_closed;
```
ehuelsmann, line 936: Same as for the GL trigger: do we really need to disable?

ehuelsmann (commented):
I'm surprised: our schema had an "fx_transaction" indicator. I'm pretty sure that came from SQL Ledger. "mc" removed that indicator by moving from the fx values into the "amount_tc" column. By doing so, many rows would be dropped.

What happened to the rows with the "fx_transaction" == TRUE?

sql/upgrade/sl3.0.sql
ylavoie: Reverted, 1.6 was indeed the proper endpoint.

sql/upgrade/sl3.0.sql
ylavoie: The code was producing the proper currency because of the rn restriction. Nevertheless, there is no need to aggregate a singleton array.

sql/upgrade/sl3.0.sql
ylavoie: Currencies were not migrated in SL3.0 transactions, so all were appearing under the default currency.,

sql/upgrade/sl3.0.sql
ehuelsmann: Ok. Where is this reversal?

sql/upgrade/sl3.0.sql
ylavoie: Removed

sql/upgrade/sl3.0.sql
ylavoie: Removed

sql/upgrade/sl3.0.sql
ylavoie: Transactions table in 1.6 didn't have all the fields it now possesses, so we cannot rely on the current trigger to fill the transaction table.

sql/upgrade/sl3.0.sql
ylavoie: Fixed

sql/upgrade/sl3.0.sql
ylavoie: Trigger insert into transaction table

sql/upgrade/sl3.0.sql
ehuelsmann: > Reverted, 1.6 was indeed the proper endpoint.

1.6 doesn't have `amount_tc` and `amount_bc`, yet your migration script uses these. So the target is *not* 1.6?

sql/upgrade/sl3.0.sql
ehuelsmann: Right. I guess my point was that after INSERT INTO ar SELECT ... FROM :slschema.ar, you could: INSERT INTO transactions (id, ...) SELECT id, ... FROM ar; then you don't need the trigger.

Come to think of it, I think the insertions need to happen into the transactions table before the insertions into the ar/ap/gl tables.

sql/upgrade/sl3.0.sql
ylavoie: SL migrations have been set to 1.6+mc since Dec 2017

ehuelsmann (approved):

PR #7427, opened 2023-05-27
Add ADR describing the decisions for our REST API

ehuelsmann:


ylavoie (approved):

freelock (commented):
This seems ok for a start, although a bit long in stretches, and I think we should also include GraphQL as we were considering that early on...

ehuelsmann:
@freelock, thanks for this rewrite! No comments; as far as I'm concerned, this can be integrated.

ylavoie (approved):
Very nice

PR #7444, opened 2023-06-08
Fix empty triggers

ylavoie:
Migration from previous versions should be independent of modules and only rely on schema.
This PR fixes the empty triggers in the database schema and removes the dependency on the trigger module.


ehuelsmann (approved):
Looking good.

PR #7477, opened 2023-07-02
Remove unused 'entity_id' and 'till' columns from AR/AP

ylavoie:
Complete removal of `entity_id` and `till` on AP/AR aging reports


ehuelsmann (approved):
Could you please add tests to trigger the function you've now fixed? we would have run into this problem earlier if there would ave been tests.

PR #7494, opened 2023-07-10
Update translations on master

ehuelsmann:
And fix the scripts to generate the translations.


utils/devel/extract-vue-json-to-po.sh
ylavoie, line 1: Isn't utils/Devel/extract-vue-template-translations-references in root?

ehuelsmann: yes. fixed.

ylavoie (approved):

PR #7531, opened 2023-07-23
Add more ADR states for a more complete process

ehuelsmann:


neilt (changes requested):
It is not clear to me what these terms mean.  For example, "Rejected", what was rejected, the text, or the idea of the change itself?  If it is the text, then I don't see the value in rejected or deprecated.

How can an architecture change be deprecated, it seems that it can be removed, or replaced.  But deprecated means it still there but will not remain in the future.  If this is meant to mean that we are replacing one with another, then I think that is probably too fine of detail.  

'Replaced by xxxxx' might be better.  I think it would be understood that there might be a transition time that is not tracked by the ADR entry.

If "Draft" means a proposed architecture change, then "Rejected" makes sense.  But then is "Proposed" a better state.   Of course "Draft" make sense if it only refers to the text.

I wonder if "No Longer Applicable" might be needed.

There needs to be clear explanation of what these terms mean.

ehuelsmann:
> There needs to be clear explanation of what these terms mean.

I've pushed a new variant of the text. Is that more like it?

neilt (approved):
Crystal clear.

Nitpicking, not sure if this is needed; "meaning that the statusses all relate to the
idea of the decision, not the current version of the text.".  This is clear from the numbered list.

PR #7534, opened 2023-07-23
Account checkpoint balances should equal cumulative transaction totals

ehuelsmann:


ehuelsmann:
From my 4-year old backlog; hidden in my repository stash...

PR #7541, opened 2023-07-25
Load fallback (two-letter) locale

ehuelsmann:
When the full fledge locale (e.g. en_US) isn't available, fall back to two letter locale name (i.e. 'en'). Also: catch the error(s) when either or both aren't available.

### DIFF
--- a/UI/src/i18n.js
+++ b/UI/src/i18n.js
@@ -15,7 +15,7 @@ function _mapLocale(locale) {
 }
 
 // eslint-disable-next-line import/no-unresolved
-import messages from '@/locales/en.json'
+import messages from '@/locales/en.json';
 import { nextTick } from "vue";
 
 const i18n = createI18n({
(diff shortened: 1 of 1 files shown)


UI/src/i18n.js
```diff
@@ -30,12 +30,25 @@ const i18n = createI18n({
...
+        }
+        catch (e) {
+            const strippedLocale = locale.replace(/_[a-z]+/i, '');
+            try {
+                const _messages = await import(/* webpackChunkName: "lang-[request]" */ `@/locales/${strippedLocale}.json`);
+                i18n.global.setLocaleMessage(strippedLocale, _messages.default);
+                locale = strippedLocale;
+            }
+            catch (f) {
+                /* eslint-disable no-empty */
```
christian-eriksson, line 49: you might want to set `locale = "en";` here otherwise it may be set to a non-existing locale `xy_AB` when the default locale is `en` when it is used later on at line 53. Not sure this is actually an issue in practice but then the `locale` variable should at least match the currently set locale.

christian-eriksson: not sure about the `i18n` api but maybe you can get the current locale from there

ehuelsmann: @ylavoie if we don't change locale to the fallback here, below, the page's text direction gets set to the direction associated with the locale passed in, but the actual locale is "en" so that it should be ltr, even if the preferred locale is arabic...

ylavoie: ltr should always follow the effective locale.

ylavoie (approved):

PR #7543, opened 2023-07-25
Save 'Ship To' data

### LINKED ISSUE #7453: shipto* variables do not use Shipping addresses
### Version

1.10.17

### What happened?

The invoice template populates a bunch of variables from the contact account. There is one set of variables for the address, email addresses, etc, and a different set prefixed with "shipto" presumably for the shipping addresses.

However, the shipto* versions all contain the same values as the non-shipto variables, even when there are different billing and shipping addresses set.

I tested moving the two addresses around between billing, sales, and shipping. Only the "Billing" address was used, for both the unprefixed and the shipto-prefixed variables. Sales and shipping addresses were entirely ignored.

The "shipto" variables should populate with the shipping details, if they are set, and only use the billing address info if they are not set.

There should be clear rules around how the template variables get populated, ideally shown on the template editing screens.

ehuelsmann:
Fixes #7453

### DIFF
--- a/old/bin/io.pl
+++ b/old/bin/io.pl
@@ -1354,17 +1354,20 @@ sub print_form {
         }
     }
 
-    # $logger->trace("\$form->{formname}=$form->{formname} \$form->{fax}=$form->{fax} \$shipto=$shipto \$form->{shiptofax}=$form->{shiptofax}");
+    $form->{shipto} = $shipto;
     if (! $shipto) {
         if (   $form->{formname} eq 'purchase_order'
-            || $form->{formname} eq 'request_quotation' )
-        {
+            || $form->{formname} eq 'request_quotation' ) {
             $form->{shiptoname}     = $form->{company};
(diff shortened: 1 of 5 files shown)


ehuelsmann:
@freelock, @tonka128: should this be backported?

freelock:
I'm not using this feature, so no need to backport for me...

freelock (commented):
Fixes look reasonable to me -- does this suppress printing the default address twice, if a shipping address is not selected?

ehuelsmann:
> Fixes look reasonable to me -- does this suppress printing the default address twice, if a shipping address is not selected?

No, the old behavior (shipping address = billing address) is still in place when no explicit shipping address was selected. No template variable gets set to indicate this defaulting rule was executed. That is relatively easy to add. It will allow for easy exclusion of the shipping address from the template. Is that an idea you would be able to work with?

PR #7548, opened 2023-07-27
Make right-to-left locale always assigned

ylavoie:
Left-to-right and right-to-left indicators must always follow the assigned locale


ehuelsmann (changes requested):
In general, the change looks good, but I think the trigger to update the UI is timed incorrectly, screwing up the tests.

ehuelsmann (approved):

PR #7639, opened 2023-10-05
Remove version number from cookies

ehuelsmann:
Without the version number, we don't need to rename the cookie on each release, removing a major maintenance nuisance.

[ci skip]


ylavoie (commented):
bin/ledgersmb-server.psgi Still refers to a versioned cookie.

ehuelsmann:
> bin/ledgersmb-server.psgi Still refers to a versioned cookie.

Thanks! Addressed.

PR #7641, opened 2023-10-06
Work around Locale::CLDR::Locales::* loading 'bignum'

ehuelsmann:
Loading 'bignum' causes $Math::BigInt::upgrade and $Math::BigFloat::downgrade to be overwritten. However, we want to make sure these values stay as they are set by LedgerSMB::PGNumber (which is: no downgrading!)

This is necessary for Perl 5.36+ compatibility.

### DIFF
--- a/lib/LedgerSMB.pm
+++ b/lib/LedgerSMB.pm
@@ -233,6 +233,8 @@ use List::Util qw( pairgrep );
 use Locale::CLDR;
 use Locales unicode => 1;
 use Log::Any;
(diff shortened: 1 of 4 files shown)


ehuelsmann:
Stimulating the condition this needs to fix, is done by loading a page which triggers Locale::CLDR to be loaded (e.g. by loading a search page) and then loading a number-heavy page.

ylavoie (approved):

PR #7652, opened 2023-10-09
Add the ability to have documentation in CoA definition XMLs

ehuelsmann:


neilt:
Is that documentation for root or for coa?

ehuelsmann:
It's for the whole: there's no way to acquire only the CoA; the rest is actually in support of the CoA: tax rates, default accounts and currencies.

PR #7692, opened 2023-11-04
Msw 2

ylavoie:
Msw 2 changed the API and required an adjustment of all tests.
It also enforce the validity of URL used in fetches, so we had to make sure that the current location was always set before fetching.


ehuelsmann (approved):
Overall, I think this does the job of moving to msw2. I'm not seeing the benefits in our tests for the API changes that msw2 created. However, I guess our tests are too small to show the benefit.

The thing I should have brought to your attention when these tests were originally developed, is that these tests use a different coding style ("modern JavaScript") than the main JS code base: it doesn't use semi-colons  to terminate statements. It would have been nice to stick to one style, but I guess that's water under the bridge.

PR #7695, opened 2023-11-05
Include only our static html in the cache

ylavoie:
Make sure to exclude Vue locales from the Dojo cache to prevent invalidation between github steps.


ehuelsmann (approved):
Globs look good now. Thanks!

PR #7727, opened 2023-11-23
Remove axios

ylavoie:
Standardize OpenAPI tests by moving from axios to fetch


ehuelsmann (changes requested):
Thanks! I see this stops the failures in the test site. I like that.

Could you add the missing semi-colons? When that's done, I think we're all set with this PR.

ehuelsmann:
Why is it fine to transform a strong ETAG to a weak etage?!

ehuelsmann (approved):

PR #7744, opened 2023-12-01
Fix encoding of the column data, including col_id

### LINKED ISSUE #7739: Calling General Ledger->Chart of Accounts in PDF format fails to render the Credits and Debits
### Version

1.10.25

### What happened?

Calling General Ledger->Chart of Accounts in PDF format fails to render the Credits and Debits.  Calling for CSV format presents the values as shown on the relevant screen.

The "number" columns should have rendered.

ehuelsmann:
Due to the fact that all column data was encoded for output-format-safety, the CoA and Trial Balance reports didn't show some columns in PDF format (with LaTeX intermediate representation). The column *keys* in the ROW data do *not* get encoded, whereas the column keys (stored in the column data as values) *do* get encoded. Due to this mismatch, the column name 'credit_balance' differed between the row data and the column data (the latter containing the encoded value 'credit\_balance').

Fixes #7739.

### DIFF
--- a/lib/LedgerSMB/Report.pm
+++ b/lib/LedgerSMB/Report.pm
@@ -576,11 +576,13 @@ sub _render {
             new_heads       => $replace_hnames,
             name            => $self->name,
             hlines          => $self->header_lines,
-            columns         => \@columns,
             order_url       => $self->order_url,
             buttons         => $self->buttons,
             options         => $self->options,
             rows            => \@rows,
+        },
+        {
+            columns         => \@columns
         });
 }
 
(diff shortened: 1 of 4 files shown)


neilt:
1. In the PDF the values do not align.
Screenshot 2023-12-01 at 1 46 13 PM

2. All formats except PDF exported the last column filled with [Delete], which seems superfluous, but not necessarily a bug.

Otherwise, all exports worked.

neilt (commented):
See comment above, otherwise looked ok

PR #7770, opened 2023-12-03
Add various ADRs about the use of state machines

ehuelsmann:
Before merging:

* [ ] Remove empty enumerations

### DIFF
--- a/doc/adr/0016-state-machines-for-UI-state.md
+++ b/doc/adr/0016-state-machines-for-UI-state.md
@@ -12,7 +12,9 @@ Modern (JavaScript-based) User Interfaces have many components and lots of
 concurrent actions going on.  Due to the combination and number of possible
 actions being triggered, the integration between components has become
 increasingly complex over the years -- especially when considering
-asynchronicity and error handling.
+asynchronicity and error handling.  See the post ['Modeling UI state using
+a finite state machine'](https://xiaoyunyang.github.io/post/modeling-ui-state-using-a-finite-state-machine/)
+for an extensive treatment of the topic.
 
 The JavaScript/front-end development community has increasingly been using
 [state machines](https://en.wikipedia.org/wiki/Finite-state_machine) and
@@ -24,16 +26,23 @@ machine engine in the JavaScript ecosystem, featuring over 1 million downloads
 per week (ref: 2023-12-02).  It implements the *state chart* state machine
 type.  XState however is an extensive library with an extensive 'DSL' to
 describe state machines.  There are many alternatives implementing state charts
-or state machines that are less extensive.  One such example is [robot3](https://thisrobot.life/)
-which is extremely small, supports sub-state machines and features integration
-for React, PReact and Lit.  Its DSL is similarly small and thereby easy to
-read.  None of the integrations are of interest for LedgerSMB, but they
-indicate interest from a wider community.
+or state machines that are less extensive.  One such example is
+[robot3](https://thisrobot.life/) which is extremely small, supports sub-state
+machines and features integration for React, PReact and Lit.  Its DSL is
+similarly small and thereby easy to read.  None of the integrations are of
+interest for LedgerSMB, but they indicate interest from a wider community.
 
 Considering that with its move to Vue, LedgerSMB is going in the same
-direction as other projects -- featuring a heavy (JavaScript-based) client
+direction as other projects -- featuring a rich (JavaScript-based) client
 application -- the project similarly needs a way to control the expected
 complexity.
+
+Note that - as pointed out by [Alex Russell in his post called 'The market
+for lemons'](https://infrequently.org/2023/02/the-market-for-lemons/) - the
+client can be rich, but should be lightweight in order, sticking as close
+as possible to web standards. (In his [footnote
+[1]](https://infrequently.org/2023/02/the-market-for-lemons/#fn-alex-approved-1),
+he explains that he considers Vue to fall into that category as well.)
 
 ## Decision
 
--- a/doc/adr/0016-state-machines-for-UI-state.md
+++ b/doc/adr/0016-state-machines-for-UI-state.md
@@ -1,0 +1,59 @@
+# 0016 State machines for UI state management
+
+Date: Unknown
+
+## Status
+
+Accepted
+
+## Context
+
+Modern (JavaScript-based) User Interfaces have many components and lots of
+concurrent actions going on.  Due to the combination and number of possible
+actions being triggered, the integration between components has become
+increasingly complex over the years -- especially when considering
+asynchronicity and error handling.
+
+The JavaScript/front-end development community has increasingly been using
+[state machines](https://en.wikipedia.org/wiki/Finite-state_machine) and
+[state charts](https://en.wikipedia.org/wiki/UML_state_machine) to solve this
+issue of complexity.
+
+[XState](https://stately.ai/docs/xstate) is the - by far - largest state
+machine engine in the JavaScript ecosystem, featuring over 1 million downloads
+per week (ref: 2023-12-02).  It implements the *state chart* state machine
+type.  XState however is an extensive library with an extensive 'DSL' to
+describe state machines.  There are many alternatives implementing state charts
+or state machines that are less extensive.  One such example is [robot3](https://thisrobot.life/)
+which is extremely small, supports sub-state machines and features integration
+for React, PReact and Lit.  Its DSL is similarly small and thereby easy to
+read.  None of the integrations are of interest for LedgerSMB, but they
+indicate interest from a wider community.
+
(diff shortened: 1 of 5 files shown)


doc/adr/0016-state-machines-for-UI-state.md
```diff
@@ -0,0 +1,59 @@
...
+type.  XState however is an extensive library with an extensive 'DSL' to
+describe state machines.  There are many alternatives implementing state charts
+or state machines that are less extensive.  One such example is [robot3](https://thisrobot.life/)
+which is extremely small, supports sub-state machines and features integration
+for React, PReact and Lit.  Its DSL is similarly small and thereby easy to
+read.  None of the integrations are of interest for LedgerSMB, but they
+indicate interest from a wider community.
+
+Considering that with its move to Vue, LedgerSMB is going in the same
+direction as other projects -- featuring a heavy (JavaScript-based) client
```
freelock, line 34: s/heavy/rich/ ? fat?

I think we should strive to keep the front end light, not heavy -- but add rich functionality to improve the user experience. https://infrequently.org/2023/02/the-market-for-lemons/ (author is the original author of Dojo, Alex Russell -- see footnote [1] where he does mention Vue in the list of frameworks that care about performance/small footprint).

So we should emphasize trying to stick as close to vanilla js and css as we can, and use tools like Vue and Pinia as added sugar...

doc/adr/0016-state-machines-for-UI-state.md
```diff
@@ -0,0 +1,59 @@
...
+management as well as state transition explicit.  Similar to UI composition
+with components, will different parts of the UI be governed by their own state
+machines.
+
+The state machine library used by the project will be robot3, which uses XState
+as an example, but chooses to be as lean as possible in its implementation.
+
+## Consequences
+
+1. Any view or component will come with a state machine to describe and manage
```
freelock, line 50: Hmm I'm thinking it's a "document" that would have an associated state machine, and then present different views and components based on what's appropriate for its current state

doc/adr/0016-state-machines-for-UI-state.md
```diff
@@ -0,0 +1,59 @@
...
+   its internal state and the allowed transitions
+2. Components don't change their own state: in response to events (their own or
```
freelock, line 52: I think of a component as a visual element or widget, something used to display some subset of the fields/information about something else, which I'm calling here a document.

Components would be:
- Address widget
- Line item widget
- Collection of line items
- Collection of order adjustments
- Order display (combining many of those)
- Action buttons (to trigger state transitions)
- Filters (to filter a list of documents)

... there is some internal/local state on many of these (read-only, disabled, active, selected, etc) but not to the level they are linked directly to a state machine -- their state comes from the state of the parent document.

"Documents" would be:

- Sales invoice
- Vendor invoice
- AR transaction
- AP transaction
- GL transaction
- Sales Order
- Purchase Order

... etc.

These would be the objects tightly linked to one (or more) state machines, with an active state.

doc/adr/0017-state-machines-for-resource-state-management.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Consequences
+
+1. A state machine engine must be selected
+2. Existing resources must be enhanced with state machines
+   and their implied state must be reverse-engineered
+3. New resources must not be introduced without state machines
+4. A design needs to be created mapping available state machine
+   actions to web-API responses and requests to trigger actions
+
```
freelock, line 46: 5. State transitions must be logged, with the time of the transition and the user initiating it.

(suggestion... otherwise this looks good).

This does make me think of another ADR to consider -- move the permission model from applying only to data access, to one that is used to enforce access on state transitions.

There are probably a few places where we want to restrict all access to particular columns/rows in the database -- but for the most part this is way too granular, and workflow transitions are a lot more appropriate place to enforce access controls.

doc/adr/0018-resource-state-machine-engine-with-dependency-injection.md
```diff
@@ -0,0 +1,63 @@
...
+   been selected as the state machine engine.
+
+## Consequences
+
+1. In combination with [ADR 0002](), this ADR implies that modification of
+   *database* state needs to happen through the procedural database API.
+2. A design needs to be created supporting easy customization of state
+   machines (further: workflows).
+3. 
+
```
freelock, line 62: This sounds great! I would recommend taking a look at the Drupal ECA module - https://www.drupal.org/project/eca , and the BPMN modeler: https://www.drupal.org/project/bpmn_io .

BPMN.io is a standalone JS modeler for managing events, conditions, actions, and gateways, all through an interface that essentially lets you draw flow charts, and is gaining traction across other projects.

Drupal's ECA module uses BPMN with Drupal's standard plugin system, providing Event, Condition, and Action plugins that can be defined in code, and easily added/configured/implemented through BPMN with no code.

This would be tremendous for adding things like payroll rules, taxes, etc.

doc/adr/0019-configuration-using-dependency-injection.md
```diff
@@ -0,0 +1,69 @@
...
+## Consequences
+
+1. The current INI format configuration file will be replaced,
+   because Beam::Wire needs a hierarchical format.
+2. The code base needs carefully designed extension points for which the
+   configuration file provides the implementation (with fixed APIs); natural
+   candidates are functionalities which have a high likelihood of needing
+   customization, such as document generators and bank statement importers.
+3. 
+
```
freelock, line 68: This seems a bit long, for just an ADR saying we're going to use Yaml instead of INI...

But I'm not sure how to improve it...

freelock (commented):
Some thoughts...

doc/adr/0016-state-machines-for-UI-state.md
ehuelsmann: Added.

doc/adr/0016-state-machines-for-UI-state.md
ehuelsmann: We discussed this in the channel just a few moments ago, but summarizing here for other reviewers: I actually *did* mean UI component, following the thoughts as expressed in https://xiaoyunyang.github.io/post/modeling-ui-state-using-a-finite-state-machine/.
I'm considering how to incorporate the nuances we discussed.

doc/adr/0017-state-machines-for-resource-state-management.md
ehuelsmann: > This does make me think of another ADR to consider -- move the permission model from applying only to data access, to one that is used to enforce access on state transitions.

This is (at a very small scale) already what I'm doing: e.g. posting a transaction or invoice is restricted by ACL (https://github.com/ledgersmb/LedgerSMB/blob/master/lib/LedgerSMB/Workflow/Condition/ACL.pm) where a user is being checked to have the right role to even be offered the button (but enforcement is within the workflow, not at the UI level).

ehuelsmann: So, yes, lets write that ADR too.

doc/adr/0019-configuration-using-dependency-injection.md
ehuelsmann: By which I mean that the selection of YAML is a consequence of going with dependency injection and Beam::Wire.

doc/adr/0018-resource-state-machine-engine-with-dependency-injection.md
ehuelsmann: This is an interesting comment to follow up with a new ADR. The engine we have now allows to install listeners (called observers). These are invoked at specific points in the processing of the workflow's actions. Workflow execution can be influenced using Conditions. Observers can start (independent) workflows. All of these components can be loaded and combined through dependency injection at configuration time (= server startup time). It doesn't (yet) bind to any existing processes (although as soon as we rewrite all documents to use workflows for their lifecyle, I guess we're almost there)

doc/adr/0016-state-machines-for-UI-state.md
ehuelsmann: So we agreed that there could be use-cases, but we need to be careful not to put too much (incomprehensible) boilerplate in place.

doc/adr/0017-state-machines-for-resource-state-management.md
ehuelsmann: Suggestion to this extent added to the README which lists the "ADRs still to be done"

doc/adr/0018-resource-state-machine-engine-with-dependency-injection.md
ehuelsmann: For now, I've moved this idea to #7782 .

ehuelsmann:
Added @neilt's suggestion in a new issue.

PR #7820, opened 2023-12-15
Fix 'delete' link shown on account used in parts config

### LINKED ISSUE #7812: Unable to delete an account
### Version

1.11.4

### What happened?

Trying to delete an account item that was previously an Inventory summary account (but is no longer) throws the following error.

`23503: ERROR: update or delete on table "account" violates foreign key constraint "parts_inventory_accno_id_fkey" on table "parts"
DETAIL: Key (id)=(22) is still referenced from table "parts".
CONTEXT: SQL statement "DELETE FROM account WHERE id = in_id"
PL/pgSQL function account__delete(integer) line 17 at SQL statement

dbversion: 1.11.4, company: PlayAMS`

This is probably as a result of the error reported in #7811 not allowing a part to be deleted

ehuelsmann:
Fixes #7812

### DIFF
--- a/sql/modules/Account.sql
+++ b/sql/modules/Account.sql
@@ -720,6 +720,19 @@ eca(account_id) AS (
     SELECT DISTINCT cash_account_id
       FROM entity_credit_account
 ),
+p(account_id) AS (
+  SELECT DISTINCT inventory_accno_id
+    FROM parts
+           UNION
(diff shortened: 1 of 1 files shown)


ehuelsmann:
This completed successfully, but wasn't reported back to GitHub...

PR #7844, opened 2023-12-16
Change setup.pl's page to show the loader icon while it's loading

### LINKED ISSUE #7710: Page code not hidden while page is loading resulting in wrong image being shown when logging into setup.pl
### Version

1.11.3

Firefox, Safari (MacOS X), Other

### What happened?

Incorrect page layout during setup.pl login while CSS is loading.  Repeatable on Firefox with private browsing. The other browsers seem to work correctly when the page and/or CSS has been cached, but also fail when there is no cache.

See attached video

A correctly formatted page should have been initially shown.

ehuelsmann:
To achieve this, loading the JS code needs to be moved to the body; however, since the HtmlWebpackPlugin doesn't distinguish LINK and SCRIPT tags, this also moves the 'bootstrap.css' stylesheet link to the body...

Fixes #7710

### DIFF
--- a/UI/lib/ui-header.html
+++ b/UI/lib/ui-header.html
@@ -70,7 +70,6 @@
             version: "[% version %]"
         };
     </script>
-    ${htmlWebpackPlugin.tags.bodyTags}
     <meta name="robots" content="noindex,nofollow" />
 </head>
 [% content %]
--- a/UI/css/system/setup.css
+++ b/UI/css/system/setup.css
@@ -8,9 +8,30 @@ div#loading {
     background: white;
     height: 100%;
     left: 0;
+    position: absolute;
     top: 0;
     width: 100%;
     z-index: 100;
+}
+
+div.setupconsole {
+    position: relative;
+}
+
+body:not([data-lsmb-done]) #loading {
+    display: block;
+}
+
+body:not([data-lsmb-done]) #logindiv {
+    visibility: hidden;
+}
+
+body[data-lsmb-done="true"] #loading {
+    display: none;
+}
+
+body[data-lsmb-done="true"] #logindiv {
+    visibility: visible;
 }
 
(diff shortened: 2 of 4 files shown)


ehuelsmann:
@neilt here's a branch with the same patch applied as the one posted in #7710.

neilt:
> @neilt here's a branch with the same patch applied as the one posted in #7710.

I am probably not going to be able to review this.

ehuelsmann:
@ylavoie this works, but inserts a stylesheet link in the body instead of in the head. Reading the docs of the responsible plugins, I'm not seeing how to correct that.

Do you?

UI/lib/ui-header.html
```diff
@@ -23,7 +23,6 @@
         # *persistent*
         # http://www.w3.org/TR/html401/present/styles.html#h-14.3.1
     -%]
-    <%= htmlWebpackPlugin.tags.headTags %>
```
ylavoie, line 26: Let the headTags there, remove the bodyTags below and assign the HtmlWebpackPlugin `inject` parameter to 'body' instead of false in webpack.config.js.
This will let HtmlWebpackPlugin manage assets and load them in the desired section

ylavoie (changes requested):

UI/lib/ui-header.html
ehuelsmann: Just checked. The effect is that the scripts are injected *after* the closing `</html>` tag. Are we Ok with that?

ylavoie: It does load the css in head, scripts after body and shows the desired behaviour, no?

ehuelsmann: Absolutely. I'll PR.

ylavoie (approved):

PR #7874, opened 2023-12-25
Add ADR - for discussion - about the use of global state

ehuelsmann:


neilt (approved):
I don't think the justifications given in the 2 paragraphs starting at line 48 justify the decisions if they were made today.  Maybe just add a more 'at the time of this decision' vibe, because some of these are not so true today.

ehuelsmann:
Based on discussion in the chat, I now believe this ADR is a *new* one and we need to document the original decision in an ADR that's to be written.

ehuelsmann:
@neilt I've repurposed this PR for a different ADR since the prior one needs more (technical) arguments.

doc/adr/0100-restrictions-on-use-of-global-state-and-localized-variables.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Context
+
+Global state is the prevalent design in 'old code'[^1]; this means that
+variables are either declared as `our` variables (or not declared at all,
+which is the same thing when 'strict' and 'warnings' are turned off).
+Originally, there was *no* use of `my` (local) variables.  This design
+often causes unexpected and breaking side-effects.
+
+In more recent code, global state is used store per-request values such
```
neilt, line 17: s/used store/used to store/

doc/adr/0100-restrictions-on-use-of-global-state-and-localized-variables.md
```diff
@@ -0,0 +1,77 @@
...
+    assume data is being presented in internal formats (i.e.
+    PGNumber/PGDate/PGTimestamp)
+
+## Consequences
+
+ 1. The outer-most layers must stop passing data from `$form` or `$request`
+    'as-is' to the next layer; instead taking responsibility for explicit
+    data type conversion
+ 2. `LedgerSMB::App_State` needs to be removed from the code base
+ 3. `LedgerSMB::MooseTypes` `coerce`-support needs to be removed
```
neilt, line 63: Both of the above lines should be worded the same. Either "removed" or "removed from the code base"

doc/adr/0100-restrictions-on-use-of-global-state-and-localized-variables.md
```diff
@@ -0,0 +1,77 @@
...
+    input is expected
+ 5. Routines for formatting user readable output need to be used everywhere
+    where output is expected to be generated according to user preferences
+ 6. `local` should be used to limit the scope of impact on Perl built-in
+    variables such as `$@` and `%ENV` as well as per-request (global)
+    variable assignment to ensure automatic state cleanup
+ 7. The "domain specific language" (DSL) used to specify API entry points
+    is a good example of the exception to rule (1) as it simplifies the
+    code base; yet the global state is restricted to the start-up phase
+    of the application thereby not hiding side-effects during execution
```
neilt, line 74: This paragraph was not entirely clear to me.  As an example maybe it could be a little more specific.

neilt (approved):
Looks good except as noted

doc/adr/0100-restrictions-on-use-of-global-state-and-localized-variables.md
ehuelsmann: Done.

doc/adr/0100-restrictions-on-use-of-global-state-and-localized-variables.md
ehuelsmann: Done.

ehuelsmann:
Thanks for your review!

PR #7930, opened 2024-01-15
Fix warning during startup without YAML config

### LINKED ISSUE #7927: Resolve ledgersmb.yaml
### Version

1.11.8

### What happened?

Subject file exists in two locations in different forms - t/ledgersmb.yaml and doc/conf/ledgersmb.yaml.
Information I have received indictaes that it should exist at the root level in order to avoid complaints when starting starman. starman does actually start despite the error.

Resolve the correct location of ledgersmb.yaml by either mv, cp, or ln -s

ehuelsmann:
Fixes #7927


ehuelsmann:
Doesn't apply cleanly.

PR #7942, opened 2024-01-20
Require support for 'SameSite=Strict' cookies

ehuelsmann:
By requiring 'SameSite=Strict' cookie support, we "loose" support for Safari 1.11 -- however, the effect of not supporting "SameSite" at all isn't apparent from various searches on the web. Meaning that it may very well still work, but just not as securely as desired.

The browser is almost 6 years old, so it's time for people to migrate to newer browsers. Even if no safari is available on the MacOS version, both Chrome/Chromium and Firefox will be.

@ylavoie, could you please review?


ehuelsmann:
@ylavoie sorry about the merge. it was unintended. Please still review!

PR #8068, opened 2024-03-09
Add timeframes for unspecified 'Date's on ADRs

### LINKED ISSUE #7783: Enhance ADRs with date periods where they now list "Unknown"
One of the things that might help newcomers to LedgerSMB would be a link to the PR for some code that shows how the implementation was done in LSMB.   That might be the first implementation, if it was done right the first time, or a later implementation, if it serves as a better example.  

In this case, implementation means how existing code was changed to utilize robot3 or how a new view or component was implemented using robot3.

Rather than have a status date of 'Unknown' which provides no useful implementation, how about considering 'Prior to 4 Dec 2023' or a less specific guess, like '2009'. Another example 'between 2014 and 2015', or with 'LSMB release 1.6'.  Any way that we can get more context is good.  Right now it does not matter, but this will provide better insight in 10 years.

_Originally posted by @neilt in https://github.com/ledgersmb/LedgerSMB/pull/7770#pullrequestreview-1762294192_

ehuelsmann:
Closes #7783

### DIFF
--- a/doc/adr/0001-database-restricted-to-postgresql.md
+++ b/doc/adr/0001-database-restricted-to-postgresql.md
@@ -1,6 +1,6 @@
 # 0001 Only support PostgreSQL at the database layer
 
-Date: Unknown
+Date: Early in the 1.3 release cycle
 
 ## Status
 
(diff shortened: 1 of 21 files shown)


neilt (approved):
Much better

PR #8070, opened 2024-03-11
Lint markdown

ylavoie:
- Reenable linting markdown files
- Cleanup MD files


ehuelsmann (commented):
I don't see the value of this linting, to be honest. If the other developers want it, I'm fine, but for me, this seems like mostly red tape. E.g. the space indenting that's being removed, was added explicitly for readability as raw text files...

ehuelsmann:
@neilt, @freelock any comments to my comment? Do you think we need this?

neilt (commented):
In general, I am in favor of lint processing because it improves consistency.

I also do not like the indentation space changes.  These files read better with the existing indentation when not using markdown. However, the files are not consistent. 

Are there rules we can adopt with the linter that will consistently force 2 lines before every '##' and a single space prior to any list delimiter ("*").

I like that all "-" are changed to "*", this should be consistent across all md files.

The change to doc/adr/README.md is better. Forcing ascii art to be inside the code delimiters is better.

I don't understand the changes in doc/csv-import/README.md. What rule is forcing this change?  It looks like the line lengths are being shortened, but the lines above the changes are longer.

neilt (commented):
Also, I did not find the config file `../.markdownlint.json` in the PR.  Did I miss it?

PR #8086, opened 2024-03-18
Remove entity_class from employee search

ehuelsmann:
When searching for employees, the "Entity class" drop-down is irrelevant. So is the "Account Number" search criterion.
Further more should the columns reported on employee search pertain to the employee, not to customers/vendors and don't show "Credit Accounts" on the employee page.

Most importantly: find all entities for which an `entity_employee` record exists, instead of searching for entities with `entity_class == 3`.

Remarks:

* The employee page's "Contact Info" still shows the ability to link billing related contact data (Billing E-mail, Billing Cc and Billing Bcc). This should probably be suppressed and for the employee view -- only shown on the customer/vendor view?


ehuelsmann:
Remarks:

* The employee page's "Contact Info" still shows the ability to link billing related contact data (Billing E-mail, Billing Cc and Billing Bcc). This should probably be suppressed and for the employee view -- only shown on the customer/vendor view?

ehuelsmann:
@neilt thanks for the review!

PR #8097, opened 2024-03-26
Remove 'entity_class' from the entity

ehuelsmann:
This removes the confusing point where entities are employees, customers or vendors, but they are not truely, because either an employee, customer or vendor (i.e. "accounts") still need to be created.

### DIFF
--- a/sql/modules/Company.sql
+++ b/sql/modules/Company.sql
@@ -14,7 +14,6 @@ DROP TYPE IF EXISTS company_entity CASCADE;
 
 CREATE TYPE company_entity AS(
   entity_id int,
-  entity_class int,
   legal_name text,
   tax_id text,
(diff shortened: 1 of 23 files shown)


sql/modules/Company.sql
```diff
@@ -14,7 +14,6 @@ DROP TYPE IF EXISTS company_entity CASCADE;
 
 CREATE TYPE company_entity AS(
```
neilt, line 15: Since we are modifying the type 'company_entity' we should take the opportunity to add a SQL comment explaining the type and its use.

neilt (approved):

sql/modules/Company.sql
ehuelsmann: Yep. Done.

PR #8113, opened 2024-04-03
Remove dead code in workflow loader

ehuelsmann:


ehuelsmann:
This thing needs more checking.

PR #8136, opened 2024-04-08
Move some functionality from 'workflow scripts' to workflows

ehuelsmann:
Moves the `del` and `approve` actions from workflow scripts (`io.pl`, `is.pl`, `io.pl` and `gl.pl`) to workflow actions.


ehuelsmann:
The reason that the action only modified the workflow context is that it can not modify the database' it's the persister's role to know how to reflect the workflow state (ie the context) to the database. 

In order not to hard for database states into the persister, the actions trigger persister behavior through the context state.

PR #8169, opened 2024-04-20
Fix Balance Sheet & Income Tax comparisons report by dates or by periods

ylavoie:
Restore Balance Sheet & Income Tax Comparison reports when selection is done by dates or by periods

### DIFF
--- a/lib/LedgerSMB/Report/Dates.pm
+++ b/lib/LedgerSMB/Report/Dates.pm
@@ -151,7 +151,7 @@ sub _collect_dates_comparisons {
         push @dates, {
             from_date => LedgerSMB::PGDate->from_input($args{"from_date_$i"},$args{formatter_options}),
             to_date => LedgerSMB::PGDate->from_input($args{"to_date_$i"},$args{formatter_options}),
-            column_path_prefix => [ $i ]
+            column_path_prefix => [ sprintf('%02u', $i) ]
         };
     }
 
--- a/lib/LedgerSMB/Report/Dates.pm
+++ b/lib/LedgerSMB/Report/Dates.pm
@@ -125,7 +125,6 @@ These are the first and last date in acc_trans.
 
 =cut
 
-
 sub get_bracket_dates {
     my ($self) = @_;
     my $return_hashref = {};
@@ -241,9 +240,6 @@ sub _set_lazy_dates {
               $self->date_to;
               return;
 }
-
-
-
 
 before 'render' => sub {
               my ($self) = @_;
--- a/lib/LedgerSMB/Report/Dates.pm
+++ b/lib/LedgerSMB/Report/Dates.pm
@@ -150,8 +150,8 @@ sub _collect_dates_comparisons {
 
     foreach my $i (1 .. $args{comparison_periods}) {
         push @dates, {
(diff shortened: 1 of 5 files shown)


lib/LedgerSMB/Scripts/pnl.pm
```diff
@@ -74,9 +74,8 @@ sub generate_income_statement {
         $rpt = LedgerSMB::Report::PNL::Income_Statement->new(
             %$request,
             formatter_options => $request->formatter_options,
-            from_date  => $request->parse_date( $request->{from_date} ),
-            to_date  => $request->parse_date( $request->{to_date} ),
             column_path_prefix => [ 0 ]);
+        #die np $rpt;
```
ehuelsmann, line 78: could you remove this stray comment?

ehuelsmann:
I'm afraid that fixing the comparisons broke the date intervals...

ehuelsmann:
@ylavoie how about approaching this differently: how about removing the complexity of needing to resolve this all server-side? How about making a bit of JS which calculates the correct start and end dates and sets the from and to dates accordingly?

lib/LedgerSMB/Scripts/pnl.pm
ylavoie: Done

lib/LedgerSMB/Report/Dates.pm
```diff
@@ -150,9 +149,9 @@ sub _collect_dates_comparisons {
 
     foreach my $i (1 .. $args{comparison_periods}) {
         push @dates, {
-            from_date => LedgerSMB::PGDate->from_input($args{"from_date_$i"}),
-            to_date => LedgerSMB::PGDate->from_input($args{"to_date_$i"}),
-            column_path_prefix => [ $i ]
+            from_date => LedgerSMB::PGDate->from_input($args{"from_date_$i"},$args{formatter_options}),
+            to_date => LedgerSMB::PGDate->from_input($args{"to_date_$i"},$args{formatter_options}),
+            column_path_prefix => [ sprintf('%02u', $i) ]
```
ehuelsmann, line 154: Nice find! I guess I never tested with more than 9 columns...

lib/LedgerSMB/Scripts/reports.pm
```diff
@@ -193,27 +193,25 @@ Generates a balance sheet
...
-        to_date  => $request->parse_date( $request->{to_date} ),
+        from_date  => $request->{from_date} ? $request->parse_date( $request->{from_date} ) : undef,
+        to_date  => $request->{to_date} ? $request->parse_date( $request->{to_date} ) : undef,
         column_path_prefix => [ 0 ]);
     $rpt->run_report($request);
 
-    for my $key (qw(from_month from_year from_date to_date internal)) {
-        delete $request->{$_} for (grep { /^$key/ } keys %$request);
+    for my $key (qw(from_date to_date date_from date_to)) {
+        delete $rpt->{$_} for (grep { /^$key/ } keys %$rpt);
```
ehuelsmann, line 207: why do we not need to delete `from_month` and `from_year`?

ylavoie: Once builders ran, all comparisons are set and we don't even need to cleanup $request or $rpt between passes.

PR #8170, opened 2024-04-21
1099 tax form printing fixed

ehuelsmann:
Various fixes:

* The Print button didn't pass the right values in its request
* The Print button wasn't a PrintButton, failing to download the result
* Company parameters (name, etc) not passed to form templates
* Missing drop-down to select which form to print
* Missing address data in query feeding forms
* Incomprehensible pdflatex errors thrown when company configuration data missing ("no line to terminate") [replaced missing data with default values pointing the user to a solution to the problem]

### DIFF
--- a/sql/modules/1099_reports.sql
+++ b/sql/modules/1099_reports.sql
@@ -24,6 +24,13 @@ CREATE TYPE tax_form_report_item AS (
     meta_number character varying(32),
     tax_id text,
     sales_tax_id text,
+    line_one text,
+    line_two text,
+    line_three text,
+    city text,
+    state text,
+    mail_code text,
+    country_id integer,
     acc_sum numeric,
     invoice_sum numeric,
     total_sum numeric);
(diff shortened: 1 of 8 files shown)


sql/modules/1099_reports.sql
```diff
@@ -116,8 +132,36 @@ RETURNS SETOF tax_form_report_item AS $BODY$
...
+        from eca_to_location eca2l
+               join location_class lc on lc.id = eca2l.location_class
+               join location l on l.id = eca2l.location_id
+       where eca2l.credit_id = entity_credit_account.id -- this is the LATERAL!
+         and lc.authoritative
+       order by lc.id, l.created
+      ) y
+       union all
+      select * from (
+      select entity.id, l.*
```
ehuelsmann, line 148: @ylavoie, should this have been `entity_credit_account.id` so that there is at most 1 record in the right side of the join for each record on the left?

ehuelsmann:
@ylavoie, I'm having a hard time figuring out how to test the fixes in this PR. Do you have ideas? Setting up the required data in a test database is a *lot* of work...

PR #8233, opened 2024-06-01
Eslint9 no ext option

ehuelsmann:


ehuelsmann:
Solved with Yves's PR.

PR #8245, opened 2024-06-09
Fix comparison reports csv

ylavoie:
This removes a spurious heading from Balance Sheet & PNL csv reports.
It also properly sorts the comparison columns on Balance Sheet csv reports.


ehuelsmann (approved):
Looks good!

PR #8318, opened 2024-07-29
Add ADR wrt duplicate 'accno' (and prevent duplicates)

### LINKED ISSUE #861: Schema doesn't disallow headings and accounts with same name
While the schema doesn't disallow headings with account numbers/accnos equal those found in accounts, especially old code queries the 'chart' view for account IDs, which then break: the query returns multiple rows, where only zero or one are allowed.

ehuelsmann:
Closes #861

### DIFF
--- a/doc/adr/0024-chart-of-accounts-schema.md
+++ b/doc/adr/0024-chart-of-accounts-schema.md
@@ -55,7 +55,7 @@ By changing the above, where the commonality will be stored in the `coa_node`
 table directly, adding additional fields to an `account` table which links
 back to the `coa_node` for the common fields, like:
 
-``sql
+```sql
 CREATE TABLE coa_node (
     id serial NOT NULL UNIQUE,
     parent int,
--- a/doc/adr/0024-chart-of-accounts-schema.md
+++ b/doc/adr/0024-chart-of-accounts-schema.md
@@ -1,0 +1,93 @@
+# 0024 Central CoANode table defining accounts and headings
+
+Date: 2020-01-12
+
+## Status
+
+Accepted
+
+## Context
+
+Issue #861 complains there's no guarantee in the schema which prevents
+accounts and headings being assigned the same `accno`.  The schema proposed
+below (taken from the wiki), seems to prevent that.  This is however not the
+case, because `UNIQUE` constraints and primary key constraints are not carried
+over from the inherited to the inheriting table.
+
+```sql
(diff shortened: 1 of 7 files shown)


doc/adr/0024-chart-of-accounts-schema.md
```diff
@@ -0,0 +1,93 @@
...
+) INHERITS (coa_translation);
+
+ALTER TABLE coa_translation ADD CHECK NOINHERIT (FALSE);
+```
+
+By changing the above, where the commonality will be stored in the `coa_node`
+table directly, adding additional fields to an `account` table which links
+back to the `coa_node` for the common fields, like:
+
+``sql
```
neilt, line 58: This should be ``` instead of ``

doc/adr/0024-chart-of-accounts-schema.md
```diff
@@ -0,0 +1,93 @@
...
+This is, however, worse than the problem being addressed: by adding a record
+to the `account` table for an existing `coa_node` record, the heading is
+transformed from a heading to an account from one moment to another.
+
+Additionally, there's no way to guarantee that accounts are the leafs in
+the account hierarchy tree in this schema.
+
+## Decision
+
+The downsides of current schema outweigh those of the proposed schema, so
```
neilt, line 81: Don't you have these backwards.

Do you mean the '... the downsides of proposed schema outweigh those of the current schema so there will be no change ...'

neilt (changes requested):

doc/adr/0024-chart-of-accounts-schema.md
ehuelsmann: Ouch. Yes, I mean this the other way around...

doc/adr/0024-chart-of-accounts-schema.md
ehuelsmann, line 58:
```suggestion
```sql
```

doc/adr/0024-chart-of-accounts-schema.md
ehuelsmann, line 81:
```suggestion
The downsides of the proposed schema outweigh those of the current schema, so
```

ehuelsmann (commented):
Thanks for the comments. Adding proposed changes, which I can incorporate through GitHub.

PR #8319, opened 2024-07-29
Add ADR on the database schema upgrade procedure

ehuelsmann:
Move one more wiki document to the ADR repository.


neilt (approved):
Does status of "accepted" mean "fully implemented"?

ehuelsmann:
> Does status of "accepted" mean "fully implemented"?

Hmm. I didn't think it would mean that in general; I think that in general "Accepted" means that this is ready for implementation and how it should be.

In this case specifically: yes, it's fully implemented, although the declaration of multiple "pre-change-checks" works a bit different.

PR #8329, opened 2024-08-02
Fix Ship To on purchase documents, orders and quotes

ehuelsmann:
Ship To didn't work at all on purchases (invoices, orders, requests for quote). Additionally, Ship To didn't work on orders and quotes. Even more, Ship To didn't work when invoked multiple times in a row, e.g. when clicking "Add To List" or when forgetting to select a shipping address.

### DIFF
--- a/old/bin/io.pl
+++ b/old/bin/io.pl
@@ -718,7 +718,7 @@ sub display_form {
 
     &{$subroutine}($numrows) if $numrows;
 
-    $form->hide_form(qw|locationid|);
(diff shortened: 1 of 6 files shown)


ehuelsmann:
It looks like I was a bit over-active: there's no provision for using shipping addresses on the sales invoice screen. Although I made the code work (there *are* provisions in the database...), I'm preparing a commit to disable the "Ship To" button on purchase documents.

Leaves that this PR is still very relevant to make shipping screens work for sales orders, quotes and invoices.

PR #8342, opened 2024-08-04
Remove stale LedgerSMB::App_State->Locale() references

ehuelsmann:


ehuelsmann:
Merging without CCI -- it's not being run.

PR #8363, opened 2024-08-12
Add the concept of an 'upgrade run'

ehuelsmann:


ehuelsmann:
Next step is to show the current logs at the end of an upgrade run. (Which was the point of starting this PR in the first place.)

PR #8366, opened 2024-08-14
Reorganize ShipTo screen for better alignment

ehuelsmann:
Put the "new location" and "new contact" entry rows in the same table as the listing of the existing ones: this aligns the column titles with the inputs, making the screen more self-explanatory.

At the same time, fix the "contacts" section to have radio buttons to select the appropriate contact information.  Do note that there's no way to store the selected value; the best way forward is to be determined.

The existing screen looks like:
image

The improved screen looks like:
image

### DIFF
--- a/old/bin/io.pl
+++ b/old/bin/io.pl
@@ -1618,6 +1618,12 @@ sub ship_to {
     $form->header;
 
 
+    for (qw| address1_ address2_ address3_ city_ state_ zipcode_ country_ type_ contact_ description_|) {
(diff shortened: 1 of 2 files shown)


neilt:
It is definitely an improvement.  

In a perfect life, the shipping address boxes would be bigger, but I get that you probably don't want a multiline form.  You might consider moving type, description, and contact to a second line.

I don't use shipping addresses so feel free to ignore.

PR #8374, opened 2024-08-18
Mention consistency checks in the action completion screen

ehuelsmann:
Also reflow the screen's HTML for readability.


neilt:
If you are working on consistency checks, it would be nice to be able to run those outside of the upgrade process.  Any chance of that?

ehuelsmann:
> If you are working on consistency checks, it would be nice to be able to run those outside of the upgrade process. Any chance of that?

Yep. Exists as a button on the setup.pl opening page. Since 1.11. please do submit requests for checks that have not been implemented yet.

PR #8382, opened 2024-08-22
Fix orders and quotes being frozen after Update

ehuelsmann:
Follow-up to #8348 which fixes most of the frozen orders and quotes, but not quotes and orders becoming frozen after Update in SAVED state.


ehuelsmann:
@tonka128 confirmed the fix.

PR #8414, opened 2024-09-27
feat: allow hyphen in host

walidmujahid:
For the case where host has hyphens. Like when trying to use to a digitalocean managed database server for your ledgersmb, the host looks like this: ledgersmb-do-user-66666666-0.k.db.ondigitalocean.com


ehuelsmann:
Thank you for sending in this fix! Your contribution is highly appreciated. I saw you backported this to 1.11. Could you do the same for 1.10¿ It has a few more days of community support left, after which I will make a final release. It would be nice to include this in that release.

walidmujahid:
@ehuelsmann Sorry, I just saw your comment. I just opened a backport PR for 1.10.

PR #8428, opened 2024-10-08
Restore heading_account in header

ylavoie:
Restore a heading from Balance Sheet & PNL csv reports
Undo part of #8245


ehuelsmann:
Thanks! These documents are versioned in the header. Could you increase the version number?

PR #8532, opened 2024-12-02
Fix 'New' (customer/vendor) link on invoices and transactions

ehuelsmann:
Thanks to @neilt for finding this problem!


neilt:
Confirmed fixed

PR #8566, opened 2025-01-02
Add ADR documenting what to do in Actions vs Persisters

ehuelsmann:
We currently have the JournalEntry persister handle deletion/posting of a transaction based on a value set in the workflow. (I.e. not as part of the execution of the action itself, which sets only the value in the workflow to trigger the persister behaviour.)


doc/adr/0101-separating-concerns-between-persisters-actions.md
```diff
@@ -0,0 +1,133 @@
...
+8. The database transaction is committed
+
+At various points in this flow observers are sent events from processing the
+action and optional state changes.
+
+### The problem with an active storage layer
+
+The `Workflow` state management library divides the concerns "performing actions"
+and "storing resulting workflow state" between the `Workflow::Action` and the
+`Workflow::Persister` respectively. Is the context of a passive storage backend
```
freelock, line 73: Typo - "Is" or "In"?

doc/adr/0101-separating-concerns-between-persisters-actions.md
```diff
@@ -0,0 +1,133 @@
...
+   i.e. actions operate solely on workflow state, unable to access the
+   persistence layer
+
+The downside of (1) is that in its design, Workflow didn't account for
+actions which need to access the database (storage) layer, requiring
+customization of the workflow to support the pattern.
+
+The downside of (2) is that it results in tight coupling between actions
+and persisters: any new actions which require invocation of stored
+procedures will need changes in the persister as well.
```
freelock, line 106: I'm trying to argue for a loose coupling here -- perhaps allow actions/conditions to get "auto-discovered" or registered in some dependency injection system or wired up through some external module, so the "persister" doesn't need to be modified at all to correctly support any number of actions

doc/adr/0101-separating-concerns-between-persisters-actions.md
```diff
@@ -0,0 +1,133 @@
...
+  Define associated actions, their associated conditions and their transition
+  target states
+* Conditions  
+  Define conditions (Perl code) for availability of actions
+* Actions  
+  Define which Perl code is run when the action is triggered -- this may include
+  running stored procedures in the PostgreSQL database
+* Validators  
+  Define checks to establish input validity in the workflow state before
+  executing the code associated with an action
```
freelock, line 47: What's the difference between a validator and a condition? (is a distinction necessary?)

doc/adr/0101-separating-concerns-between-persisters-actions.md
```diff
@@ -0,0 +1,133 @@
...
+* Persisters  
+  Store and load workflow state; in case of LedgerSMB, the state is stored in
+  a PostgreSQL database in a.o. the `workflow` and `workflow_context` tables
+* Observers  
+  Define Perl code to be triggered on events in a workflow
```
freelock, line 52: What's the difference between an action and an observer?

It sounds like you're making a distinction mainly on where the code is hooked in -- as a stored function in Postgres, vs a code plugin that stands alone outside the database?

I would argue that the role for both is the same -- we may just need to define two interfaces (inside postgres vs in a Perl file) to support where the code for an action/observer or condition/validator lives.

freelock (commented):
One typo found, and some more broad questions/comments...

doc/adr/0101-separating-concerns-between-persisters-actions.md
ehuelsmann, line 73:
```suggestion
`Workflow::Persister` respectively. In the context of a passive storage backend
```

ehuelsmann: In, indeed.

doc/adr/0101-separating-concerns-between-persisters-actions.md
ehuelsmann: Added documentation of the difference to the file. As for whether the distinction is necessary: they serve a different purpose. Even when we don't find the distinction important, they *are* components that the Workflow library provides (which is why I added them to the explanation).

doc/adr/0101-separating-concerns-between-persisters-actions.md
ehuelsmann: A description of the difference between the two is now in the file; in summary: an action triggers a state change, an observer is notified of a state change only.

doc/adr/0101-separating-concerns-between-persisters-actions.md
ehuelsmann: The workflow is fully dependency-injected, or at least: actions are linked to code dynamically through configuration. The only reason a persister and an action can be mutually dependent is when it's the persister's role to do *all* database communication, where the action "only" adds flagging to the workflow state so the persister performs the expected database actions.

The route this PR is now taking is to have loose coupling: actions have access to the active layer of the database (stored procedures) while the persister has access to the passive layer of the database (tables).

PR #8655, opened 2025-02-15
Add ADR revisiting deletion of core data model elements

ehuelsmann:


neilt (changes requested):
It is not clear to me what the rules are for coding deletions.  Any data actually used should not be delete-able.  For example a customer that has never been sent an invoice or quote can be deleted, but a customer that has been sent a quote, invoice, etc. cannot be deleted.  I think the delete-able data rules need to be more clearly stated. 

Same for inventory items.  If the inventory item was never quoted, purchase, was in inventory, or sold, then it can be deleted, otherwise not.

I guess the definition of "used" needs to be clarified.

ehuelsmann:
Thanks for the review! I've tried to explain much more thoroughly what's meant. I hope it adds clarity.

neilt (approved):
Yes, I think that is much better, thank you.

PR #8662, opened 2025-02-16
Fix and test AP/AR Search by date period

nick-prater:
AP/AR search was not applying a date period filter (from_year, from_month,
interval) and was instead returning transactions for all time.

This patch fixes this issue and adds tests for all time period searches of
AP/AR transactions, including the alternative from_date/to_date range
selection.


ehuelsmann:
Hi Nick! Good to see your contribution (which I take it means you still use LedgerSMB :-) ) ! Hope you're doing wellThanks for going the extra mile and even add tests.

PR #8666, opened 2025-02-19
Remove irrelevant date search periods

nick-prater:
On AR/AP search, dates can be filtered by period, specifying a start month, start year and period.

This patch removes non-working 'Day' and 'Week' periods, leaving the working 'Current', 'Month', 'Quarter' and 'Year' options in place.

The 'Day' and 'Week' periods make no sense, as the start point can only be specified with granularity of a month.

Furthermore, the 'Day' and 'Week' periods are unimplemented in the backend code, so do not work.


ehuelsmann:
Truely bizarre how that ever went in...

PR #8683, opened 2025-02-26
Remove double-encoding from labels; fix setup.pl with French (due to quoting)

### LINKED ISSUE #8663: setup.pl doesn't load using French language
### Version

1.13.0-dev

Firefox, This problem isn't browser related

### What happened?

Set browser language to French.
Browse to /setup.pl
Page does not render. Console gives error:
`dojo/parser::parse() error Error: SyntaxError: missing } after property list in data-dojo-props='value:'', placeHolder:'Sélectionnez ou entrez l'utilisateur''`

Issue is caused by apostrophes in translated text used as DB admin login field placeholder.

Page should have rendered.

ehuelsmann:
Fixes #8663

### DIFF
--- a/UI/lib/elements.html
+++ b/UI/lib/elements.html
@@ -336,7 +336,7 @@
         ELSE;
            label_class = "";
         END;
-     "<label $label_id $label_for $label_class >" _ text(element_data.label) _ "</label>";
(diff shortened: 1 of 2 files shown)


nick-prater (approved):
Confirmed fix - setup.pl now loads fine and displays apostrophes using French language for browser. Thanks!

PR #8735, opened 2025-03-24
Entity search by country with new <lsmb-country> Web Component

nick-prater:
Introduce new `<lsmb-country>` country select drop-down Web Component and
use it for specifying a Country filter on the Entity Search page.


UI/src/elements/lsmb-country.js
```diff
@@ -0,0 +1,72 @@
...
+            /* Instantiate *and* initialize store before assigning to
+             * the object property to avoid a race condition.
+             */
+            const store = useCountriesStore();
+            await store.initialize();
+            this.store = store;
+        }
+
+        const options = this.store.items.map((element) => {
+            return { label: element.name, value: element.code };
```
ehuelsmann, line 41: you may be able to skip this "renaming" operation by setting the `idProperty` and `labelAttr`  to `code` and `name` respectively. Then you can return the elements as-is.

UI/src/elements/lsmb-country.js
```diff
@@ -0,0 +1,72 @@
...
+        });
+
+        if ("default_blank" in props) {
+            options.unshift({ label: "", value: "" });
```
ehuelsmann, line 45:
```suggestion
            options.unshift({ code: "", name: "" });
```

which then changes this line to unshift the named properties instead of "label" and "value".

UI/src/elements/lsmb-country.js
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        this.dojoWidget = new Select({
+            name: props.name,
+            options: options
+        });
```
ehuelsmann, line 51:
```suggestion
        idProperty: "code",
        labelAttr: "name"
        });
```

ehuelsmann (commented):
Love it!

Please see my comments inline.

UI/src/elements/lsmb-country.js
nick-prater: Unfortunately, as far as I can see, the dojo Select widget doesn't support `idProperty` and `labelAttr` attributes - I've tested it and looked at the docs here: https://dojotoolkit.org/reference-guide/1.10/dijit/form/Select.html

Am I missing something?

ehuelsmann: I'm using https://dojotoolkit.org/api/ ; but, no, I don't think you missed something. I genuinely thought that the `idProperty` existed. (A major shortcoming it doesn't, really, but that's beyond the scope of this PR.)

ehuelsmann (approved):

PR #8747, opened 2025-04-05
1.12 fix entity search by country

nick-prater:
Minimal fix for 1.12 Entity Search by country.

### DIFF
--- a/sql/modules/Company.sql
+++ b/sql/modules/Company.sql
@@ -354,16 +354,13 @@ RETURN QUERY EXECUTE $sql$
                    OR state ILIKE '%' || $7 || '%')
               AND ($8 IS NULL
                    OR mail_code ILIKE $8 || '%')
-              AND ($9 IS NULL
-                   OR EXISTS (select 1 from country
-                               where name ilike '%' || $9 || '%'
-                                  or short_name ilike '%' || $9 || '%'))
-                       )
+   )
    SELECT e.id, e.control_code, ec.id, ec.meta_number::text,
           ec.description, ec.entity_class,
           c.legal_name, c.sic_code::text, b.description , ec.curr::text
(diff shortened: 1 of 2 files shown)


sql/modules/Company.sql
```diff
@@ -389,6 +385,9 @@ LEFT JOIN business b ON (ec.business_id = b.id)
                              join entity_to_location etl
                                   ON m.id = etl.location_id
                             where etl.entity_id = e.id))
+           AND ($9 IS NULL
+                OR entity_country.short_name ILIKE $9
+                OR entity_country.name ILIKE $9 || '%')
```
ehuelsmann, line 390: Is it accidental that you didn't write:

```suggestion
                OR entity_country.name ILIKE '%' || $9 || '%')
```

?

nick-prater: Thanks for reviewing and asking the question. I appreciate it.

It was deliberate that the search works on short name (country code) or the first part of the full country name. While not perfect, but I think it's the best compromise for a minimal backport.

There are other fields for which we apply the same pattern, such as control_code or mail_code. (Of course it's flawed that the user has to guess, for each field, how the search works, with no way of knowing from the interface, but that's a different problem to solve).

Consider searching for "ES", looking for entities from Spain. Currently that will return results for Spain (because the short_name matches exactly) and Estonia (because the country name starts with those letters).

If I broadened the search as per your suggestion, it would also return results for United States, Phillipines, Seychelles and another dozen or so countries with `es` somewhere in their name.

It's why I used a drop-down for the fix in master. Which also allows localisation of the country names. But I think backporting a dropdown is too invasive - it's a fundamental UI change.

Happy to consider changing if you disagree.

ehuelsmann (approved):

PR #8749, opened 2025-04-06
Localise country web component

nick-prater:
This PR:
  1) Extends the countries api to additionally return a localized version of the
country name
  2) Updates the `<lsmb-country>` web component to present
a sorted list of the localized country names as options.


ehuelsmann (approved):
Looking good. Including the extension of the API definition. The correctness of it is checked as part of the webservice tests and those pass. So the definition is valid as well.

PR #8766, opened 2025-04-20
Fix Formatting of Password Change Screen

nick-prater:
Fields and labels were not aligned on the password change screen for certain
stylesheets. This PR makes the page consistent with other lsmb forms, using
our standard `two-column-grid` layout.

No change to functionality.

### DIFF
--- a/UI/js-src/lsmb/users/ChangePassword.js
+++ b/UI/js-src/lsmb/users/ChangePassword.js
@@ -33,7 +33,6 @@ define([
                 "new password": "New Password",
                 verify: "Verify",
                 change: "Change Password",
-                "no-oldpw": "No Old Password",
                 strength: "Strength"
             },
             lstrings: {},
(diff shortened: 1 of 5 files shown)


UI/js-src/lsmb/users/ChangePassword.js
```diff
@@ -30,7 +30,7 @@ define([
             _lstrings: {
                 title: "Change Password",
                 "old password": "Old Password",
-                "new password": "New password",
+                "new password": "New Password",
```
ehuelsmann, line 33: This changes a translated string, which means see need to update the translation template file. Please include a run of "make pot".

ehuelsmann (changes requested):
Overall good to move to grid layouts! A nice improvement!

UI/js-src/lsmb/users/ChangePassword.js
nick-prater: A fix is needed so that `make pot` parses multiple `text(..)` strings
placed on a single line, which occurs in this case. I'll do that and submit
the results in a separate PR.

For now, the highlighted string is not included in translations, so we're no
worse off.

ehuelsmann (approved):
After discussion on my prior point, agree and approve.

PR #8814, opened 2025-05-18
Fix inactive Login button on login page

ehuelsmann:
When navigating back to the login page using the Back button, the Login button remains disabled, no matter what. The only way to activate it again is to reload the login page (F5 or Ctrl-R).

This change accepts any edit in the input fields to enable the Login button.


ehuelsmann:
Maybe this failure is on my repository only. I see no failed on the main repo...

PR #8824, opened 2025-05-25
Fix 8822 112

einhverfr:


ehuelsmann:
You tried to to merge into 1.2 instead of 1.12. I think it works now that I changed the base branch...

einhverfr:
Thanks!

PR #8846, opened 2025-05-31
Make setup.pl schema path consistent for login and upgrade operations

nick-prater:
When logging in to setup.pl, the `public` database schema is included in the search path, after that configured in `ledgersmb.yaml`.

This patch is to use the same search path during the upgrade operation, where previously only the configuration file schema was used.

Previously a database created in the `public` schema would be offered for upgrade, but then the upgrade would fail with a cryptic error `Upgrade workflow error: no next step for ''`.

Now a more relevant error is displayed, indicating that the schema set in `ledgersmb.yaml` does not exist in the current database.

### DIFF
--- a/lib/LedgerSMB/Scripts/setup.pm
+++ b/lib/LedgerSMB/Scripts/setup.pm
@@ -907,6 +907,7 @@ sub upgrade {
 
     my $settings = $request->{_wire}->get( 'setup_settings' );
     my $auth_db = ($settings and $settings->{auth_db}) // 'postgres';
+    $database->{connect_data}->{options} = "-c search_path=$database->{schema},public";
     my $dbinfo = $database->get_info($auth_db);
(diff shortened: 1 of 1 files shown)


lib/LedgerSMB/Scripts/setup.pm
```diff
@@ -907,6 +907,7 @@ sub upgrade {
 
     my $settings = $request->{_wire}->get( 'setup_settings' );
     my $auth_db = ($settings and $settings->{auth_db}) // 'postgres';
+    $database->{connect_data}->{options} = "-c search_path=$database->{schema},public";
```
ehuelsmann, line 910: Could you add a comment here and in `sub login()` to cross reference these and make sure we keep them in sync?

ehuelsmann (changes requested):

ehuelsmann (approved):

PR #8848, opened 2025-05-31
Add ADR describing the switch from Dojo Toolkit to Quasar

ehuelsmann:


nick-prater (approved):
Reads well to me and provides the context I was hoping to find. Thank you.

PR #8874, opened 2025-06-07
Templating using custom elements ('input' elements)

ehuelsmann:
Other elements to follow, but lets start with 'input' elements as these already allow us to hove away from some of the code in `UI/js-src/lsmb/`.

Encapsulating Dojo widgets helps to ultimately replace the widgets with something else.


nick-prater (approved):
You've been busy!
These changes read well to me and bring more consistency.
Thank you. Good with me.

PR #8876, opened 2025-06-09
Make menu setup declarative and decouple Roles.sql

ehuelsmann:
Note that loading the entire menu also removes the need to update it through change files, which can be deleted now.


lib/LedgerSMB/Company/Menu.pm
```diff
@@ -0,0 +1,232 @@
...
+            }
+            for my $acl ($item->getChildrenByTagName('acl')) {
+                $self->_add_menu_acl($id, $acl);
+            }
+        }
+
+        my @submenus = $xml_node->getChildrenByTagName('menu-item');
+        local $ctx->{position} = 0;
+        for my $submenu (@submenus) {
+            $self->_add_menu_node( $ctx, $submenu, $id );
```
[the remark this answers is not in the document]
ehuelsmann: Keeping it this way: the `$ctx` is supposed to be applicable to the entire duration of the load, except the `position`.

locale/menu.xml
```diff
@@ -0,0 +1,750 @@
+<?xml version="1.0"?>
+<menu xmlns="http://ledgersmb.org/xml-schemas/menu" id="0" label="Top-level">
```
neilt, line 2: When I tried to load https://ledgersmb.org/xml-schemas/menu in a browser it returned a page not found.  how can I view this?

neilt (approved):
Not really reviewed, but browsed. Did not find any other obvious issues.

locale/menu.xml
ehuelsmann: The URL is nothing more than a namespace. The same happens in the 'initial-data.xml' and CoA xml files. I have the XSD added to my tree locally. I'll push that when the tests complete so I know the tests work.

ehuelsmann: Pushed now.

nick-prater (approved):
Reads well to me. Thanks. Good bit of work!

Issue #8850, opened 2025-06-01
Improve main menu structure


neilt:
A discussion came up about changing the menu structure while considering renaming the menu item "Cash" to "Cash & Banking" about other changes to the menu structure.

This issue is to track these discussions. Legend ✅ = Approved, ❌ = Discarded, ⚠️ = Still in discussion

Proposed unordered changes for discussion:
1. ✅ AR should be named Accounts Receivable. All of the other menu items are spelled out, so this improves consistency
2. ✅ AP should be named Accounts Payable. All of the other menu items are spelled out, so this improves consistency
3. ✅ HR should be named Human Resources. All of the other menu items are spelled out, so this improves consistency, but see below for removing HR completely.
4. ✅ Reports -> Trial Balance, Income Statement, and Balance Sheet should be under General Journal -> Reports or the Reports menu item should be renamed because all LSMB reports in are NOT under this menu item. An alternative would be to move all reports here, but I think that is not as good as it is now.
5. ✅ Reports -> Inventory and COGS should be under Goods and Services -> Reports  (EH: we need a report explaining the "Inventory" account from a financial perspective, which then should be under General Journal, I think?)
7. ⚠️ HR -> Employees should be moved to Contacts. I understand there used to be HR functionality and that it might come back, but let's handle that when or if it ever happens. Probably change the name to Payroll.
8. ❌ If keeping HR then TimeCards should be under HR with Employees. And this would be a good reason to keep HR
9. ❌ Rename AR to Sales. This is less technical and would make easier understanding for new users.  It would also make LSMB seem less like an accounting system and more like an ERP or Sales system. Then it would also make sense to move Order Entry -> Sales Order and Order Entry -> Reports -> Sales Orders menu items to the new Sales menu.
10. ❌ Rename AP to Purchases. This is less technical and would make easier understanding for new users. It would also make LSMB seem less like an accounting system and more like an ERP or Sales system. Then it would also make sense to move Order Entry -> Purchase Order and Order Entry -> Reports -> Purchase Orders to the new Purchases menu
11. ✅ Rename Cash to Cash & Banking

Proposed unordered menu design rules (when approved will be moved to an ADR):
    1. ✅ All menu and sub-menu names should pass English spellcheck.  For example, "Pricegroups" and "Partsgroup" fail in the current menu structure.  This rule would make both translations and understanding easier.
[...]

PR #8880, opened 2025-06-09
Improve menu as proposed in #8850

ehuelsmann:
This PR addresses points (1), (2), (4), (5) and (10) from #8850. (3 - Renaming HR to Human Resources) waits for a better name than Human Resources.

~~THIS IS A DRAFT PR -- Please rebase and merge *only* after #8876 has been merged.~~


ehuelsmann:
@auspex, this includes the majority of your feedback, I think. Please have a look.

locale/menu.xml
```diff
@@ -24,7 +24,7 @@
       </acls>
     </menu-item>
   </menu-item>
-  <menu-item id="1" label="AR">
+  <menu-item id="1" label="Accounts Receivable">
```
nick-prater, line 27: Do we have a way to update the translations, so that existing translations of "AR" will instead match "Accounts Receivable"? Or are we just going to accept breakage, then have a push for translations at a later date?

ehuelsmann: I guess we can pull the translations from Transifex using their client. Then replace `msgid "AR"` with `msgid "Accounts Receivable"` and push the translations back?

ehuelsmann: I tried to do some juggling with translations on transifex. I failed... I did manage to create translatable entries for "Accounts Receivable" and "Accounts Payable".  However, they ended up empty in all translations.

ehuelsmann: At this point, I'm happy to accept the breakage because this should be 3 minutes of work for all translators and it's apparently more work on my side...

ehuelsmann: Solved. I've backported these with the help of Text::PO::Parser :tada:

PR #8886, opened 2025-06-12
Fix overpayments

einhverfr:
This branch and (initially draft) Pull Request fixes Overpayments issues and allows overpayments to be posted on the master branch (tests forthcoming while in draft mode).

Please note that while I intend to backport this, I have also fixed two fundamentally broken uses of the translation API (which lead to untranslated strings in 1.12.  I have left several other bad calls which will be a separate bug fix for master after this gets merged.

### DIFF
--- a/lib/LedgerSMB/Scripts/payment.pm
+++ b/lib/LedgerSMB/Scripts/payment.pm
@@ -1673,7 +1673,7 @@ sub use_overpayment2 {
 
     #list all the vendor/customer
     @vc_info = $Payment->get_entity_credit_account();
-    push @vclist = {}; # inserting a blank at the start
+    push @vc_info, {}; # inserting a blank at the start
     for my $ref (0 .. $#vc_info) {
         my ($name) = split(/:/, $vc_info[$ref]->{name});
         push @vc_list, { value            => $vc_info[$ref]->{id},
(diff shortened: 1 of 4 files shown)


lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1677,6 +1672,7 @@ sub use_overpayment2 {
 
     #list all the vendor/customer
     @vc_info = $Payment->get_entity_credit_account();
+    push @vclist = {}; # inserting a blank at the start
```
ehuelsmann: Needs comma instead of assignment.

lib/LedgerSMB/Scripts/payment.pm
```diff
@@ -1935,12 +1877,13 @@ sub post_overpayment {
...
     {
 
-        if ($request->{"checkbox_$count"})
+        if ($request->{"checkbox_$count"} or !$request->{"amount_$count"})
         {
             $count++;
             next;
         }
 
+        # complication -- this does not set $entity_name
```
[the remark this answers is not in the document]
einhverfr: Yes, this is a note to just remove the ambiguity here once things work.

lib/LedgerSMB/Scripts/payment.pm
einhverfr: Also needed the right variable :-P

PR #8888, opened 2025-06-12
Add AI warning recommended by ChatGPT master

neilt:
so that it will not simplify docker installations


neilt:
OK, I'll redo later

PR #8996, opened 2025-08-15
Add freedom for users to store extra attributes

ehuelsmann:
Start with several good candidates: GL accounts, parts, entities, SIC codes and asset items.

Re #6964

### DIFF
--- a/doc/adr/0107-using-postgresql-json-jsonb-data-type.md
+++ b/doc/adr/0107-using-postgresql-json-jsonb-data-type.md
@@ -37,9 +37,8 @@ to the software; no ADR should reduce the importance of that.
 There are however, cases where users want to add data to LedgerSMB that is
 neither foreseen nor necessary for the execution of its primary function. To
 that extent, there is no consistency requirement within the application nor
-any means for the application to verify consistency. Since the structure of
-this additional data is - by definition - not known beforehand, it's not
-possible to incorporate in its schema.
+any means for the application to verify consistency. A characteristic of this
+data is that its structure cannot be known beforehand.
 
 ## Decision
 
--- a/doc/adr/0107-using-postgresql-json-jsonb-data-type.md
+++ b/doc/adr/0107-using-postgresql-json-jsonb-data-type.md
@@ -1,0 +1,67 @@
+# 0107 Using PostgreSQL's JSONB data type
+
+Date: 2025-08-15
+
+## Status
+
+Draft
+
+## Context
+
+PostgreSQL has added the JSON data type in version 9.2 and the JSONB data type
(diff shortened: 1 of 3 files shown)


Alt-Tech-Au:
We would use this immediately for generic parts data or instructions for automation, for example information like "is this dangerous goods" that may only apply to us as we ship batteries etc. (I an think of at least 3 cases where we are using fields in a part for something other than what they are supposed to be used for so that information can be moved there and clean up our database a lot!

doc/adr/0107-using-postgresql-json-jsonb-data-type.md
```diff
@@ -0,0 +1,67 @@
...
+of using a relational database such as indexing and data consistency. These
+are properties that the LedgerSMB team has worked hard on since 2006 to *add*
+to the software; no ADR should reduce the importance of that.
+
+There are however, cases where users want to add data to LedgerSMB that is
+neither foreseen nor necessary for the execution of its primary function. To
+that extent, there is no consistency requirement within the application nor
+any means for the application to verify consistency. Since the structure of
+this additional data is - by definition - not known beforehand, it's not
+possible to incorporate in its schema.
```
neilt: I disagree with this sentence "Since the structure of this additional data is - by definition - not known beforehand, it's not possible to incorporate in its schema."

It should be "cannot be known beforehand" Not known is potentially a copout for lack of rigor and effort. So how about something like:

 "A characteristic of this data is that its structure cannot be known beforehand."

neilt (changes requested):
Have not completed my review yet

doc/adr/0107-using-postgresql-json-jsonb-data-type.md
neilt: I was reviewing before you added the decision section so it is ultimately covered nicely. But I would still like to see the wording tightened up here.

sql/changes/1.13/custom-attributes.sql
```diff
@@ -0,0 +1,118 @@
+
+create table custom_attribute_metadata (
+  id uuid default gen_random_uuid() primary key,
+  name text not null,
+  obsolete boolean not null default false,
```
neilt: Why is obsolete not a date where the attribute becomes obsolete?  This could be set to now when editing or it could be set to expire when created.

ehuelsmann: Nice enhancement. Added.

doc/adr/0107-using-postgresql-json-jsonb-data-type.md
ehuelsmann: I've taken your wording. I like it.

PR #9068, opened 2025-09-27
Add 'Contributing' section from the Wiki to CONTRIBUTING.md

neilt:
When merged I'll remove the content from the Wiki.


ehuelsmann (approved):
Nice follow-up would be to mention Starring our repo also helps as a vote of support of the developers.

PR #9316, opened 2026-03-06
Start numbering Control Code at A-00001 instead of A-00002

ehuelsmann:


ehuelsmann:
Misunderstood the situation. Not merging.

PR #9508, opened 2026-04-27
Payments as first order transactions

ehuelsmann:
Changes recording of payments as first order transaction.

Re #7460


sql/modules/Payment.sql
```diff
@@ -880,177 +907,162 @@ BEGIN
...
+              in_op_source[out_count],
+              in_op_memo[out_count],
+              currval('open_item_id_seq'));
+    END LOOP;
+  END IF;
+  return var_payment_id;
 END;
+
 $$ LANGUAGE PLPGSQL;
 -- I HAVE TO MAKE A COMMENT ON THIS FUNCTION
```
neilt: "I HAVE TO MAKE A COMMENT ON THIS FUNCTION" comment unnecessary unless the comment is wrong.

xt/data/42-pg/Reconciliation.sql
```diff
@@ -1,28 +1,128 @@
...
+  │  -202    │ 1000-01-01 │ '1'      │ GL EXCLUDED: transdate ≠ pay. date │
+  │  -203    │ 1000-01-01 │ '1'      │ GL EXCLUDED: transdate ≠ pay. date │
+  ├──────────┼────────────┼──────────┼────────────────────────────────────┤
+  │  -2000    │ 1000-01-01 │ '1'      │ EXCLUDED: cleared + unapproved     │
+  │  -212    │ 1000-01-03 │ '1'      │ EXCLUDED: cleared                  │
+  │  -214    │ 1000-01-03 │ '1'      │ EXCLUDED: unapproved line          │
+  └──────────┴────────────┴──────────┴────────────────────────────────────┘
+
+  Expected reconciliation output (1 aggregate row):
+    payment -201  date 1000-01-03  total amount = sum of all included lines
```
neilt, line 125: Comments ending here are great

xt/data/42-pg/Reconciliation.sql
```diff
@@ -43,6 +143,11 @@ INSERT INTO account(id, accno, description, category, heading, contra)
 values (-201, '-11112', 'Test Act 2', 'A',
```
neilt: From here on it would be nice to have better comments

See that there is a strong association with the file's initial comments, but I was somewhat lost in trying to figure out what the various inserts purpose was.

neilt (commented):
I really don't have enough knowledge to provide any logic comments.

I liked the improved comments. Of course, more would be better for someone like me that does not understand it as well as you.

PR #9638, opened 2026-06-12
Upgrade dojo webpack plugin dependency

ehuelsmann:
This should enable upgrading the vulnerable 'tmp' dependency.


ehuelsmann:
Doesn't work due to Yarn applying 1 day minimal release age. Need to wait.

PR #9742, opened 2026-07-23
Simplify use of configuration hashes in ledgersmb.yaml

ehuelsmann:
Beam::Wire 1.031 supports traversal of bare hashes using $ref, eliminating the need to instantiate a Beam::Wire object on every intermediate level being traversed. (Yes, this was possible with $path, but (a) that's deprecated now and (b) using that keyword would add a slew of extra dependencies -- which is why we didn't use either. The old config is fully forward compatible; this is just simpler.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -5,7 +5,7 @@ requires 'perl', '5.38.0';
 requires 'Array::PrintCols';
 requires 'Archive::Zip';
 recommends 'Authen::SASL';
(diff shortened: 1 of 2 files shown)


ehuelsmann:
Thanks to preaction/Beam-Wire#93

ehuelsmann:
Filed preaction/Beam-Wire#97 about this failure.

PR #9786, opened 2026-08-08
Start phasing out vouchers

ehuelsmann:


doc/adr/0112-vouchers-no-longer-necessary.md
```diff
@@ -0,0 +1,44 @@
...
+`payment_links` for single payments.
+
+As payments are now first-order transactions - i.e. no longer lumped into
+`ar`/`ap` transactions - they can be approved on a transaction basis:
+vouchers are no longer required to identify transaction journal sub-groups.
+
+## Decision
+
+1. Batches will be groups of transactions
+2. Vouchers lost their function of grouping journal lines
```
neilt: The tense of 1 and 2 should be the same.

For example:

2. Vouchers will lose their function of grouping journal lines

sql/changes/1.14/transaction-batches.sql
```diff
@@ -0,0 +1,35 @@
+
+/*
+ * Implement ADR 0112 for GL and Upgrade vouchers
+ */
+
+
+alter table transactions
+  add column batch_id int references batch(id);
+
```
neilt, line 9: I would add a comment to the column, for example:

comment on column transactions.batch_id is 'The batch id if the transaction is part of a batch, see ADR 0112';

neilt (changes requested):

doc/adr/0112-vouchers-no-longer-necessary.md
ehuelsmann: Done.

sql/changes/1.14/transaction-batches.sql
ehuelsmann: Thanks for the suggestion. Yes, that's much better than all these columns without documentation.

PR #9842, opened 2026-08-28
Prevent editing existing GIFI codes

### LINKED ISSUE #9493: Configure GIFI codes UI error
1. Go to General Journal -> GIFI
2. Enter a new code
3. Select "Modify" on the row just entered
4. Both the Code and Description appear in the UI as modifiable. Both accept a cursor and allow editing.
5. But changing the Code has no effect when you click Save

Expected:

The column Code either:
1. should not be editable and should not accept the cursor, or
2. the code should be changed in the database when edited and saved

ryanduguid:
## Summary

- add optional immutable column metadata for configuration tables
- keep existing GIFI codes read-only while descriptions remain editable
- preserve code entry for new rows and add regression coverage for both cases

Fixes #9493

## Verification

- `corepack yarn test:unit` (62 tests passed)
- `corepack yarn eslint tests/specs/views/GIFI.spec.js`
- `corepack yarn eslint src/views/GIFI.vue src/components/ConfigTableRow.vue`
- `corepack yarn build`
- `git diff --check`

## Unverified checks

- The API-backed Vitest project was not run because this change is confined to the Vue configuration-table UI.
- Repository-wide lint could not provide a clean signal in this Windows checkout because Prettier reports CRLF endings across untouched files. The three changed files were normalised to LF and pass the targeted lint commands above.

### DIFF
--- a/UI/src/components/ConfigTableRow.vue
+++ b/UI/src/components/ConfigTableRow.vue
@@ -140,7 +140,9 @@ let mouseOverDefault = ref(false);
                 :type="column.type"
                 :value="data[column.key]"
                 :name="column.key"
-                :readonly="!editing"
+                :readonly="
+                    !editing || (props.type === 'existing' && column.immutable)
+                "
                 :class="editing ? 'editing' : 'neutral'"
                 class="input-box"
                 @input="(e) => send({ type: 'update', key: column.key, value: e.target.value })"
--- a/UI/src/views/GIFI.vue
+++ b/UI/src/views/GIFI.vue
@@ -10,7 +10,7 @@ import ImportCsvGifi from "@/views/ImportCSV-GIFI";
 
 const { t } = useI18n();
 const COLUMNS = [
-    { key: "accno",       type: "text", head: t("Code") },
+    { key: "accno", type: "text", head: t("Code"), immutable: true },
     { key: "description", type: "text", head: t("Description") }
 ];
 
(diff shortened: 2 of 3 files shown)


ehuelsmann (commented):
Thank you for your contribution! The code changes look good!

The webservice API also allows modification of GIFI codes and descriptions. For consistency, I think it would be good if that would disallow changing the codes. Did you verify it does? If not, that would be a nice expansion of this PR or a good follow-up.

Would you be up for that?

ryanduguid:
I traced the webservice path at this PR's current head. `PUT /gl/gifi/{id}` in `lib/LedgerSMB/Routes/ERP/API/GeneralLedger.pm` takes `accno` from the route's `{id}`, reads only `description` from the JSON body, and calls `_update_gifi`. That SQL updates `description` only, with the route code and ETag in the WHERE clause. A different `accno` in the body is therefore ignored and cannot rename the existing GIFI code. The model's `code` attribute is also read-only.

So the existing update path already preserves the code, matching this UI change. This is source-level verification; I have not run the API-backed integration suite locally because the Perl/PostgreSQL service is unavailable here. If the intended API behaviour is to reject a mismatching body code explicitly rather than ignore it, that would be an additional behaviour change.

ehuelsmann:
Thanks for all these verifications! I think this PR aligns the GIFI UI code paths with the API code paths. That's a good thing.

Having the API code throw an error if the gifi code in the URL differs from the code in the body would be a nice change. I won't consider it blocking to merge this PR.

Thanks for the contribution and welcome to the team!

Issue #9277, opened 2026-02-21
Some SQL functions don't adhere to naming convention


ehuelsmann:
Functions have the following naming convention: `<module>__<function>`, where the module and function part have only single underscores. These functions don't adhere to this naming scheme and need to be renamed:

* [ ] `_entity_location_save`
* [ ] `account_get`
* [ ] `account_has_transactions`
* [ ] `account_heading_get`
* [ ] `account_heading_list`
* [ ] `account_heading_save`
* [ ] `array_endswith`
* [ ] `array_splice_from`
* [ ] `array_splice_to`
* [ ] `asset_dep_get_usable_life_yr`
* [ ] `asset_dep_straight_line_month`
* [ ] `asset_dep_straight_line_yr_d`
* [ ] `asset_dep_straight_line_yr_m`
* [ ] `asset_nbv_report`
* [ ] `asset_report_partial_disposal_details`
* [ ] `avgcost`
* [ ] `batch_create`
* [ ] `batch_delete`
* [ ] `batch_get_class_id`
* [ ] `batch_get_class_name`
* [ ] `batch_get_users`
* [ ] `batch_list_classes`
* [ ] `batch_post`
* [ ] `batch_search_empty`
* [ ] `batch_search_mini`
* [ ] `business_unit_get`
* [ ] `cdc_update_last_updated`
* [ ] `chart_get_ar_ap`
* [ ] `chart_list_all`
* [ ] `chart_list_cash`
* [ ] `chart_list_discount`
* [ ] `chart_list_overpayment`
* [ ] `chart_list_search`
* [ ] `check_expiration`
* [ ] `company_get_billing_info`
* [ ] `compound_array`
* [ ] `cr_coa_to_account_save`
* [ ] `cr_report_block_changing_approved`
* [ ] `cr_report_line_cleared_update`
* [ ] `cr_report_line_link_insert`
* [ ] `cr_report_submitted_update`
* [ ] `customer_location_save`
* [ ] `date_get_all_years`
* [ ] `days_in_month`
* [ ] `defaults_get_defaultcurrency`
* [ ] `draft_approve`
* [ ] `draft_delete`
* [ ] `eca_bu_trigger`
* [ ] `employee_search`
* [ ] `entity_credit_get_id`
* [ ] `entity_credit_get_id_by_meta_number`
* [ ] `entity_save`
* [ ] `eoy_close_books`
* [ ] `eoy_create_checkpoint`
[...]

PR #9845, opened 2026-08-30
Rename array_endswith SQL function

Yurii201811:
## Summary

- rename `array_endswith` to `util__array_endswith`
- update both financial-statement callers and the pgTAP assertion
- regenerate the function blacklist
- add a schema change that removes the legacy function before modules reload

## Verification

- `git diff --check`
- blacklist compared with the module function inventory
- disposable PostgreSQL 14 fresh-install and upgrade-path simulation

## Not run locally

- repository Perl and pgTAP suites; required local modules are unavailable, so CI will provide those results

Part of #9277

### DIFF
--- a/sql/changes/1.14/drop-function-array_endswith.sql
+++ b/sql/changes/1.14/drop-function-array_endswith.sql
@@ -1,0 +1,1 @@
+DROP FUNCTION IF EXISTS array_endswith(anyelement, anyarray);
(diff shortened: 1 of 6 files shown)


ehuelsmann (approved):
Thank you for picking up this change. The code looks good and the tests all run fine, including the ones you indicate you didn't verify.

ehuelsmann:
Congratulations on your first contribution and welcome to the LedgerSMB project!

PR #9848, opened 2026-08-31
Payments - being a single exchange of cash - can *never* hit multiple cash accounts

ehuelsmann:
Split any payments which did do so (now prevented, but historically allowed).


neilt (approved):
Didn't check the logic. Great job on the comments

